PR #44, opened 2023-05-17
Add support for Azure OpenAI

DennizSvens:
This PR adds support for Azure OpenAI. Azure OpenAI is similar to OpenAI but requires a base URL for the Azure instance, deployment name of the model and the API version to be passed as arguments. The openai Python package is still used as the underlying technology for both.

Here's a detailed list of changes:
1. Added support for Azure OpenAI as a model hosting provider.
2. Included additional arguments to pass the base URL for the Azure instance, the deployment name of the model and the API version.
3. Updated the relevant sections of the codebase to accept and process these new arguments.
4. Update README with Azure OpenAI support

Here's a GIF that demonstrates that the model is working:
Azure OpenAI model hosting provider in e2b


ValentaTomas:
Thanks for the PR and sorry for the delay!

I just changed some details and now I'm merging it.

(The GIF especially was really welcome!)

Alaaryad:
Main

PR #51, opened 2023-06-12
Templates messages

jakubno:
# Templates for smol developer

Simple prompts to try / use for inspiration

### DIFF
--- a/api-service/Makefile
+++ b/api-service/Makefile
@@ -14,6 +14,3 @@ generate-playground-client:
 
 generate-api-schema:
 	poetry run python3 generate.py
-
-install-instrumentation:
-	poetry run opentelemetry-bootstrap -a install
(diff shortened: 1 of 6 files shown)


ValentaTomas:
I fixed merge conflicts with `production`, updated excluded extensions' list and did some general cleanup of the smol agent class.

I'll be merging this to `production` but I think we should improve the templates further:

1. Let's focus on making them smaller in scope - each template should be the minimal starting point for the project that ideally is working out of the box.
2. Let's test the templates to be sure they generate useful code
3. We are open source so let's use that - assume the least and let people add what templates they want


(Btw, I liked the web scraper for a template 👍 )

ValentaTomas:
For a very on case basis overview:
- The Rails template generated 43 files and it ran almost exactly 10 minutes.

ValentaTomas:
I'll be merging the PR so we have everything up to date on production.

PR #52, opened 2023-06-14
Supabase - New user function

jakubno:
Edge function to create a new contact in Loops.so after the new contact is created in the `users` table.

### DIFF
--- a/supabase/functions/new_user/index.ts
+++ b/supabase/functions/new_user/index.ts
@@ -1,4 +1,3 @@
-
 import { serve } from 'https://deno.land/std@0.168.0/http/server.ts'
 
 serve(async (req) => {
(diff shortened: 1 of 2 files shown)


ValentaTomas:
It looks good - I'll be merging it.

Some details that I wanted to point out:
- By default we are using "-" dashes instead of "_" underlines in the branch names
- I try to write commit messages as imperative (https://medium.com/@rahul.allada.rao/imperative-mood-for-git-commit-message-72b61418271e)
- You probably missed the first empty line in the https://github.com/e2b-dev/e2b/pull/52/files#diff-84c19cf6a71f2117e5e492665cba367a4b3a13019d0103384a724159dc91c0efR1 file

We would also ideally set up the GH Action deployment for this new Supabase function so we can be certain what is really deployed - right now we have different priorities but let's add it as soon as we have the time or the next time we are modifying the Supabase functions.

ValentaTomas:
I also discovered that we needed to add `supabase` dir to the `tsconfig.json` exclude — otherwise the `npm run build` (and the building of the app on Vercel) would fail because it would try to transpile the code of the Supabase functions too.

PR #53, opened 2023-06-15
Supabase deploy functions

jakubno:
Added a GitHub action deployment for Supabase Edge function


ValentaTomas:
If you tested it (some GH Action cases can be tested via https://github.com/nektos/act) we can merge it.
If you need to test it via a "real" GH Action you can temporarily change the branch that triggers the action and test in on this branch before merge.

PR #55, opened 2023-06-19
Supabase - Changed group name for a new user

jakubno:
I am open for other suggestions. The reason for choosing this name:
- It's descriptive
- Will be always true

### DIFF
--- a/supabase/functions/new_user/index.ts
+++ b/supabase/functions/new_user/index.ts
@@ -8,7 +8,7 @@ serve(async (req) => {
 
   const body = {
     email: data.record.email,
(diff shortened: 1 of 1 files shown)


supabase/functions/new_user/index.ts
```diff
@@ -8,7 +8,7 @@ serve(async (req) => {
 
   const body = {
     email: data.record.email,
-    userGroup: 'signUpUser'
+    userGroup: 'Supabase Authenticated'
```
jakubno:
```suggestion
    userGroup: 'Authenticated'
```

PR #75, opened 2023-08-18
Require API key to interact with e2b infra (v0.2.0)

jakubno:


packages/python-sdk/e2b/session/session_connection.py
```diff
@@ -63,9 +65,8 @@ def __init__(
         on_close: Optional[Callable[[], Any]] = None,
     ):
         if api_key is None:
-            warnings.warn(
-                "API key will be required, please see https://e2b.dev/docs",
-                FutureWarning,
+            raise AuthenticationException(
+                "API key is required, get yours at https://app.e2b.dev"
```
jakubno: Correct url
```suggestion
                "API key is required, please see https://e2b.dev/docs"
```

PR #80, opened 2023-08-22
feat(docs): Auth & API key, and Feedback

Strajk:
https://github.com/e2b-dev/e2b/assets/697301/9e0f1ab9-a788-4b0a-95a3-f2fe087c64d4

Double-check before merging:

* Sentry logging
* Posthog logging


apps/docs/src/components/Code.jsx
```diff
@@ -280,6 +208,11 @@ export function CodeGroup({ children, title, ...props }) {
 
 export function Code({ children, ...props }) {
   let isGrouped = useContext(CodeGroupContext)
+  
+  /* <DYNAMIC-API-REPLACEMENT> */
```
Strajk, line 212: commented out for now, will iterate on this later

apps/docs/src/utils/useUser.jsx
```diff
@@ -0,0 +1,139 @@
...
+  
+  useEffect(() => {
+    async function getUserCustom() {
+      const { data: teams, teamsError } = await supabase
+        .from('users_teams')
+        .select('*')
+        .eq('user_id', session?.user.id) // Due to RLS, we could also safely just fetch all, but let's be explicit for sure
+      if (teamsError) Sentry.captureException(teamsError); 
+      
+      const { data: apiKeys, error: apiKeysError } = await supabase
```
jakubno, line 76: this is okay for now. There'll be issues when we introduce teams. I would add some comment about this

apps/docs/src/components/Textarea.jsx
```diff
@@ -0,0 +1,53 @@
...
+    <textarea
+      ref={ref}
+      placeholder={placeholder}
+      value={value}
+      autoCapitalize="off"
+      autoCorrect="on"
+      autoComplete="off"
+      className={clsx(
+        'w-full',
+        'p-4',
```
jakubno: The space between the text and top looks too big to me. This is probably subjective, so feel free to ignore
```suggestion
        'py-3',
        'px-4',
```

jakubno (commented):
I am still receiving hydration error
image

..
..
..

and posthog response has some extra stuff
image

apps/docs/src/utils/useUser.jsx
Strajk: I'm returning all teams and all api keys from provider, and displaying the first API key of the default team in the component, so it should be easy to adjust e..g to Dropdown in the future

mlejva:
Overall it looks great 👍!

I would just improve the UI/UX when copying the API key. The background is too transparent when I copy the key and move my mouse away - it becomes illegible. I think we're applying only alpha channel now, maybe adding a blur would solve this. What do you think?

https://github.com/e2b-dev/e2b/assets/5136688/cacbb4de-56c1-4a44-8002-a14d6b6c9ecb

PR #91, opened 2023-08-25
WebSocket in separate thread

jakubno:


packages/python-sdk/e2b/session/websocket_client.py
```diff
@@ -0,0 +1,91 @@
...
+                    self._send_message(), name="send_message"
+                )
+                self._process_cleanup.append(send_task.cancel)
+
+                receive_task = asyncio.create_task(
+                    self._receive_message(), name="receive_message"
+                )
+                self._process_cleanup.append(receive_task.cancel)
+
+                while not self.stopped.is_set():
```
ValentaTomas, line 75: We could have probably use a normal asyncio.Event here now?

ValentaTomas: I see, now it makes sense why - you are also using the stopped event from the main loop here.

jakubno: Isn't it problem it isn't thread safe? We are setting it from the other thread

ValentaTomas: You are right - I missed that the self.stopped is passed from main loop and not something just in the ws thread.

ValentaTomas (approved):

PR #93, opened 2023-08-28
Start requiring API key for starting environment sessions

ValentaTomas:
- Added API key requirement to JS SDK
- Fixed types for the required API key in Python SDK
- Remove unused optional parameter when starting session
- Removed mangling and compression in JS SDK for better DX


ValentaTomas:
We want to start returning an error from backend when user tries to start a session without API key.
We will need to update the backend to return the error and add API keys to existing projects.

jakubno (changes requested):
Please add changeset

jakubno (approved):

PR #94, opened 2023-08-28
Improve python logging

jakubno:
```python
session = await Session.create("Nodejs", api_key="e2b_d08384135c56541ec315245d773a5ac0e8a781e3")

await session.filesystem.write("/tmp/test.txt", "Hello World!")
proc = await session.process.start(
    'while IFS= read -r line; do echo "$line"; sleep 1; done',
    rootdir="/code",
)
await proc.send_stdin("ping\n")
await session.filesystem.read("/tmp/test.txt")

await proc.kill()
await session.close()
 ```

INFO:
image

DEBUG: 
image


ValentaTomas:
I have two small concerns/ideas in mind:
- If we are logging actions like creating a file with a log before the action ("Creating file..."), should we also log after the action finished ("File created...")?
- Should we log errors that are already thrown exceptions?

jakubno:
> If we are logging actions like creating a file with a log before the action ("Creating file..."), should we also log after the action finished ("File created...")?
Good point. I added it. I was thinking if not make the initial log `INFO`. WDYT?

> Should we log errors that are already thrown exceptions?
Think about this little bit more, imho it usually doesn't make sense, if you aren't adding some extra context

ValentaTomas:
> Good point. I added it. I was thinking if not make the initial log `INFO`. WDYT?
I see why we would add this but I'm not sure. *Does any other lib do this?* If we are adding this there should be straightforward way to disable it that we tell users about.

> I thought about this little bit more, imho it usually doesn't make sense, if you aren't adding some extra context
Ok, do you think the errors should be logged as warning/debug though?

jakubno:
> > Good point. I added it. I was thinking if not make the initial log `INFO`. WDYT?
> I see why we would add this but I'm not sure. _Does any other lib do this?_ If we are adding this there should be straightforward way to disable it that we tell users about.

Let's keep it all debug then. At least for now

> > I thought about this little bit more, imho it usually doesn't make sense, if you aren't adding some extra context
> Ok, do you think the errors should be logged as warning/debug though?

I think it isn't needed when we don't catch them. If we want to add some context it should be also error

ValentaTomas:
> > > Good point. I added it. I was thinking if not make the initial log `INFO`. WDYT?
> > > I see why we would add this but I'm not sure. _Does any other lib do this?_ If we are adding this there should be straightforward way to disable it that we tell users about.
> 
> Let's keep it all debug then. At least for now
> 
> > > I thought about this little bit more, imho it usually doesn't make sense, if you aren't adding some extra context
> > > Ok, do you think the errors should be logged as warning/debug though?
> 
> I think it isn't needed when we don't catch them. If we want to add some context it should be also error

Ok, I think we can merge this then.

PR #97, opened 2023-08-29
Add timeout handling to the relevant methods in python SDK

jakubno:


packages/python-sdk/e2b/session/filesystem.py
```diff
@@ -52,43 +52,52 @@ def __init__(self, session: SessionConnection):
     #         self.service_name, "writeBase64", [path, base64_content]
     #     )
 
-    async def read(self, path: str) -> str:
+    async def read(self, path: str, timeout: int = None) -> str:
         """
         Reads the whole content of a file as an array of bytes.
 
         :param path: path to a file
+        :param timeout: Timeout for the call
```
ValentaTomas, line 60: Do we want all parameters descriptions to start with with capital letter or not?

packages/python-sdk/e2b/session/filesystem.py
```diff
@@ -52,52 +52,65 @@ def __init__(self, session: SessionConnection):
...
+        self, path: str, content: str, timeout: Optional[int] = TIMEOUT
+    ) -> None:
         """
         Writes content to a file.
 
-        :param path: path to a file
-        :param content: content to write
+        :param path: Path to a file
+        :param content: Content to write
+        :param timeout: How many seconds to wait for the server to send data before giving up (defaults to 60 seconds). Setting it to None has the effect of infinite timeout
```
ValentaTomas: I would rephrase this everywhere as:

> Specify the duration, in seconds to give the method to finish its execution before it times out (default is 60 seconds). If set to `None`, the method will continue to wait until it completes, regardless of time.

jakubno: Done

ValentaTomas (commented):
I think it looks good, I would just explicitly test if the timeout for `process.start` works ok and if the trigger exit for timeout there works.

ValentaTomas (approved):

PR #107, opened 2023-09-05
feat(js-sdk): JS SDK logging v2

Strajk:
* By default, only warn and error logs are logged to the console
* Is it possible to override the logger with custom implementation

```js
  const session = await Session.create({
    id: "Nodejs",
    apiKey: process.env.E2B_API_KEY,
    logger: {
      info: console.info, // also log info level
      warn: console.warn,
      error: console.error,
    }
  });
```

```js
  const session = await Session.create({
    id: "Nodejs",
    apiKey: process.env.E2B_API_KEY,
    logger: winston.createLogger({
      transports: [new winston.transports.File({ filename: ... })],
      level: 'debug',
    })
  });
```


packages/js-sdk/test/logging.test.mjs
```diff
@@ -0,0 +1,72 @@
+import { Session } from '../src'
+import { expect, test, vi } from 'vitest'
+import winston from 'winston';
```
jakubno: Are we adding winston just for one test?

jakubno (commented):
I would add logging also to `handleNotification` and `call` method. Otherwise it looks good to me.

packages/js-sdk/test/logging.test.mjs
Strajk: yea but only in `devDeps`, those does not count to the deps budget :D

jakubno: I don't see it there yet 😏

Strajk: i've added it to prev PR and forgot to cherry-pick it here, fixed 🙏

apps/docs/package.json
jakubno, line 1: wrong directory

jakubno (changes requested):

packages/js-sdk/src/session/sessionConnection.ts
```diff
@@ -181,8 +191,12 @@ export class SessionConnection {
     const protocol = this.opts.__debug_devEnv === 'local' ? 'ws' : 'wss'
     const sessionURL = `${protocol}://${hostname}${WS_ROUTE}`
 
-    this.rpc.onError(e => {
-      this.logger.log('Error in WS session:', this.session, e)
+    this.rpc.onError(err => {
+      this.logger.warn?.(
+        `Error in WS session "${this.session?.sessionID}": ${
+          err.message ?? err.code ?? err.toString()
+        }`,
```
jakubno: One last thing, I would add something like 
```suggestion
        `Error in WS session "${this.session?.sessionID}": ${
          err.message ?? err.code ?? err.toString()
        }. Trying to reconnect...`,
```

packages/js-sdk/test/logging.test.mjs
```diff
@@ -0,0 +1,61 @@
...
+    info: console.info,
+    warn: console.warn,
+    error: console.error,
+  }
+
+  const session = await Session.create({ id: "Nodejs", apiKey: process.env.E2B_API_KEY, logger });
+  await session.filesystem.write("test.txt", "Hello World");
+  await session.close()
+
+  expect(consoleSpy.info).toHaveBeenCalledTimes(2);
```
jakubno: I would change it to this. Otherwise adding some logs will break it
```suggestion
  expect(consoleSpy.info).toHaveBeenCalled();
```

jakubno (approved):

PR #114, opened 2023-09-05
[E2B-286] - Run release GHA only when really releasing

jakubno:


jakubno:
Done

ValentaTomas:
I originally thought that keeping everything in the same job could potentially make the workflow faster, because now you are installing Node modules every time, but we can probably just add the caching like in the release step.

Everything else looks good to me.

jakubno:
In the job `is_release` I am installing only `changeset` packages. It should be pretty fast. 

I think than more than 50% of commits to main won't trigger the release, so it make sense to have the first part as fast as possible and the little overhead for the second part won't hurt that much

ValentaTomas:
Yeah, it's probably ok (5 sec).

PR #115, opened 2023-09-06
CodeGroupAutoload

Strajk:
`````
/* BEFORE */
<CodeGroup>
```python {{ language: 'python', snippet: 'python/basics/init.py' }}
```
```js {{ language: 'js', snippet: 'js/basics/init.js' }}
```
</CodeGroup>

/* AFTER */
<CodeGroupAutoload path="basics/init" />
`````


ValentaTomas:
Super nice, you changed the order of plugins and you are adding the annotations used during the loading of snippets when the jsx gets parsed, right? 👌

Strajk:
@ValentaTomas i've added the CodeGroupAutoload as the first plugin, but kept the rest in the same order. This way we can simultaneously use the "new" syntax and also the old one for special cases.

ValentaTomas (approved):

PR #116, opened 2023-09-06
Add max session duration

jakubno:


ValentaTomas:
Closing - we are handling session/sandbox length differently.

PR #121, opened 2023-09-08
Add API key to header in SDKs

jakubno:


openapi.yml
jakubno, line 1: Should I add this? I will maybe setup git submodule from API repo, wdyt?

jakubno (commented):
Add python-dotenv to dependencies

openapi.yml
ValentaTomas: I would go with git subtree here:
- It behaves more like ordinary git files and it looks simpler to me
- The api repo is private so people may not have access to the yaml file with submodules

ValentaTomas:
I think we should be merging this to the `infra-update` branch for now and not to `main`.

ValentaTomas:
I'm also not sure if it should be `Authorization`: `Bearer ...` header (which I think is more for *user* access tokens) and not a simple `X-API-KEY` (https://swagger.io/docs/specification/authentication/api-keys/). If we are using `Authorization` header the `Basic ...` could probably be used (https://blog.stoplight.io/api-keys-best-practices-to-authenticate-apis)?

PR #122, opened 2023-09-08
Decrease websocket reconnect period

jakubno:
Not the nicest solution, but waiting for this issue, which doesn't seem to be solved soon
https://github.com/python-websockets/websockets/issues/1395


ValentaTomas:
It is possible to do 500ms reconnect like in the JS SDK?

PR #128, opened 2023-09-12
feat(docs, cli): CLI auth

Strajk:
https://github.com/e2b-dev/e2b/assets/697301/9adc4192-cc6d-42dd-bc84-caafd973699b

### DIFF
--- a/packages/cli/src/commands/auth.ts
+++ b/packages/cli/src/commands/auth.ts
@@ -66,7 +66,7 @@ export function getUserConfig(): UserConfig | null {
 async function signInWithBrowser(): Promise<UserConfig> {
   const server = http.createServer()
   const { port } = await listen(server, 0, '127.0.0.1')
-  const loginUrl = new URL(DOCS_BASE + '/docs/api/cli')
+  const loginUrl = new URL(`${DOCS_BASE}/api/cli`)
   loginUrl.searchParams.set('next', `http://localhost:${port}`)
   loginUrl.searchParams.set('cliVersion', pkg.version)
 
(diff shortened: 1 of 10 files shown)


packages/cli/src/commands/auth.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export function getUserConfig(): UserConfig | null {
+  if (!fs.existsSync(USER_CONFIG_PATH)) return null
+  return JSON.parse(fs.readFileSync(USER_CONFIG_PATH, 'utf8'))
+}
+
+async function signInWithBrowser(): Promise<UserConfig> {
+  const server = http.createServer()
+  const { port } = await listen(server, 0, '127.0.0.1')
+  const loginUrl = new URL(DOCS_BASE + '/docs/api/cli')
```
jakubno: Double `docs`, made me remember`log/blog` :D 
```suggestion
  const loginUrl = new URL(DOCS_BASE + '/api/cli')
```

packages/cli/src/commands/auth.ts
```diff
@@ -0,0 +1,100 @@
...
+  loginUrl.searchParams.set('next', `http://localhost:${port}`)
+  loginUrl.searchParams.set('cliVersion', pkg.version)
+
+  return new Promise((resolve, reject) => {
+    server.once('request', (req, res) => {
+      // Close the HTTP connection to prevent `server.close()` from hanging
+      res.setHeader('connection', 'close')
+      const followUpUrl = new URL(DOCS_BASE + '/docs/api/cli')
```
jakubno:
```suggestion
      const followUpUrl = new URL(DOCS_BASE + '/api/cli')
```

jakubno (dismissed):
There's still a lot of devbook reference/ devbook sdk imports. We don't have to do it here, but I think we should clean it up (at least readme).

packages/cli/src/commands/auth.ts
Strajk: danke!

PR #134, opened 2023-09-14
chore(js-sdk): skip flaky test blocking CI (more)

Strajk:


.changeset/mean-shoes-smile.md
jakubno, line 1: Probably not needed, last version wasn't released

PR #143, opened 2023-09-15
chore: trunk init & yamllint

Strajk:


.trunk/configs/.shellcheckrc
```diff
@@ -0,0 +1,7 @@
+enable=all
+source-path=SCRIPTDIR
+disable=SC2154
```
jakubno, line 3: I would add it and use this:
```
If you know for a fact that the variable is set, you can use ${var:?} to fail if the variable is unset (or empty)
```

PR #165, opened 2023-09-26
feat(js-sdk,py-sdk): Rename rootDir/rootdir to cwd, add cwd option to Session

Strajk:


packages/js-sdk/src/session/index.ts
```diff
@@ -228,13 +233,19 @@ export class Session extends SessionConnection {
...
-            throw new Error('cmd is required')
+          if (!cwd && rootDir) {
+            this.logger.warn?.('The rootDir parameter is deprecated, use cwd instead.')
+            cwd = rootDir
           }
-          this.logger.debug?.(`Starting process "${processID}"`)
+          if (!cwd && this.opts.cwd) {
+            cwd = this.opts.cwd
+          }
+          if (!cmd) throw new Error('cmd is required')
```
jakubno: This is little bit inconsistent. It doesn't work like this for terminal (either we should delete it or keep it maintained), nor in python,

jakubno (changes requested):

.github/workflows/js_sdk_compat_tests.yml
```diff
@@ -30,6 +30,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
+      - uses: pnpm/action-setup@v2
```
Strajk, line 33: I swear I remember adding this yesterday or a day before that 😕

packages/js-sdk/src/session/index.ts
```diff
@@ -4,8 +4,8 @@ import { components } from '../api'
 import { id } from '../utils/id'
 import { createDeferredPromise, formatSettledErrors, withTimeout } from '../utils/promise'
 import {
-  ScanOpenedPortsHandler as ScanOpenPortsHandler,
   codeSnippetService,
+  ScanOpenedPortsHandler as ScanOpenPortsHandler,
```
Strajk, line 8: just prettier being prettier

packages/python-sdk/e2b/session/main.py
```diff
@@ -159,17 +162,23 @@ async def create(
...
             on_scan_ports=on_scan_ports,
             _debug_hostname=_debug_hostname,
             _debug_port=_debug_port,
             _debug_dev_env=_debug_dev_env,
         )
         await session.open(timeout=timeout)
+        if cwd:
+            logger.info(f"Custom cwd for Session set: {cwd}")
+            proc = await session.process.start(f"mkdir -p {cwd}", cwd="/")
+            await proc
```
jakubno: What about this? 🤔 
```suggestion
            await session.filesystem.make_dir(cwd)
```

packages/js-sdk/src/session/index.ts
```diff
@@ -296,7 +310,17 @@ export class Session extends SessionConnection {
...
   static async create(opts: SessionOpts) {
-    return new Session(opts).open({ timeout: opts?.timeout })
+    return new Session(opts).open({ timeout: opts?.timeout }).then(async session => {
+      if (opts.cwd) {
+        console.log(`Custom cwd for Session set: "${opts.cwd}"`)
+        const proc = await session.process.start({
+          cmd: `mkdir -p ${opts.cwd}`,
+          cwd: '/',
+        })
+        await proc.finished
```
jakubno: And here?
```suggestion
        const proc = await session.filesystem.makeDir({path: cwd})
```

packages/python-sdk/tests/test_session.py
```diff
@@ -15,3 +15,26 @@ async def test_create_multiple_sessions():
...
+
+
+async def test_custom_cwd():
+    session = await Session.create("Nodejs", api_key=E2B_API_KEY, cwd="/code/app")
+
+    proc = await session.process.start("pwd")
+    output = await proc
+    assert output.stdout == "/code/app"
+
+    await session.filesystem.write("hello.txt", "Hello VM!")
```
jakubno, line 28: This creates a file in `/`. I am changing it in another PR, it will respect `cwd`. This will return error, if you try e.g. this `cat hello.txt`

jakubno: You can just add todo with E2B-184

packages/python-sdk/e2b/session/session_connection.py
```diff
@@ -40,6 +40,10 @@ class Subscription(BaseModel):
 class SessionConnection:
     _refresh_retries = 4
 
+    @property
+    def cwd(self):
```
jakubno, line 44: I like this!

packages/python-sdk/e2b/session/process.py
```diff
@@ -208,13 +209,14 @@ async def start(
         :param on_stderr: A callback that is called when stderr with a newline is received from the process
         :param on_exit: A callback that is called when the process exits
         :param env_vars: A dictionary of environment variables to set for the process
-        :param rootdir: The root directory for the process
+        :param cwd: The root directory for the process
+        :param rootdir: (DEPRECATED - use cwd) The root directory for the process
```
jakubno, line 213: We could add the deprecated directive, even though we aren't using sphinx atm
[deprecated directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated)
```suggestion
        :param rootdir: (DEPRECATED - use cwd) The root directory for the process
        .. deprecated:: 0.3.2
            Use cwd instead.
```

packages/python-sdk/tests/test_session.py
Strajk: ooof, very good catch!
Changed the tests for now, as it's a little unintuitive, as it should be tested in the following PR
Screen 2023-09-27 at 08 38 58

Strajk: It's confusing also Copilot :D 
Screen 2023-09-27 at 08 41 42

PR #195, opened 2023-10-13
Introduce Data Analytics template

jakubno:


out.txt
jakubno, line 1: What's this?

mlejva: mistake

mlejva (approved):
Great job

PR #198, opened 2023-10-16
Setup trunk

jakubno:


ValentaTomas:
When you get to this again check out https://publint.dev/ - it actually caught some errors when I was fixing JS SDK so it may be useful to add it.

ValentaTomas:
Closing - we will see if we are going to be adding Trunk.

PR #201, opened 2023-10-19
Feature/e2b 424 restructure docs and add data analysis

mlejva:


mlejva:
The docs aren't fully done yet but I want to merge it now and keep working on it later today

PR #202, opened 2023-10-19
Export `Artifact` class

mlejva:


jakubno:
Not sure if we want to export it from top level, if it's scoped to DataAnalysis package. You should be import it anyway

mlejva:
No need to export from the top level as we can import from the `template` module

PR #214, opened 2023-11-09
Use __future__ annotation to fix backwards compatibility for union

Kav-K:
In python 3.9 and below you can't do something like str | None in the type definition of a function's arguments. However, this can be fixed by using __future__. 

Python stores the annotations as strings and doesn't try to evaluate them. This allows you to write type hints using Python 3.10's syntax in a Python 3.9 environment. When the code is actually run, the annotations aren't evaluated, so there's no issue with the interpreter not understanding the | operator for union types.

### DIFF
--- a/.changeset/five-lemons-join.md
+++ b/.changeset/five-lemons-join.md
@@ -1,0 +1,5 @@
+---
+"@e2b/python-sdk": patch
+---
+
+Fix Python 3.8 compatibility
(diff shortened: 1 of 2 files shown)


ValentaTomas:
Good catch! I checked and the | is only in this one file, elsewhere we are using Union. I replaced the | with Unions to make it consistent.

PR #233, opened 2023-11-22
Deprecate `id` parameter when spawning `Sandbox` and add `template` parameter instead

ValentaTomas:
- [x] Change JS SDK
- [x] Change Python SDK
- [x] Change in CLI
- [x] Update docs examples
- [x] Update docs text mentions of the `id`
- [x] Decide on `id` vs `template` vs `template_id` in the e2b.toml
- [x] Test


ValentaTomas:
PR should be ready 👌

PR #234, opened 2023-11-22
Improve handler typing

ValentaTomas:
Improve on_exit handler type so it can accept functions ignoring the exit code argument without type errors.

### DIFF
--- a/packages/python-sdk/e2b/sandbox/process.py
+++ b/packages/python-sdk/e2b/sandbox/process.py
@@ -333,7 +333,7 @@ class ProcessManager:
                     if len(params) == 0:
                         on_exit()
                     elif len(params) == 1:
-                        on_exit(*[output.exit_code or 0])
+                        on_exit(output.exit_code or 0)
                     else:
                         raise TypeError(
                             f"on_exit callback must take 0 or 1 arguments, not {len(params)}"
--- a/packages/python-sdk/e2b/sandbox/process.py
+++ b/packages/python-sdk/e2b/sandbox/process.py
@@ -331,7 +331,7 @@ class ProcessManager:
                 params = sig.parameters.values()
                 try:
                     if len(params) == 0:
-                        on_exit(*[])
+                        on_exit()
                     elif len(params) == 1:
                         on_exit(*[output.exit_code or 0])
                     else:
--- a/packages/python-sdk/e2b/sandbox/process.py
+++ b/packages/python-sdk/e2b/sandbox/process.py
@@ -8,6 +8,7 @@ from typing import (
     Dict,
     List,
     Optional,
+    Union,
 )
 
 from pydantic import BaseModel
@@ -213,7 +214,7 @@ class ProcessManager:
         sandbox: SandboxConnection,
         on_stdout: Optional[Callable[[ProcessMessage], Any]] = None,
         on_stderr: Optional[Callable[[ProcessMessage], Any]] = None,
-        on_exit: Optional[Callable[[int], Any]] = None,
+        on_exit: Optional[Union[Callable[[int], Any], Callable[[], Any]]] = None,
     ):
         self._sandbox = sandbox
         self._process_cleanup: List[Callable[[], Any]] = []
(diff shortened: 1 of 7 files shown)


packages/python-sdk/e2b/sandbox/process.py
```diff
@@ -326,7 +327,7 @@ def exit_handler():
                 unsub_all()
             if on_exit:
                 try:
-                    on_exit(output.exit_code or 0)
+                    on_exit(*[output.exit_code or 0])
```
jakubno: why like this?

jakubno (approved):

packages/python-sdk/e2b/sandbox/process.py
```diff
@@ -213,7 +214,7 @@ def __init__(
         sandbox: SandboxConnection,
         on_stdout: Optional[Callable[[ProcessMessage], Any]] = None,
         on_stderr: Optional[Callable[[ProcessMessage], Any]] = None,
-        on_exit: Optional[Callable[[int], Any]] = None,
+        on_exit: Optional[Union[Callable[[int], Any], Callable[[], Any]]] = None,
```
jakubno, line 218: What can it receive? I thought it receives only exit code

ValentaTomas: You can pass method without any argument that threw type error before.

packages/python-sdk/e2b/sandbox/process.py
ValentaTomas: I have type error otherwise because the function is accepting none or one argument.

packages/python-sdk/e2b/sandbox/process.py
jakubno: Make sense 👍🏻

packages/python-sdk/e2b/sandbox/process.py
jakubno: `output.exit_code or 0` is exactly one argument

ValentaTomas: Slight correction in phrasing:

> I have a type error because the *function `on_exit` could be a function accepting none or a function accepting one argument*.

~~The spread is more like a cheat here - of course at runtime the function "accepting none" can be called with the arg, but the linter will still complain.~~

jakubno (changes requested):
Actually this won't work, if you try to pass argument to the function without arguments, it will raise exception, we would have to add check how many arguments the function accepts and based on that pass or not the exit code

ValentaTomas:
I'll fix and probably also add the exitCode to TS.

packages/python-sdk/e2b/sandbox/process.py
```diff
@@ -325,8 +327,17 @@ def exit_handler():
...
+                    if len(params) == 0:
+                        on_exit(*[])
+                    elif len(params) == 1:
+                        on_exit(*[output.exit_code or 0])
+                    else:
+                        raise TypeError(
+                            f"on_exit callback must take 0 or 1 arguments, not {len(params)}"
+                        )
                 except TypeError as error:
                     logger.exception(f"Error in on_exit callback: {error}")
```
jakubno: We can probably simplify this

packages/python-sdk/e2b/sandbox/process.py
```diff
@@ -325,8 +327,17 @@ def exit_handler():
...
                 unsub_all()
             if on_exit:
+                sig = inspect.signature(on_exit)
+                params = sig.parameters.values()
                 try:
-                    on_exit(output.exit_code or 0)
+                    if len(params) == 0:
+                        on_exit(*[])
+                    elif len(params) == 1:
+                        on_exit(*[output.exit_code or 0])
```
jakubno:
```suggestion
                        on_exit(output.exit_code or 0)
```

packages/python-sdk/e2b/sandbox/process.py
```diff
@@ -325,8 +327,17 @@ def exit_handler():
             if unsub_all:
                 unsub_all()
             if on_exit:
+                sig = inspect.signature(on_exit)
+                params = sig.parameters.values()
                 try:
-                    on_exit(output.exit_code or 0)
+                    if len(params) == 0:
+                        on_exit(*[])
```
jakubno:
```suggestion
                        on_exit()
```

PR #247, opened 2023-11-27
Add envd start cmd handling

ValentaTomas:


ValentaTomas:
We want to rethink the proper way to handle start command — even when we move the management to envd there are questions about storing start command logs and how to handle output subscriptions/reconnecting.

PR #323, opened 2024-02-28
Standardize text format and grammar across CLI commands, args, and options

mlejva:
- The hints start always with lowercase
- The hints end with a period if there's more than 1 sentence


jakubno (approved):
Add changeset and merge

PR #327, opened 2024-03-05
Add CLI command for printing logs from sandbox

ValentaTomas:
This PR adds the CLI command for printing logs for any sandbox you spawned.

Tasks:
- [x] Add flag for specifying format (json)
  - [x] Update docs
- [x] Add flag for filtering level
  - [x] Update docs
- [x] Add flag for "following"
  - [x] Update docs
  - [x] Test
- [x] Filter internal and "user" — rename logger field
  - [x] Test usability
  - [x] Update docs


spec/openapi.yml
```diff
@@ -286,6 +310,39 @@ paths:
...
+      tags: [sandboxes]
+      security:
+        - ApiKeyAuth: []
+      parameters:
+        - $ref: "#/components/parameters/sandboxID"
+        - in: query
+          name: start
+          schema:
+            type: integer
+          description: Starting timestamp of the logs that should be returned
```
jakubno: To match infra specs
```suggestion
          description: Starting timestamp of the logs that should be returned in milliseconds
```

apps/docs/src/app/cli/commands/page.mdx
```diff
@@ -241,3 +241,24 @@ Immediately kill a running sandbox.
 e2b sandbox kill <sandboxID>
 ```
 
+## `sandbox logs`
+Start printing logs from the specified sandbox.
+If the sandbox is running new logs will be streamed to the terminal.
```
jakubno, line 246: Do you think streaming by default is the best? I would add a flag to follow as everybody does.

apps/docs/src/app/cli/commands/page.mdx
```diff
@@ -241,3 +241,24 @@ Immediately kill a running sandbox.
...
+
+This command is useful if you need to debug a running sandbox or check logs from a sandbox that was already closed.
+
```
jakubno, line 251: I would add mention about the timezone

packages/cli/src/commands/sandbox/logs.ts
```diff
@@ -0,0 +1,173 @@
...
+      console.error(err)
+      process.exit(1)
+    }
+  })
+
+function printLog(timestamp: string, line: string) {
+  const log = JSON.parse(line)
+
+  const time = `[${new Date(timestamp).toISOString().replace(/T/, ' ').replace(/\..+/, '')}]`
+  let level = log['level'].toUpperCase()
```
jakubno, line 220: would be nice to have all the levels with the same width, so you have all the logs aligned

ValentaTomas: Done

ValentaTomas:
I'll also check alternative sorting of the log keys to print the `message` first.

PR #346, opened 2024-03-27
add keepAlive docs for cumulative

im-calvin:


jakubno:
Thanks @im-calvin for your PR. How `keepAlive` can be little confusing. Actually it's no cumulative, it just takes maximum of current time to live and required duration. What can make it confusing is how we internally handle inactive sandboxes. The SDK sends ping every 5 seconds, if API receive the ping it won't kill the sandbox for next 15 seconds (it's like health check and if it's 3 times missing, the sandboxed is considered abandoned).

jakubno:
It's internal, sandbox lifetime can't exceed the maximum. If you call to extend the lifetime for another hour and the sandbox is already alive for 30 minutes. It will cap it to the maximum, aka extend the lifetime just by 30 minutes.

im-calvin:
Cool thanks for clarifying saves me writing a test to figure that out myself. Is there any way of implementing that feature that I just outlined?

jakubno:
The 1 hour limit is a limit for base tier, there's possibility to go Pro tier, more info [here](https://e2b.dev/docs/pricing#pro-tier). 

Would you share what are you working on? Also are you on our discord?

PR #359, opened 2024-04-18
Add missing parentheses in `sandbox.close()` call

ggorlen:
`sandbox.close` would be a no-op without parentheses to invoke the function.


jakubno (approved):

jakubno:
Hi @ggorlen, nice catch, thanks!

PR #360, opened 2024-04-18
Use `Buffer.from()` when writing a file

ggorlen:
Running the JS [file download snippet](https://e2b.dev/docs/sandbox/api/download#use-case-for-downloading-files) in Node 20 gave me:

```
Uncaught:
TypeError [ERR_INVALID_ARG_TYPE]: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received an instance of ArrayBuffer
```

on `writeFileSync`.

But using [`Buffer.from(buffer)`](https://stackoverflow.com/a/46779188/6243352) fixed the issue and wrote the CSV file successfully. I haven't tested this extensively since I'm working with a remote client who's using the library on his end, so more validation that this change makes sense would be good before merging (this may be specific to our text file use case).

### DIFF
--- a/apps/reverse-proxy/vercel.json
+++ b/apps/reverse-proxy/vercel.json
@@ -37,6 +37,11 @@
   ],
   "redirects": [
     {
+      "source": "/careers",
+      "destination": "https://e2bdev.notion.site/Careers-at-E2B-2163f176991f43f69b0984bf2a142920",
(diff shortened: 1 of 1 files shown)


ValentaTomas:
Hey @ggorlen, thank you for catching this and opening the PR!
We will check this out 👌

PR #366, opened 2024-04-26
Update metadata.py to not use deprecated Sandbox.create() method

FanaHOVA:
Sandbox.create() is deprecated in favor of Sandbox(), updating docs example.


mlejva:
Thank you and merged!

PR #368, opened 2024-05-01
Remove `runCode` and `runCmd` from SDKs

ValentaTomas:
Remove `runCode` and `runCmd` and clean up SDKs files and docs. These features were not ergonomic and we released a new package [code-interpreter](https://github.com/e2b-dev/code-interpreter) that handles this better.

Tasks:
- [x] Fix 404 errors for the deleted guide


apps/docs/src/app/sandbox/api/execute/page.mdx
ValentaTomas, line 1: We should mention the code-interpreter package here.

ValentaTomas: It looks like this page is also not used, so we might just delete it.

PR #373, opened 2024-05-08
Fix websocket compatibility for Deno and Bun

ValentaTomas:
This PR uses isomorphic WS package supporting Deno and Bun. The WS was the only thing blocking the compatibility with Deno and Bun.

Tasks:
- [ ] Test with cloudflare workers/edge
- [ ] Add test for bun, deno, worker compatibility
- [ ] Test in browser


ValentaTomas:
Out beta SDK (https://github.com/e2b-dev/E2B/pull/389) should resolve problems with Bun and Deno compatibility.

PR #374, opened 2024-05-09
Enforce calling create method in JS SDK

jakubno:
This prevents to accidentally use `new Sandbox()`, which doesn't initialize `Sandbox` class correctly

### DIFF
--- a/packages/js-sdk/src/sandbox/index.ts
+++ b/packages/js-sdk/src/sandbox/index.ts
@@ -78,9 +78,9 @@ export class Sandbox extends SandboxConnection {
    * @internal
    * @access protected
    */
-  constructor(opts?: SandboxOpts) {
+  constructor(opts?: SandboxOpts, protected createCalled: boolean = false) {
     opts = opts || {}
-    super(opts)
+    super(opts, createCalled)
     this.onScanPorts = opts.onScanPorts
 
     // Init Filesystem handler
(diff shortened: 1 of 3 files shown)


packages/js-sdk/src/sandbox/sandboxConnection.ts
```diff
@@ -111,7 +111,11 @@ export class SandboxConnection {
   private readonly client: APIClient
 
   // let's keep opts readonly, but public - for convenience, mainly when debugging
-  protected constructor(readonly opts: SandboxConnectionOpts) {
+  protected constructor(readonly opts: SandboxConnectionOpts, protected createCalled: boolean) {
+    if (!createCalled) {
+      throw new Error('Cannot instantiate Sandbox directly, use a `create()` method instead.')
```
mlejva: I'd frame it as this
```js
throw new Error('Cannot instantiate Sandbox directly. Use `const ci = await CodeInterpreter.create()` to get a new code interpreter instance.')
```
what do you think?

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -78,9 +78,9 @@ export class Sandbox extends SandboxConnection {
    * @internal
    * @access protected
    */
-  constructor(opts?: SandboxOpts) {
+  constructor(opts?: SandboxOpts, protected createCalled: boolean = false) {
```
ValentaTomas, line 81: Add `protected`. Check also the protected constructor in `code-interpreter`.

packages/js-sdk/src/sandbox/sandboxConnection.ts
jakubno: This will be an error message for a standard sandbox and all extended classes (e.g. CodeIntrepreter), we can override this for CodeIntrepreter, but it would probably require quite a lot of boilterplate code. That's why I wanted to omit the class name

jakubno: what about something like this?
```
throw new Error('Cannot instantiate sandbox directly, use a `.create()` method instead of calling `new` to get a new sandbox.\ne.g. instead of `new CodeInterpreter(...)` do `await CodeInterpreter.create(...)`')

```

packages/js-sdk/src/sandbox/index.ts
jakubno: I created issue #375

PR #384, opened 2024-06-11
Improve SDK communication with sandbox and Sandbox lifecycle handling

ValentaTomas:
Switch the communication between SDKs and Sandboxes from WebSocket and get rid of the lifecycle management via periodic refresh in the background in favor of setting an explicit timeout.


ValentaTomas:
We opened a new PR that includes changes from this one — https://github.com/e2b-dev/E2B/pull/389

PR #389, opened 2024-06-27
SDK V1 Beta

ValentaTomas:
This PR introduces changes to the SDKs that:
- Fix the lifecycle management by letting users specify sandbox timeout and getting rid of periodic refreshes in the background
- Changes communication from WebSocket to HTTP1.1 with streams which removes the need for establishing the initial connection and allows better control over requests to the sandbox
- Improves some long-standing issues with the SDK like default user
- Introduces requested methods like `exists` or `rename`


apps/web/prebuild.js
jakubno, line 1: Convert to TS?

packages/cli/package.json
```diff
@@ -55,7 +57,7 @@
     "command-exists": "^1.2.9",
     "commander": "^11.1.0",
     "console-table-printer": "^2.11.2",
-    "e2b": "^0.16.2",
+    "e2b": "^0.16.2-beta.44",
```
jakubno, line 61: Change after release

PR #393, opened 2024-06-30
Use docker API instead of cmd

mishushakov:
Instead of executing the docker commands in the code, use the Docker API to communicate with the Docker server.


packages/cli/package.json
```diff
@@ -24,13 +24,17 @@
     "build": "tsc --noEmit --skipLibCheck && tsup --minify",
     "dev": "tsup --watch",
     "test": "pnpm build && cd testground/demo-basic && ../../dist/index.js template build",
+    "test:simple": "pnpm build && cd testground/demo-simple && ../../dist/index.js template build",
+    "test:failing": "pnpm build && cd testground/demo-failing && ../../dist/index.js template build",
     "check-deps": "knip",
     "update-deps": "ncu -u && pnpm i"
   },
   "devDependencies": {
     "@types/command-exists": "^1.2.3",
```
ValentaTomas, line 33: This types should be redundant after removing the actual package.

packages/cli/src/commands/template/build.ts
```diff
@@ -261,47 +264,53 @@ export const buildCommand = new commander.Command('build')
...
-            DOCKER_CLI_HINTS: 'false',
-          },
+
+        const dockerImageTag = `docker.${e2b.SANDBOX_DOMAIN}/e2b/custom-envs/${templateID}:${template.buildID}`
+        const buildStream = await docker.buildImage(await prepareDockerContext(root), {
+          t: dockerImageTag,
+          platform: 'linux/amd64',
+          buildargs: dockerBuildArgs,
+          dockerfile: dockerfileRelativePath,
         })
```
ValentaTomas: After some brief tests (dockerignore, nested dir in context) I think it might be possible to just pass root and list of top level files (from https://github.com/apocas/dockerode/blob/49bd8c4c231efe9bb432b28d51dbae3e50ddd010/test/docker.js#L174 it looks like they are already handling this internally) instead of using tar-fs + filtering.

```suggestion
        const buildStream = await docker.buildImage({
          context: root,
          src: fs.readdirSync(root),
        }, {
          t: dockerImageTag,
          platform: 'linux/amd64',
          buildargs: dockerBuildArgs,
          dockerfile: dockerfileRelativePath,
        })
```

ValentaTomas:
Also if the terminal is too narrow the printing of logs (here ubuntu pull) gets split into lines.

Screenshot 2024-07-02 at 20 32 50

ValentaTomas:
Btw, the `"@types/command-exists"` is still in `package.json` even after the actual package was removed.

mishushakov:
Hey @ValentaTomas, have implemented your suggestions and added some cherries on top. Removed the blue colouring to reduce eye-strain and the progress bar should now be truncate when the terminal becomes too narrow.

Screenshot 2024-07-08 at 12 50 27

ValentaTomas:
Just a note — we will need to add the equivalent of `--pull` to the docker build, because otherwise making sure users build the custom templates with the latest E2B Docker image is hard.

jakubno:
Moving to a branch in this repo to allow for running a release candidate pipeline

PR #401, opened 2024-07-17
Fix docs links

mlejva:
The recent changes in our reverse proxy broke links in the docs because the old docs link didn't need to have the /docs prefix. Now they do.


apps/docs/src/app/(dashboard)/dashboard/page.tsx
```diff
@@ -123,6 +123,7 @@ export default function Dashboard() {
         <div className="flex-1 md:pl-10">
           <h2 className='text-2xl mb-2 font-bold'>{selectedItem[0].toUpperCase() + selectedItem.slice(1)}</h2>
           <div className='border border-white/5 w-full h-[1px] mb-10' />
+          {/* @ts-ignore */}
```
jakubno: I don't see why is this needed

PR #404, opened 2024-07-17
Fix links from apps/doc to apps/web

shreyas-sreedhar:
Corrected documentation links in the README.md for "Documentation" and utils/const.ts.

Review Required for `pnpm run dev:docs`: The dev:docs script in package.json currently links to apps/docs and cannot be run locally unless it is changed to apps/web. 

Didn't want to change the package.json right away, thought i'd let you know about this first instead.


jakubno:
Hi @shreyas-sreedhar, thanks for catching this! I will change `package.json` as well

PR #423, opened 2024-08-13
Add `download_url`/`downloadUrl` method to sandbox

ValentaTomas:
If you want to download a file from sandbox you need to use the SDK right now.
The download URL method we are adding lets you get the link to a file in sandbox that you can then use to download the file, for example on the frontend, with a direct link.

### Tasks
- [ ] Test the changes (JS+TS SDKs)
- [ ] Update Core SDK in Code Interpreter so this functionality is enabled there too


jakubno (approved):
Lgtm

PR #432, opened 2024-08-25
Docker SDK Usage

danstarns:
# What? 

- Refactors the usage of spawning docker CLI directly using `child_process`
    * see https://github.com/e2b-dev/E2B/blob/main/packages/cli/src/commands/template/build.ts#L271
    * see https://github.com/e2b-dev/E2B/blob/main/packages/cli/src/commands/template/build.ts#L286
    * see https://github.com/e2b-dev/E2B/blob/main/packages/cli/src/commands/template/build.ts#L304    
- Adds testing layer on top of the template build command
- Refactors the template build file for approachability and testing with mocks.

# Why 
Using `child_process` can be brittle and there are already tested tools that allow us to interact with Docker programmatically, we should use those for better readability and maintainability. 

The CLI had few tests, so to make assertive changes here, I first added coverage for the CLI and then returned to make the dockerode / SDK changes. 

# How

Usage of npm package https://github.com/apocas/dockerode replaces `child_process` calls. 

Vitest is added under `pnpm test:unit` in the CLI package, mocks are used to interact with and assert the CLI. 

Refactors have been made for readability and to enable the mocking of modules in tests.


packages/cli/package.json
```diff
@@ -55,9 +59,11 @@
     "command-exists": "^1.2.9",
     "commander": "^11.1.0",
     "console-table-printer": "^2.11.2",
+    "dockerode": "^4.0.2",
     "e2b": "^0.16.2",
     "inquirer": "^9.2.12",
     "open": "^9.1.0",
+    "ssh2": "^1.15.0",
```
danstarns, line 67: Added as a peer of `dockerode`, resolves issue in testing: 

```
Can't resolve './crypto/build/Release/sshcrypto.node'
```

Related: 
* https://github.com/apocas/dockerode/issues/775

packages/cli/src/commands/template/utils.ts
```diff
@@ -0,0 +1,118 @@
+import * as e2b from 'e2b'
```
danstarns, line 1: This file was moved from `/template/build.js` for readability and mocking in the added tests.

packages/cli/tsup.config.js
```diff
@@ -13,7 +13,7 @@ export default defineConfig({
   sourcemap: true,
   clean: true,
   noExternal: Object.keys(packageJSON.dependencies).filter(
-    f => !excludedPackages.includes(f)
+    (f) => !excludedPackages.includes(f)
```
danstarns, line 16: It seems that if I use either: 

1. vscode prettier
2. `npx prettier --fix`
3. `npx eslint --fix` 

They all are formatting the code in different ways, is this known, or a problem with my local environment?

packages/cli/test/template.test.ts
```diff
@@ -0,0 +1,158 @@
+import { test, expect, describe, vi, afterEach } from 'vitest'
```
danstarns, line 1: These tests can be executed in the `packages/cli` directory using `pnpm test:unit` command, they use vitest to mock the commander template build command. The final test "should perform a build" will mock all points of the build, from requesting and saving a template to pushing and building with docker.

PR #441, opened 2024-09-17
Auto-generate API reference markdown files for the python-sdk

0div:
# Description

- Installed and configured [pydoc-markdown](https://niklasrosenstein.github.io/pydoc-markdown/).
- Created a script that generates and cleans markdown files with a configurable package list.
- Added docstrings to relevant public methods of the various sdk classes in order for pydoc to detect them.

# Test

```bash
cd packages/python-sdk
./scripts/generate_api_ref.sh 
ls ./api_ref
```

# Considerations

This could eventually be used directly as `.mdx` in the docs nextjs app, it would look like this with current styling:
Screenshot 2024-09-17 at 17 22 43


packages/python-sdk/scripts/generate_api_ref.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
```
jakubno, line 2: add `set -euo pipefail`, we want to use this in cicd, so it would be nice to know it failed

packages/python-sdk/scripts/generate_api_ref.sh
```diff
@@ -0,0 +1,19 @@
...
+mkdir api_ref
+
+for package in "${packages[@]}"; do
+    # generate raw api reference markdown file
+    pydoc-markdown -p e2b.${package} >api_ref/${package}.mdx
+    # remove package path display
+    sed -i '' '/<a[^>]*>.*<\/a>/d' api_ref/${package}.mdx
+    # remove empty hyperlinks
+    sed -i '' '/^# /d' api_ref/${package}.mdx
+    # move to docs (for later use)
```
jakubno, line 19: is there anything wrong with it right now? Also feel free to add the generated files. Could you also add the route to `.gitattributes` 🙏🏻 (feel free to ignore this if the route isn't decided yet)

packages/python-sdk/scripts/generate_api_ref.sh
0div: yeah always good to use it, i'm not sure if this will be the final script once i piece it all together in workflow but ill add this nevertheless

packages/python-sdk/scripts/generate_api_ref.sh
0div: I remove the package paths bc `pydoc-markdown` includes all paths, even for packages that are not documented

0div: and the naming becomes redundant with this as well IMO

packages/python-sdk/e2b/sandbox_sync/main.py
```diff
@@ -209,6 +213,7 @@ def set_timeout(  # type: ignore
         timeout: int,
         request_timeout: Optional[float] = None,
     ) -> None:
+        """Configure request timeout"""
```
ValentaTomas: This does configure the *sandbox* timeout.
We might consider changing the method's name if the use is not apparent though.

packages/python-sdk/e2b/sandbox_sync/main.py
```diff
@@ -34,24 +29,29 @@ def handle_request(self, request):
 class Sandbox(SandboxSetup, SandboxApi):
     @property
     def files(self) -> Filesystem:
+        """Get a Filesystem Object"""
```
ValentaTomas, line 32: I think some of the docstrings could be improved, but the PR is for adding the API generation — I think we needed to add just some brief descriptions so there is something to generate, right?

packages/python-sdk/e2b/sandbox_sync/process/main.py
```diff
@@ -138,6 +137,7 @@ def run(
         timeout: Optional[float] = 60,
         request_timeout: Optional[float] = None,
     ):
+        """Run command"""
```
ValentaTomas, line 140: Will the API generation work with the overloads? Maybe just having the base method with all the formats enumerated is enough — in some cases like `commands.run` where the overload changes the behavior it might be tricky because the docstring for each overload explains the behavior.

packages/python-sdk/e2b/sandbox_sync/main.py
0div: Got it, something like `set_sandbox_timeout` ? In the meanwhile i updated the docstring

packages/python-sdk/e2b/sandbox_sync/main.py
0div: That's correct, a docstring flags it for being documented. I can focus on improving the descriptions once I have all the packages in the auto api-ref generation pipeline. Tracking it [E2B-799](https://linear.app/e2b/issue/E2B-799/write-uniform-and-improved-api-ref-descriptions-across-packages)

packages/python-sdk/e2b/sandbox_sync/process/main.py
0div: The way it generates the docs is on a per package bases, so part of the targets are `packages/python-sdk/e2b/sandbox_{async, sync}` , as long as it is not overloaded inside those directories there should be no conflict.

mlejva (approved):

PR #447, opened 2024-09-18
Auto-generate API reference markdown files for the js-sdk

0div:
# Description 
- Installed and configured [typedoc](https://github.com/TypeStrong/typedoc) and its plugin [typedoc-plugin-markdown](https://github.com/typedoc2md/typedoc-plugin-markdown)
- Created a script that generates and cleans markdown files

# Test
```bash
cd packages/js-sdk
./scripts/generate_api_ref.sh 
ls ./api_ref
```

# Considerations
This could eventually be used directly as .mdx in the docs nextjs app, it would look like this with current styling and TOC:

Screenshot 2024-09-18 at 17 29 22

Screenshot 2024-09-18 at 17 29 32

Screenshot 2024-09-18 at 17 29 45

per my previous slack messages:
> typedoc and its typedoc-plugin-markdown are more opinionated than i assumed.
The extent of how detailed it documents things about classes and interfaces is less configurable than i thought, for example it forces you to display all inheritance info (there's a plugin to remove that but it's clunky) . Also, linking is not configurable when you are converting to markdown, it seems to keep routing as if it were a static website directory with relative links so i removed all links.
Had to create a hacky post-processor to handle these obstacles.

i wish i could figure out a way to use anchor links, i looked into [typedoc-plugin-pages](https://github.com/KnodesCommunity/typedoc-plugins/tree/develop/packages/plugin-pages) but couldn't get it to behave the way I want.


mlejva:
Overall, this is a good start! We'll get a true feeling of it once we start using it in some v1 on the docs website.

> i wish i could figure out a way to use anchor links, i looked into [typedoc-plugins](https://github.com/KnodesCommunity/typedoc-plugins) but couldn't get it to behave the way I want.

Let's figure out a way to do this. This is a really big value-add for developers.

PR #448, opened 2024-09-20
Auto-generate API reference markdown files for the CLI

0div:
# Description
-  upgraded the `-cmd2json` root option to `-cmd2md`
-  it calls a `commands2md` utils custom function that converts command objects to Markdown documentation
- this function writes all root command documentation files into the `api_ref/` 
- boyscouting: added root `.prettierrc`

# Test
```bash
cd packages/cli
./scripts/generate_api_ref.sh 
ls ./api_ref
```

# Considerations
This could eventually be used directly as .mdx in the docs nextjs app, it would look like this with current styling and TOC:
Screenshot 2024-09-20 at 11 47 42
Screenshot 2024-09-20 at 11 47 34
Screenshot 2024-09-20 at 12 14 52

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -16,15 +16,19 @@ updateNotifier
   })
   .notify()
 
-const prog = program
-  .version(packageJSON.version, undefined, 'display E2B CLI version')
-  .addOption(new commander.Option('-cmd2md').hideHelp())
+const prog = program.version(
+  packageJSON.version,
+  undefined,
+  'display E2B CLI version'
+)
 
 if (process.env.NODE_ENV === 'development') {
-  prog.on('option:-cmd2md', () => {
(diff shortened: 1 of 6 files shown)


packages/cli/src/index.ts
```diff
@@ -18,25 +19,9 @@ updateNotifier
...
-              description: stripAnsi.default(y.description),
-              defaultValue: y.defaultValue,
-            })),
-          }))
-          .sort((row1: any, row2: any) =>
-            row1.name().localeCompare(row2.name()),
-          ),
-      ),
-    )
+  .addOption(new commander.Option('-cmd2md').hideHelp())
```
jakubno: I am just thinking, would it make sense to have this only in dev and not in prod version?

jakubno (approved):

PR #451, opened 2024-10-04
Add multi-file write support to the js and python sdks

0div:
# Description

- Allow the `Filesystem.write` method to accept multiple files 
- Add tests to assert that `envd` supports multipart with multiple files out of the box 

# Test
```
# Test js-sdk
cd packages/js-sdk
pnpm test

# Test python-sdk
cd packages/python-sdk
pnpm test
```


packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -34,6 +21,8 @@ export const enum FileType {
   DIR = 'dir',
 }
 
+export type WriteData = string | ArrayBuffer | Blob | ReadableStream
```
ValentaTomas: If the type is exposed to the user, we should export it from the whole package.

packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -95,8 +80,17 @@ export class Filesystem {
     return res.data
   }
 
-  async write(path: string, data: string | ArrayBuffer | Blob | ReadableStream, opts?: FilesystemRequestOpts): Promise<EntryInfo> {
-    const blob = await new Response(data).blob()
+  async write(path: string, data: WriteData | { data: WriteData; filename: string }[], opts?: FilesystemRequestOpts): Promise<EntryInfo | EntryInfo[]> {
```
ValentaTomas: I think we should create separate overloads for this method, taking one and multiple files because:

1. This way, you will get an array (in typing) as a return instead of an array or single entry
2. The `path` parameter that is always present right now is very confusing when you are uploading multiple files

ValentaTomas:
I pushed an edit that should sketch a type cleanup.

There are two unfinished parts:
1.
```ts
    const { path, writeOpts, writeFiles } = typeof pathOrFiles === 'string'
      ? { path: pathOrFiles, writeFiles: [{ data: dataOrOpts }], writeOpts: opts }
      : { path: undefined, writeFiles: pathOrFiles, writeOpts: dataOrOpts }

    const blobs = await Promise.all(writeFiles.map(f => new Response(f.data).blob()))

```
It should be possible to typeguard this so there is no type assumption when dealing with the parameters.

2.
```ts
    return files.length === 1 ? files[0] : files
```
The problem here was actually there even before — the `EntryInfo` type assumption works because the returned types are exactly of the shape of `EntryInfo`, but I think we should property type these by creating a mapping.

Also, your previous code was correct; this is more of an improvement than a fix.

ValentaTomas:
I also think the `path` might be redundant as an API argument because we can just send the file with the filename as usual. It might be hard to remove it from envd right now, but we can stop using it in SDKs so we can remove it later.

0div:
@ValentaTomas I addressed your comments and added extra tests for some edge cases.

0div:
One important thing to note, and I've added it as a code comment, is that we can't expect specified directories in path of  multipart filename to be taken into consideration; I've tested it with and sure enough only file name is used as path, the rest of the path is stripped by the std lib's `"mime/multipart"` when calling
```
pathToResolve = part.FileName()
```

ValentaTomas:
> One important thing to note, and I've added it as a code comment, is that we can't expect specified directories in path of multipart filename to be taken into consideration; I've tested it with and sure enough only file name is used as path, the rest of the path is stripped by the std lib's `"mime/multipart"` when calling
> 
> ```
> pathToResolve = part.FileName()
> ```

Ok, this is a very good find — how do you think we should handle this? We want to be able to upload files with any path, but the stripping of paths might make sense to preserve, because it allows people to upload by link easily. This might require some changes to the envd.

0div:
> > One important thing to note, and I've added it as a code comment, is that we can't expect specified directories in path of multipart filename to be taken into consideration; I've tested it with and sure enough only file name is used as path, the rest of the path is stripped by the std lib's `"mime/multipart"` when calling
> > ```
> > pathToResolve = part.FileName()
> > ```
> 
> Ok, this is a very good find — how do you think we should handle this? We want to be able to upload files with any path, but the stripping of paths might make sense to preserve, because it allows people to upload by link easily. This might require some changes to the envd.

If it's really important not to break the current API spec, we could add a custom field to the multipart dataform and look for it with some changes to `envd`

0div:
@ValentaTomas I started by adding multi file write support for `sanbdbox_sync` to get your opinion on this approach before moving forward with `_async`. 

### Considerations
I went the `OR` route when it comes to typing, as function overload is not natively supported in python. Thate being said, with some [non-negligible overhead](https://www.codementor.io/@arpitbhayani/overload-functions-in-python-13e32ahzqt), it is kindof achievable, but im not enthralled by the idea, especially when thinking about previous work generating the api-ref automatically: i have a feeling it wouldn't play well with it.

ValentaTomas:
For the overload I think you can use the same system as we already have with https://github.com/e2b-dev/E2B/blob/beta/packages/python-sdk/e2b/sandbox_sync/process/main.py#L106

ValentaTomas:
I also suggest naming and exporting all the types (from both SDKs). What do you say about having:

- `WriteData` — the is then union type for the data (string, bytes, etc)
- `WriteEntry` — this is the type that contains both the `path` and `data: WriteData` fields.

ValentaTomas:
I'm thinking about what to do when you try to invoke write for multiple files and provide an empty array.

Logically, you might want to notify the user that nothing was written, but throwing an error might not be optimal. If you are generating the field to write, you need to explicitly check if the array is empty; otherwise, you will get an error.

In contrast to this, isn't writing 0 files a valid operation and you will also get an array with 0 results so everything is ok?

ValentaTomas:
> > > One important thing to note, and I've added it as a code comment, is that we can't expect specified directories in path of multipart filename to be taken into consideration; I've tested it with and sure enough only file name is used as path, the rest of the path is stripped by the std lib's `"mime/multipart"` when calling
> > > ```
> > > pathToResolve = part.FileName()
> > > ```
> > 
> > 
> > Ok, this is a very good find — how do you think we should handle this? We want to be able to upload files with any path, but the stripping of paths might make sense to preserve, because it allows people to upload by link easily. This might require some changes to the envd.
> 
> If it's really important not to break the current API spec, we could add a custom field to the multipart dataform and look for it with some changes to `envd`

Yeah, I'm thinking that we should probably do this, because people are already using the Beta SDK.

0div:
> I'm thinking about what to do when you try to invoke write for multiple files and provide an empty array.
> 
> Logically, you might want to notify the user that nothing was written, but throwing an error might not be optimal. If you are generating the field to write, you need to explicitly check if the array is empty; otherwise, you will get an error.
> 
> In contrast to this, isn't writing 0 files a valid operation and you will also get an array with 0 results so everything is ok?

From the perpective that there will likely be less control over which files, if any, are generated, your point makes sense. I will allow empty arrays

packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -21,7 +15,8 @@ import { EnvdApiClient } from '../../envd/api'
 import { Filesystem as FilesystemService } from '../../envd/filesystem/filesystem_connect'
 import { FileType as FsFileType } from '../../envd/filesystem/filesystem_pb'
 
-import { WatchHandle, FilesystemEvent } from './watchHandle'
+import { clearTimeout } from 'timers'
```
ValentaTomas: The clearTimeout should be a built in JS method that does not need to be imported.

packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -141,40 +125,44 @@ export class Filesystem {
...
-   *   When writing to a file that already exists, the file will get overwritten.
-   *   When writing to a file that's in a directory that doesn't exist, you'll get an error.
-   *
-   * @param path Path to a new file. For example '/dirA/dirB/newFile.txt' when creating 'newFile.txt'
-   * @param data Data to write to a new file
-   * @param opts Options for the request
-   * @returns Information about the written file
-   */
+  async write(path: string, data: WriteData, opts?: FilesystemRequestOpts): Promise<EntryInfo>
+  async write(files: { path: string; data: WriteData }[], opts?: FilesystemRequestOpts): Promise<EntryInfo[]>
```
ValentaTomas: I think we should add `WriteEntry` also here.

packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -141,40 +125,44 @@ export class Filesystem {
...
         query: {
           path,
-          username: opts?.user || defaultUsername,
+          username: (writeOpts as FilesystemRequestOpts)?.user || defaultUsername,
         },
       },
       bodySerializer() {
-        const fd = new FormData()
+        return blobs.reduce((fd, blob, i) => {
+          // Important: RFC 7578, Section 4.2 requires that if a filename is provided,
```
ValentaTomas, line 305: Is there a way to add directory path that is compliant with standard spec?

packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -141,40 +125,44 @@ export class Filesystem {
...
+      throw new Error('Cannot specify path with array of files')
+    }
+
+    const { path, writeOpts, writeFiles } =
+      typeof pathOrFiles === 'string'
+        ? { path: pathOrFiles, writeOpts: opts, writeFiles: [{ data: dataOrOpts }] }
+        : { path: undefined, writeOpts: dataOrOpts, writeFiles: pathOrFiles }
+
+    const blobs = await Promise.all(
+      (writeFiles as { path: string; data: WriteData }[]).map((f) => new Response(f.data).blob())
```
ValentaTomas: It might not be possible but I'm still thinking is we need to use `as` anywhere — could the typeguarding get rid of the need so we can be sure the type is correct without `as`suming. This is not blocking, just would make it extra nice.

packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -183,11 +171,11 @@ export class Filesystem {
     }
 
     const files = res.data
-    if (!files || files.length === 0) {
+    if (!files) {
       throw new Error('Expected to receive information about written file')
     }
 
-    return files[0] as EntryInfo
+    return files.length === 1 && path ? (files[0] as EntryInfo) : (files as EntryInfo[])
```
ValentaTomas: Right not the type of files/files[0] is structurally the same as the `EntryInfo` and we are relying on that without explicitly checking. This is also not blocking right now, but something I wanted to point out.

packages/python-sdk/e2b/exceptions.py
```diff
@@ -30,10 +30,10 @@ class TimeoutException(SandboxException):
     """
     Raised when a timeout occurs.
 
-    The [unavailable] exception type is caused by sandbox timeout.\n
-    The [canceled] exception type is caused by exceeding request timeout.\n
-    The [deadline_exceeded] exception type is caused by exceeding the timeout for process, watch, etc.\n
-    The [unknown] exception type is sometimes caused by the sandbox timeout when the request is not processed correctly.\n
+    The `unavailable` exception type is caused by sandbox timeout.\n
```
ValentaTomas, line 33: There is similar comment in TS SDK like this so we might want to change it there too.
Overall we need to improve how we communicate the errors, so I'm not sure if we should be touching this right now.

packages/js-sdk/src/sandbox/filesystem/index.ts
0div: ah snap, this is my import sorter extension doing this

packages/js-sdk/src/sandbox/filesystem/index.ts
0div: > Based on the RFC 7578 specification and the search results provided, you cannot specify a custom directory path in the header or body of a multipart/form-data request

and more generally 

> According to RFC 7578, it is generally not possible to specify custom fields in the header or body of a multipart/form-data submission beyond what is explicitly defined in the specification.

from https://www.perplexity.ai/search/according-to-this-rfc-https-da-2Fy4OAKDTmy8tgR6DLdGnA

packages/js-sdk/src/sandbox/filesystem/index.ts
0div: I moved the `as` in the destructuring

packages/js-sdk/src/sandbox/filesystem/index.ts
0div: In the meanwhile i moved the `as` up to `res.data`

packages/python-sdk/e2b/exceptions.py
0div: Ok i made this change based on feedback, but we should keep it consistent so i will edit the ts comment

packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -141,53 +129,61 @@ export class Filesystem {
...
+
+    const { path, writeOpts, writeFiles } =
+      typeof pathOrFiles === 'string'
+        ? {
+            path: pathOrFiles,
+            writeOpts: opts as FilesystemRequestOpts,
+            writeFiles: [{ data: dataOrOpts as WriteData }],
+          }
+        : { path: undefined, writeOpts: dataOrOpts as FilesystemRequestOpts, writeFiles: pathOrFiles as WriteEntry[] }
+
```
ValentaTomas, line 289: We can probably short-circuit here for an empty array of files.

packages/python-sdk/e2b/sandbox_sync/filesystem/filesystem.py
```diff
@@ -122,27 +124,85 @@ def write(
...
+        Writes content to a file on the path.
+        When writing to a file that doesn't exist, the file will get created.
+        When writing to a file that already exists, the file will get overwritten.
+        When writing to a file that's in a directory that doesn't exist, you'll get an error.
+        """
+
+        path, write_files, user, request_timeout  = None, [], "user", None
+        if isinstance(path_or_files, str):
+            if isinstance(data_or_user, list):
+                raise Exception("Cannot specify path with array of files")
```
mlejva: Let's change this error string to be more explicit what user is doing wrong here:
```
Cannot specify both path and array of files.
```

maybe we could be even more helpful here:
```
Cannot specify both path and array of files. You have to specify either path and data for a single file or an array for multiple files.
```

packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -141,53 +129,61 @@ export class Filesystem {
...
-    path: string,
-    data: string | ArrayBuffer | Blob | ReadableStream,
+    pathOrFiles: string | WriteEntry[],
+    dataOrOpts?: WriteData | FilesystemRequestOpts,
     opts?: FilesystemRequestOpts
-  ): Promise<EntryInfo> {
-    const blob = await new Response(data).blob()
+  ): Promise<EntryInfo | EntryInfo[]> {
+    if (typeof pathOrFiles === 'string' && Array.isArray(dataOrOpts)) {
+      throw new Error('Cannot specify path with array of files')
```
mlejva: Let's change this error string to be more explicit what user is doing wrong here:
```
Cannot specify both path and array of files.
```

maybe we could be even more helpful here:
```
Cannot specify both path and array of files. You have to specify either path and data for a single file or an array for multiple files.
```

packages/js-sdk/tests/sandbox/files/write.test.ts
```diff
@@ -1,12 +1,120 @@
...
+  // Attempt to write with undefined path and file array
+  await sandbox.files
+    // @ts-ignore
+    .write(undefined, [{ path: 'one_test_file.txt', data: 'This is a test file.' }])
+    .then((e) => {
+      assert.isUndefined(e)
+    })
+    .catch((err) => {
+      assert.instanceOf(err, Error)
+      assert.include(err.message, 'Cannot read properties of undefined')
```
mlejva: Can we explicitly mention in the error message that a user forgot to specify a path?

packages/python-sdk/tests/async/sandbox_async/files/test_write.py
```diff
@@ -1,10 +1,16 @@
...
 
 async def test_write_file(async_sandbox: AsyncSandbox):
     filename = "test_write.txt"
     content = "This is a test file."
 
+    # Attempt to write without path
+    try:
+        await async_sandbox.files.write(None, content)
+    except Exception as e:
+        assert "object is not iterable" in str(e)
```
mlejva: Can we explicitly mention in the error message that a user forgot to specify a path?

mlejva:
@ValentaTomas @0div 

1. I left a couple of notes about the DX.
2. It's not completely clear to me how the usage looks like, @0div can you please share code snippets for JS and Python showing how to use the new API? I think especially in Python naming of these parameters in the `write` method is confusing
```python
        path_or_files: str | List[WriteEntry],
        data_or_user: WriteData | Username = "user",
        user_or_request_timeout: Optional[float | Username] = None,
```

0div:
> @ValentaTomas @0div
> 
> 2. It's not completely clear to me how the usage looks like, @0div can you please share code snippets for JS and Python showing how to use the new API? I think especially in Python naming of these parameters in the `write` method is confusing
> 
> ```python
>         path_or_files: str | List[WriteEntry],
>         data_or_user: WriteData | Username = "user",
>         user_or_request_timeout: Optional[float | Username] = None,
> ```

### js-sdk
```js
// legacy single-file write method signature
sandbox.files.write("path/to/file.txt", "This is a test file")

// new overloaded multi-file write method signature:
sandbox.files.write([
  { path: 'test_file_1.txt', data: 'This the first test file.' },
  { path: '/test_dir/test_file_2.txt', data: 'This the second test file.'},
])
```
see [js-sdk test](https://github.com/e2b-dev/E2B/pull/451/files#diff-7a70570a2ea87cb7d3e355b0aa6f011a4bec4d1733f40bf94b05e4024f232b01R35) for more examples

### python-sdk (sync & async)
```py
# legacy single-file write method signature
sandbox.files.write("path/to/file.txt", "This is a test file")

# new overloaded multi-file write method signature:
sandbox.files.write([
  { "path": 'test_file_1.txt', data: "This the first test file." },
  { "path": '/test_dir/test_file_2.txt', data: "This the second test file."},
])
```
see [python-sdk test]( https://github.com/e2b-dev/E2B/pull/451/files#diff-8629a4a0bb30054b0abe02822e9d345d28a9798afae5ba68019e20f0332e5fb7R23) for more examples

mlejva:
I added @jakubno as a reviewer to check the Python SDK as he's the one with the most Python experience on our team

jakubno (commented):
I would argue for splitting the methods, there are two different functions for the user. I think 90% of users will use the simple version for which I like the current signature (no need for custom objects). The only downside is, there will be 2 ways for users to upload data. But if we name this `multifile_upload` or `multifile_write` it shouldn't be confusing

mlejva:
> I would argue for splitting the methods, there are two different functions for the user. I think 90% of users will use the simple version for which I like the current signature (no need for custom objects). The only downside is, there will be 2 ways for users to upload data. But if we name this `multifile_upload` or `multifile_write` it shouldn't be confusing

I think we could even have just the multi-file support in both SDKs because if you want to write a single if, it's just an array with a single item. 
The question in my head is whether it's better DX of having a single method that has a clear arguments but user needs to always pass an array, even for a single file. Or if it's better to have separate functions for single file and multiple files write.

Alternatively, if we want to go Kuba's road with separate functions we could have:
1. `sandbox.files.write()` - for a single file
2. `sandbox.fiels.write_multiple()` for multiple files
Same in JS?

jakubno:
I usually don't like the passing arrays.

When you pass an array the function will return an array and will it have only one element? should I check it has the element or should I risk index out of bounds error?

mlejva:
One more thing I want to discuss - `WriteData` in JS SDK.

The `WriteData` type is this:

```
export type WriteData = string | ArrayBuffer | Blob | ReadableStream
```

We're using it for example here:
```
export type WriteEntry = {
  path: string
  data: WriteData
}
```

`WriteEntry` is a single item in multi file write - `sandbox.files.write([{ path: string, data: WriteData }])`

I'm thinking of maybe we should get rid of this type alias and have it directly as this:
```
export type WriteEntry = {
  path: string
  data: string | ArrayBuffer | Blob | ReadableStream
}
```

The reason would be to make it easier for users to immediately see what `data`  is in their type hints. Otherwise they need to do additional clicks and find what `WriteData` is in our codebase.

mlejva:
> I usually don't like the passing arrays.
> 
> When you pass an array the function will return an array and will it have only one element? should I check it has the element or should I risk index out of bounds error?

That's a good argument. I agree. I don't like dealing with arrays and checking the length as well

ValentaTomas:
I also like the single upload method because it is something users already use a lot; requiring to always pass an array didn't sit well with me.

For splitting the methods — I really don't want to end up with more methods for the same thing on the same level.

> export type WriteEntry = {
>   path: string
>   data: string | ArrayBuffer | Blob | ReadableStream
> }

I think this is the better solution here than the `WriteData` I originally proposed.

ValentaTomas:
> > I usually don't like the passing arrays.
> > When you pass an array the function will return an array and will it have only one element? should I check it has the element or should I risk index out of bounds error?
> 
> That's a good argument. I agree. I don't like dealing with arrays and checking the length as well

For the multifile upload, you have to pass an array though, right?

mlejva:
> For splitting the methods — I really don't want to end up with more methods for the same thing on the same level.

I can maybe see us using the overrides in JS but I really don't like this write signature in Python:
```python
path_or_files: str | List[WriteEntry],
data_or_user: WriteData | Username = "user",
user_or_request_timeout: Optional[float | Username] = None,
```

It's not clear at all how to use these methods just from looking at the code. We should really aim for trying to be as clear as possible just from the code.

It creates many questions in my had, why is data and user mixed together? Why is user and request timeout mixed together? It's very unintuitive

ValentaTomas:
Just to clear up a possible confusion here, because of the start of the discussion:

> ```py
>         path_or_files: str | List[WriteEntry],
>         data_or_user: WriteData | Username = "user",
>         user_or_request_timeout: Optional[float | Username] = None,
> ```
This signature should never be seen by the user, because the method is `@overload`ed, so the visible method signatures will be:


1. This will be default and selected first if you start typing:
```py
    @overload
    def write(
        self,
        path: str,
        data: WriteData,
        user: Username = "user",
    ) -> EntryInfo:
```

2. You can switch to this one explicitly or implicitly by passing an array (because of duck typing) as first argument or `files` argument.
```py
    @overload
    def write(
        self,
        files: List[WriteEntry],
        user: Optional[Username] = "user",
    ) -> List[EntryInfo]:
```

mlejva:
> Just to clear up a possible confusion here, because of the start of the discussion:
> 
> > ```python
> >         path_or_files: str | List[WriteEntry],
> >         data_or_user: WriteData | Username = "user",
> >         user_or_request_timeout: Optional[float | Username] = None,
> > ```
> 
> This signature should never be seen by the user, because the method is `@overload`ed, so the visible method signatures will be:
> 
> 1. This will be default and selected first if you start typing:
> 
> ```python
>     @overload
>     def write(
>         self,
>         path: str,
>         data: WriteData,
>         user: Username = "user",
>         request_timeout: Optional[float] = None,
>     ) -> EntryInfo:
> ```
> 
> 2. You can switch to this one explicitly or implicitly by passing an array (because of duck typing) as first argument or `files` argument.
> 
> ```python
>     @overload
>     def write(
>         self,
>         files: List[WriteEntry],
>         user: Optional[Username] = "user",
>         request_timeout: Optional[float] = None,
>     ) -> List[EntryInfo]:
> ```

I see. I must have missed the `@overload`, sorry. What does the `user` option do? Is that new?

I'll leave @jakubno to add any notes if has anything additional. As I said - he has the most Python experience

jakubno:
> This signature should never be seen by the user, because the method is @overloaded, so the visible method signatures will be:

It ain't true, users can see the source code for python libraries and I often go there

ValentaTomas:
The `user` is an optional argument that can be used to modify who you are making the filesystem operation as — if you need to delete a file owned by root or create a file that is owned by a different user that the default one.

We had a lot of confusion around the users/permissions, so having this explicitly visible and modifiable seems better than hiding it again.

ValentaTomas:
> > This signature should never be seen by the user, because the method is @OverLoaded, so the visible method signatures will be:
> 
> It ain't true, users can see the source code for python libraries and I often go there

Well, I cannot argue with that. At that point isn't the usage clear?

mlejva:
> The `user` is an optional argument that can be used to modify who you are making the filesystem operation as — if you need to delete a file owned by root or create a file that is owned by a different user that the default one.
> 
> We had a lot of confusion around the users/permissions, so having this explicitly visible and modifiable seems better than hiding it again.

Is this only filesystem thing? Or on all methods now? My immediate thought would be to expect something like this on every relevant method.

> We had a lot of confusion around the users/permissions, so having this explicitly visible and modifiable seems better than hiding it again.

I'm pretty sure this won't solve it. People will be still confused and won't know what to pass there. The solution is that users shouldn't need to mess with permissions by default. The filesystem should be regularly accessible. I had very relevant feedback for this from our hackathon:

> Problem uploading data to sandbox and not being able to open the file because of permissions. Had to use `sudo`. Everything was in `/home/user`.

0div:
Would the pythonic `*args` (Non-Keyword Arguments) be a good middle-ground here?
And its equivalent in TS `function sum(...args) {` 

That way use can use one or more args or w/o using arrays.

ValentaTomas:
> Would the pythonic `*args` (Non-Keyword Arguments) be a good middle-ground here? And its equivalent in TS `function sum(...args) {`
> 
> That way a dev can use one or more args or w/o using arrays.

I think it would be slightly problematic because we also want to pass the `opts` for the request.

mlejva:
plus don't we lose all the type hints?

mlejva:
> > Is this only filesystem thing? Or on all methods now? My immediate thought would be to expect something like this on every relevant method.
> > Yes, it is on all relevant methods.
> 
> By default you don't need to mess with this, but when interacting with filesystem, not having this exposed would lead you to not being able to do this. So that's why I choose to do less magic here — the user you pass is used for the operation, there is default of `user` that is being user, but you can override it.

What are the use cases when you need it? It's a simple parameter but it opens a whole can of new questions like "how do I create a new user?" or "how do I switch to a different user when running a command?" that we'll need to provide answers for. That's why I'm not a big fan of this. It feels only half way there because it's missing all the technical content and utilities around it.

ValentaTomas:
> plus don't we lose all the type hints?
The variadic argument could be typed, but generally, it should be the last argument. Otherwise, it won't work.

mishushakov:
Sorry to get in between, but for the Browserbase SDK I was in a similar situation, we ended up with a `load_url` method to load a single web page, `load_urls` to load multiple and a `load` method to handle both (depending on input)

mlejva:
> Sorry to get in between, but for the Browserbase SDK I was in a similar situation, we ended up with a `load_url` method to load a single web page, `load_urls` to load multiple and a `load` method to handle both (depending on input)

Thanks for the input. Would you do it the same way again?

mishushakov:
Would do two methods. My thinking was (still is): I expect a function to return X. If a function should also return Y, I make another function that returns Y. A function that returns both X and Y is flaky

ValentaTomas:
Why you ended up with three in the end?

mishushakov:
Because it felt like if we have the other two why not have it also. But tbh., the whole thing was out of necessity, because I didn't want to start a browser every time I want to fetch a URL, `load_urls` would connect once and do the same thing as `load` for multiple urls. I actually prefer the Node way - you can files.map > fs.writeFile > Promise.all(files)

ValentaTomas:
Three sounds the worst :D

Let's do the two separate if everybody agrees:
- `write_file`/`writeFile`
- `write_files`/`writeFiles`

mlejva:
> Three sounds the worst :D
> 
> Let's do the two separate if everybody agrees:
> 
> * `write_file`/`writeFile`
> * `write_files`/`writeFiles`

I agree about separate methods but not sure about the naming because there's already `files` module:
```
sandbox.files.writeFile() / sandbox.files.write_file()
sandbox.files.write_files() / sandbox.files.write_files()
```
Maybe feels a little weird?

ValentaTomas:
Alternative namespace solutions:
- `sandbox.fs.write_file`
- `sandbox.filesystem.write_file`
- `sandbox.write_file` (getting rid of all namespaces?)

0div:
> Alternative namespace solutions:
> 
> * `sandbox.fs.write_file` (Not sure if it is clear that you should use this for filesystem)
> * `sandbox.filesystem.write_file` (This seems too long)
> * `sandbox.write_file` (Can we get rid of all namespaces without making everything confusing?)
> 
> EDIT: I actually feel that maybe not having namespaces lowers the mental difficulty for me when using the SDK — you just type what you want to do as opposed to having to think in which namespace the method should be.
> 
> EDIT 2: Also I would not even argue for overloaded `write` method without the `files` namespace, because without namespaces you need to name it more specifically like `writeFile` which tells you that it takes just one file.

Keeping a namespace or prefixing the method name could help disambiguate a bit since that kinduv io touches other linux abstractions.

mishushakov:
I'm going with no namespaces. With Stripe SDK it makes sense to have namespaces as they have different "products" and the products have different methods available. We only have one "product" (or class) which is the sandbox, so we should only have sandbox methods on top-level, eg: sandbox.writeFile.

0div:
> I'm going with no namespaces. With Stripe SDK it makes sense to have namespaces as they have different "products" and the products have different methods available. We only have one "product" (or class) which is the sandbox, so we should only have sandbox methods on top-level, eg: sandbox.writeFile.

The question is then is this a product API or a VM API?

ValentaTomas:
I want to ship this by the end of the week too.

For the unclear things we talked about—I would add `sbx.files.writeFiles` for now, alongside the `sbx.files.write` that is already there.

With the next major release (which could include the code interpreter merge) we make the change to namespaces so we will probably end up with `sbx.writeFile`+`sbx.writeFiles`. The alternative is adding it directly on `sbx.writeFiles` but for now I feel that not using the namespaces could make this confusing.

0div:
> I want to ship this by the end of the week too.
> 
> For the unclear things we talked about—I would add `sbx.files.writeFiles` for now, alongside the `sbx.files.write` that is already there.
> 
> With the next major release (which could include the code interpreter merge) we make the change to namespaces so we will probably end up with `sbx.writeFile`+`sbx.writeFiles`. The alternative is adding it directly as `sbx.writeFiles` but for now I feel that not using the namespaces could make this confusing.

Should i merge this https://github.com/e2b-dev/infra/pull/196 with it?

ValentaTomas:
Yes, I left a comment there.
We would merge the envd change, then rebuild base and code interpreter templates, then release the SDK changes.

mlejva:
> Yes, I left a comment there. We would merge the envd change, then rebuild base and code interpreter templates, then release the SDK changes.

I think you might have forgotten the Desktop Sandbox template

apps/web/src/app/(docs)/docs/filesystem/read-write/page.mdx
```diff
@@ -18,20 +18,48 @@ file_content = sandbox.files.read('/path/to/file')
...
+
+You can also write multiple files to the sandbox filesystem using the `files.write()` method.
+
+<CodeGroup>
+```js
+import { Sandbox } from '@e2b/code-interpreter'
+const sandbox = await Sandbox.create()
+
+await sandbox.files.write([
+    { path: "/path/to/a", data: "file content" },
```
mlejva:
```suggestion
    { path: '/path/to/a', data: 'file content' },
```

apps/web/src/app/(docs)/docs/filesystem/read-write/page.mdx
```diff
@@ -18,20 +18,48 @@ file_content = sandbox.files.read('/path/to/file')
...
+    { path: "/another/path/to/b", data: "file content" }
```
mlejva:
```suggestion
    { path: '/another/path/to/b', data: 'file content' }
```

apps/web/src/app/(docs)/docs/filesystem/read-write/page.mdx
```diff
@@ -18,20 +18,48 @@ file_content = sandbox.files.read('/path/to/file')
...
+])
+```
+```python
+from e2b_code_interpreter import Sandbox
+
+sandbox = Sandbox()
+
+await sandbox.files.write([
+    { "path": "/path/to/a", "data": "file content" },
```
mlejva, line 61:
```suggestion
    { path: '/path/to/a", data: 'file content' },
```

apps/web/src/app/(docs)/docs/filesystem/upload/page.mdx
```diff
@@ -25,3 +27,86 @@ with open("path/to/local/file", "rb") as file:
...
+            # Read file contents in binary mode
+            with open(file_path, "rb") as file:
+                files.append({
+                    'path': file_path,
+                    'data': file.read()
+                })
+    
+    return files
+
+files = read_directory_files('/local/dir');
```
mlejva:
```suggestion
files = read_directory_files("/local/dir")
```

apps/web/src/app/(docs)/docs/filesystem/upload/page.mdx
```diff
@@ -25,3 +27,86 @@ with open("path/to/local/file", "rb") as file:
...
+print(files);
```
mlejva:
```suggestion
print(files)
```

apps/web/src/app/(docs)/docs/filesystem/upload/page.mdx
```diff
@@ -25,3 +27,86 @@ with open("path/to/local/file", "rb") as file:
...
+                files.append({
+                    'path': file_path,
+                    'data': file.read()
+                })
+    
+    return files
+
+files = read_directory_files('/local/dir');
+print(files); 
+// [
```
mlejva:
```suggestion
# [
```

apps/web/src/app/(docs)/docs/filesystem/upload/page.mdx
```diff
@@ -25,3 +27,86 @@ with open("path/to/local/file", "rb") as file:
...
+//   { 'path': '/local/dir/file1.txt', 'data': 'File 1 contents...' },
```
mlejva:
```suggestion
#  {"'path": "/local/dir/file1.txt", "data": "File 1 contents..." },
```

apps/web/src/app/(docs)/docs/filesystem/upload/page.mdx
```diff
@@ -25,3 +27,86 @@ with open("path/to/local/file", "rb") as file:
...
+//   { 'path': '/local/dir/file2.txt', 'data': 'File 2 contents...' },
+//   ...
+// ]
```
mlejva:
```suggestion
#   { "path": "/local/dir/file2.txt", "data": "File 2 contents..." },
#   ...
# ]
```

mlejva:
> > Yes, I left a comment there. We would merge the envd change, then rebuild base and code interpreter templates, then release the SDK changes.
> 
> I think you might have forgotten the Desktop Sandbox template

1. Have we rebuilt all the required templates? Base, code interpreter, and desktop
2. Users need to rebuild templates after they update to newer version, right?

mlejva:
Is this waiting for review and ready otherwise?

apps/web/src/app/(docs)/docs/filesystem/read-write/page.mdx
```diff
@@ -18,20 +18,48 @@ file_content = sandbox.files.read('/path/to/file')
 ```
 </CodeGroup>
 
-## Writing files
+## Writing single files
 
-You can write files to the sandbox filesystem using the `files.write()` method.
+You can write signle files to the sandbox filesystem using the `files.write()` method.
```
jakubno:
```suggestion
You can write single files to the sandbox filesystem using the `files.write()` method.
```

ValentaTomas (approved):

ValentaTomas:
I think this could be merged now—the envd should be updated in prod long ago.

PR #457, opened 2024-10-11
Improve file watcher in Python Sync SDK

jakubno:
# Description

Current solution didn't work well with in sync Python. If you're unsure whether a file has been created, you might end up waiting for a timeout to exit the loop. 

New implementation uses polling, you can ask for events whenever you want and it will request envd to send all events (or only new ones). This should make it much better for users with sync Python SDK

Example of a problematic usage before and after:
## Before
```python
sbx = Sandbox()
watcher = sbx.files.watch("/home/user")
sbx.files.make_dir("test")
for event in watcher:
    print(event)
     # !!! if you don't exit, you would be stuck for the rest of the timeout (default 60 seconds)
    break
watcher.close()
```
## After
```python
sbx = Sandbox()
watcher = sbx.files.watch("/home/user")
sbx.files.make_dir("test")
events = watcher.get_new_events()
watcher.stop()
for event in events:
    print(event)
```

Even worse case was if you don't know if anything will happen:
## Before
```python
sbx = Sandbox()
watcher = sbx.files.watch("/home/user")
if random.random() > 0.5:
    sbx.files.make_dir("test")
# There's 50% chance you will get stuck. The only workaround is to run in in separate thread and you kill it after a while (you aren't really sure when it's safe)
for event in watcher:
    print(event)
    #  !!! if you don't exit, you would be stuck for the rest of the timeout (default 60 seconds)
    break
watcher.close()
```
## After
```python
sbx = Sandbox()
watcher = sbx.files.watch("/home/user")
if random.random() > 0.5:
    sbx.files.make_dir("test")
events = watcher.get_new_events()
watcher.stop()
for event in events:
    print(event)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,5 +13,5 @@ generate-js:
 
 generate-python:
 	cd packages/python-sdk && make generate-api
-	buf generate --template spec/envd/buf-python.gen.yaml
-	cd packages/python-sdk && ./scripts/fix-python-pb.sh
+	cd spec/envd && buf generate --template buf-python.gen.yaml
+	cd packages/python-sdk && ./scripts/fix-python-pb.sh && black .
--- a/apps/web/src/code/python/basics/fs_watch.py
+++ b/apps/web/src/code/python/basics/fs_watch.py
@@ -2,19 +2,9 @@ import time
 
 from e2b import Sandbox
 
-watcher = None
+sandbox = Sandbox()
 
-
-def create_watcher(sandbox):  # $HighlightLine
-    # Start filesystem watcher for the /home directory
-    watcher = sandbox.filesystem.watch_dir("/home")  # $HighlightLine
-    watcher.add_event_listener(lambda event: print(event))  # $HighlightLine
-    watcher.start()  # $HighlightLine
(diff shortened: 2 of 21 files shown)


ValentaTomas:
Few notes:
1. Do we agree on the polling design?
2. Protobuf field should snake case
3. I'm thinking about how this interacts with the reconnect — the client watcher keeps around the offset + watcher id, so when you reconnect you effectively cannot reconnect to the watcher. Could be ok, just pointing this out

ValentaTomas:
[E2B-993: Fix unintuitive usage of `files.watch` method in the Python Sync SDK](https://linear.app/e2b/issue/E2B-993/fix-unintuitive-usage-of-fileswatch-method-in-the-python-sync-sdk)

mlejva:
A few notes to both SDKs:

- Instead `sbx.files.watch()` maybe we should do `sbx.files.watch_dir()` to signal better that it's about watching the directory?

- Instead of `watcher.close()`, let's do `watcher.stop()` because when thinking about it - you start watcher and then you stop it.

PR #458, opened 2024-10-16
feat: update grit guide

morgante:
1. Using the remote pattern, users don't need to add anything locally.
2. I usually recommend `--interactive` for new users to review each diff.


mlejva:
This is great thank you!

PR #464, opened 2024-11-01
Make python dependencies more lenient

brendanator:
I cannot run `pipenv run e2b-code-interpreter` as I get a dependency conflict with an existing package which requires `attr >= 24.0`

This change will allow `e2b-code-interpreter` to be used in more places

### DIFF
--- a/.changeset/tall-parrots-wink.md
+++ b/.changeset/tall-parrots-wink.md
@@ -1,0 +1,5 @@
+---
+'@e2b/python-sdk': patch
+---
+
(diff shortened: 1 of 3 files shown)


jakubno:
Hi @brendanator, for pointing this out! There probably isn't a reason to be so strict here. We use only very basic functionality from those packages, so it should be fairly safe.

I fixed the major version for `httpcore` and `httpx`, there may be breaking changes.

Thanks for your contribution!

PR #469, opened 2024-11-11
Refactor CLI to use docker directly instead of starting process

jakubno:
# Description

Using the docker directly should give us more control over Docker, also users won't need docker client installed


ValentaTomas:
We will be reworking the build process. Moved to the internal roadmap.

PR #474, opened 2024-11-22
Add pause and resume to SDKs

ValentaTomas:
Add pause and resume methods to the SDKs that can be used for persistence of sandbox data.


ValentaTomas:
This is waiting until we merge beta SDK into main.

PR #475, opened 2024-11-22
Improve SDK refs

0div:
# Description 
- [x] Version dropdown in TOC 
- [x] Improve js-sdk api ref with typedoc markdown options
- [x] Levarage auto anchor link generation with h2 and h3s for cli sdk
- [x] CSS styling


mishushakov:
I just took a quick look at the Vercel preview of curiosity:
- Don't see a dropdown anywhere?
- For JS, can you remove "Defined in", it isn't particularly useful?
- Can you make methods either bold or ### (h3) in markdown?
- Parameters look different in javascript docs (table) and python (list)

mlejva:
> I just took a quick look at the Vercel preview of curiosity:
> 
> * Don't see a dropdown anywhere?
> * For JS, can you remove "Defined in", it isn't particularly useful?
> * Can you make methods either bold or ### (h3) in markdown?
> * Parameters look different in javascript docs (table) and python (list)

We discussed it in person and made some compromises mostly around unifying styling and information. The problem is we would need to write our own docstrings parser as the current ones are sufficient for us.

> * Don't see a dropdown anywhere?

Version dropdown? We haven't figured out where to put it yet.

> * Can you make methods either bold or ### (h3) in markdown?

This one is the main thing we need to figure out if we can do.

> * Parameters look different in javascript docs (table) and python (list)

Here we compromised and it will be different

mishushakov:
Will you also be adding `Pty` and `Commands` types?

0div:
> Will you also be adding `Pty` and `Commands` types to JS ref?

good catch @mishushakov , I have included them by fixing the typedoc config

.github/scripts/is_new_sdk_ref.sh
jakubno, line 1: not needed

mlejva:
Todo list before we can merge release
- [] move the top bar navigation "Documentation", "SDK Reference", "Dashboard" to the navbar
- [] Fix the navbar item selection not staying selected when you open a page

mlejva (approved):

PR #478, opened 2024-11-23
Add route for creating streams

mlejva:
This PR adds:
- API route that creates a stream for given sandbox ID  `POST /stream/sandbox`
- Dynamic page for displaying the livestream video `/stream/sandbox/[sandboxId]


apps/web/src/app/(dashboard)/dashboard/layout.tsx
```diff
@@ -1,12 +1,14 @@
...
 
-
 export default async function Layout({ children }) {
   return (
-    <LayoutDashboard>
-      {children}
-      <Toaster />
-    </LayoutDashboard>
+    <div className="h-full w-full">
+      <main className="w-full h-full flex flex-col">
```
mishushakov, line 7: Why didn't you update LayoutDashboard instead?

apps/web/src/app/api/stream/sandbox/route.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export async function POST(request: Request) {
+  // TODO: UseG apiKey
+  // const apiKey = request.headers.get('X-API-Key')
+  const { sandboxId } = await request.json()
+  console.log('sandboxId', sandboxId)
+
+  // if (!apiKey) {
+  //   return NextResponse.json({ error: 'Missing X-API-Key' }, { status: 400 })
+  // }
```
mishushakov: There's no security?

apps/web/src/app/layout.tsx
```diff
@@ -39,8 +37,18 @@ declare global {
...
+  const isSandboxRoute = params?.sandboxId !== undefined
+
+  if (isSandboxRoute) {
+    return (
+      <html lang="en">
+        <head></head>
+        <body>{children}</body>
+      </html>
+    )
+  }
```
mishushakov: Is this the correct way to do it?

apps/web/src/app/stream/sandbox/[sandboxId]/page.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  if (error || !data) {
+    return null
+  }
+
+  return { sandboxId, playbackId: data.playback_id }
+}
+
+export default async function StreamPage({ params }: { params: { sandboxId: string } }) {
+  const stream = await fetchStream(params.sandboxId)
```
Comment (mishushakov) [inline apps/web/src/app/stream/sandbox/[sandboxId]/page.tsx]: No security?

apps/web/src/components/ui/Popover.tsx
```diff
@@ -16,7 +16,7 @@ const PopoverContent = React.forwardRef<
   <PopoverPrimitive.Portal>
     <PopoverPrimitive.Content
       ref={ref}
-      align={align}
+      align={align as any}
```
mishushakov: Avoid "any"

apps/web/src/middleware.ts
```diff
@@ -5,6 +5,7 @@ export async function middleware(req: NextRequest): Promise<NextResponse> {
   if (req.method !== 'GET') return NextResponse.next()
 
   const url = new URL(req.nextUrl.toString())
+  console.log('url', url)
```
mishushakov: Remove this

apps/web/src/types/mux-player-react.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module '@mux/mux-player-react' {
+  export interface MuxPlayerProps {
+    [key: string]: any;
+    playbackId: string;
+    metadata?: Record<string, any>;
+  }
+
+  export default function MuxPlayer(props: MuxPlayerProps): JSX.Element
+}
```
mishushakov: Why is this required? They don't have typings?

mishushakov (commented):
Added feedback

apps/web/src/app/api/stream/sandbox/route.ts
mishushakov: Also, why are we using api routes, our backend is in Go

apps/web/src/app/(dashboard)/dashboard/layout.tsx
mlejva: I deleted the component and moved the layout directly to this `layout.tsx` file. The `LayoutDashboard` wasn't used anywhere but here and there wasn't much logic in it.

apps/web/src/app/layout.tsx
mlejva: Shouldn't be there. I forgot to delete. Thanks for catching.

apps/web/src/types/mux-player-react.d.ts
mlejva: Correct. That was my experience. I installed their packages and the linter couldn't find declarations. They also don't seem to have `@types/...` package.

apps/web/src/app/api/stream/sandbox/route.ts
mlejva: Discussed this @ValentaTomas. We felt like this is a better place for the implementation given it's right next to the actual page rendering the stream.

apps/web/src/app/stream/sandbox/[sandboxId]/page.tsx
Comment (mlejva) [reply apps/web/src/app/stream/sandbox/[sandboxId]/page.tsx]: Discussed this with @ValentaTomas. Streams are public so reading them should be public. Maybe we could allow creating them only if you're sending an E2B API key @ValentaTomas ?

apps/web/src/types/mux-player-react.d.ts
mlejva: However, [their project](https://github.com/muxinc/elements/tree/main/packages/mux-player-react) is in TypeScript. It's weird that the linter is failing on it because it doesn't fail on any other package.

mlejva: Resolved. Seemed to be a cache issue.

mlejva:
@ValentaTomas I also realized - we should add a migration for the new Supabase table. Right?

apps/web/src/app/stream/sandbox/[sandboxId]/page.tsx
Comment (ValentaTomas) [reply apps/web/src/app/stream/sandbox/[sandboxId]/page.tsx]: Creating them is behind API key+sandbox ID.

For access, we want there to be just a link you can follow. The most straightforward solution here then could be that we also pass some per sandbox stream secret when we create the sandbox that is then added to the link returned in SDK in query params.

That could offer more flexibility in the future, because right now the sandbox ID also serve as the secret you use for connecting.

Comment (ValentaTomas) [reply apps/web/src/app/stream/sandbox/[sandboxId]/page.tsx]: To future-proof this, it might be worth adding and passing such a field so we don't face compatibility problems later.

mlejva:
I added additional security layer with `token` URL parameter
Two API routes:

## 1. `POST /api/stream/sandbox`
Creates a new stream for a sandbox if a stream doesn't exist yet (each sandbox can have exactly one stream at any given time now). Returns the token and stream key
- When creating a new stream, the token gets generated in the DB via a postgres function

## 2. `GET /api/stream/sandbox/[sandboxId]`
Checks the DB for existing stream for the given sandbox and returns the token for this stream

apps/web/src/app/stream/sandbox/[sandboxId]/page.tsx
Comment (mlejva) [reply apps/web/src/app/stream/sandbox/[sandboxId]/page.tsx]: Done

ValentaTomas (approved):

PR #481, opened 2024-11-26
Add command for building dependencies for generating python envd client

ValentaTomas:
Add command for building dependencies for generating python envd client.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,8 @@ generate-js:
 	cd packages/js-sdk && pnpm generate-envd-api
 	cd spec/envd && buf generate --template buf-js.gen.yaml
 
-generate-python:
+.PHONY: connect-python
+generate-python: connect-python
 	cd packages/python-sdk && make generate-api
 	cd spec/envd && buf generate --template buf-python.gen.yaml
 	cd packages/python-sdk && ./scripts/fix-python-pb.sh && black .
@@ -19,3 +20,6 @@ generate-python:
 .PHONY: init-styles
 init-styles:
 	vale sync
+
+connect-python:
+	$(MAKE) -C packages/connect-python bin/protoc-gen-connect-python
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -11,11 +11,15 @@ generate-js:
 	cd packages/js-sdk && pnpm generate-envd-api
 	cd spec/envd && buf generate --template buf-js.gen.yaml
 
-generate-python:
+.PHONY: connect-python
```
jakubno:
```suggestion
.PHONY: generate-python
```

Makefile
```diff
@@ -11,11 +11,15 @@ generate-js:
...
 	cd packages/python-sdk && ./scripts/fix-python-pb.sh && black .
 
 .PHONY: init-styles
 init-styles:
-	vale sync
\ No newline at end of file
+	vale sync
+
+connect-python:
+	$(MAKE) -C packages/connect-python bin/protoc-gen-connect-python
```
jakubno: Does it make sense to call this ever separately? I find it little bit confusing if I only see names of the methods
```suggestion
.PHONY: connect-python
connect-python:
	$(MAKE) -C packages/connect-python bin/protoc-gen-connect-python
```

ValentaTomas: Not separately, but I wanted to have the command somewhere and `make connect-python` which is the name of the package made some semantic sense, because you are building an artifact.

Probably should move the command to the connect-python package and then call it from the generate-python.

ValentaTomas:
@jakubno Should be fixed now.

packages/connect-python/Makefile
```diff
@@ -26,3 +26,6 @@ bin/$(plugin): $(wildcard cmd/$(plugin)/*.go) pyproject.toml Makefile
 	  ./cmd/$(plugin)
 
 .PHONY: dev fmt lint upload clean
```
jakubno:
```suggestion
.PHONY: dev fmt lint upload clean build
```

jakubno (approved):

PR #483, opened 2024-11-29
Template management in dashboard

mishushakov:
- Adds more info to templates section
- Publish/unpublish templates
- Delete templates
- Requires latest infra backend to test


apps/web/src/components/Dashboard/Templates.tsx
```diff
@@ -28,43 +59,128 @@ export function TemplatesContent({
...
 
   useEffect(() => {
-    function f() {
-      const accessToken = user.accessToken
-      if (accessToken) {
-        fetchTemplates(accessToken, teamId).then((newTemplates) => {
-          if (newTemplates) {
-            setTemplates(newTemplates)
-          }
+    async function fetchTemplates() {
```
0div: have u considered trying the [useQuery](https://tanstack.com/query/latest/docs/framework/react/overview) hook?

apps/web/src/components/Dashboard/Templates.tsx
```diff
@@ -28,43 +59,128 @@ export function TemplatesContent({
...
   return (
     <div className="flex flex-col justify-center">
       <Table>
         <TableHeader>
-          <TableRow className="hover:bg-orange-500/10 dark:hover:bg-orange-500/10 border-b border-white/5 ">
+          <TableRow className="hover:bg-orange-500/10 dark:hover:bg-orange-500/10 border-b border-white/5">
+            <TableHead>Visibility</TableHead>
             <TableHead>Template ID</TableHead>
-            <TableHead>Template Name</TableHead>
+            <TableHead>Aliases</TableHead>
```
0div: is `Alias` consistent with the rest of SDK's naming convention?

apps/web/src/components/Dashboard/Templates.tsx
```diff
@@ -74,34 +190,145 @@ export function TemplatesContent({
...
               >
+                <TableCell>
+                  <TooltipProvider>
+                    <Tooltip>
+                      <TooltipTrigger asChild>
+                        <Button
+                          variant="ghost"
+                          size="icon"
+                          onClick={() => {
+                            setCurrentTemplate(template)
```
0div: preventing event propagation is probably a good idea inside a dynamic table like this

apps/web/src/components/ui/tooltip.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const TooltipContent = React.forwardRef<
+  React.ElementRef<typeof TooltipPrimitive.Content>,
+  React.ComponentPropsWithoutRef<typeof TooltipPrimitive.Content>
+>(({ className, sideOffset = 4, ...props }, ref) => (
+  <TooltipPrimitive.Content
+    ref={ref}
+    sideOffset={sideOffset}
+    className={cn(
+      'z-50 overflow-hidden rounded-md border bg-popover px-3 py-1.5 text-sm text-popover-foreground shadow-md animate-in fade-in-0 zoom-in-95 data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2',
```
0div, line 22: no action required, just comment: tailwind get ugly, thank goodness for vscode tw inline fold extension

0div (approved):

apps/web/src/components/Dashboard/Templates.tsx
mishushakov: I know about it, but it will be in the new web app

apps/web/src/components/Dashboard/Templates.tsx
mishushakov: "Template Name" is the one we're using currently in the CLI, but in SDKs it's alias

apps/web/src/components/ui/tooltip.tsx
mishushakov: this is not our code, [shadcn](https://ui.shadcn.com) it is

apps/web/src/components/Dashboard/Templates.tsx
mishushakov: it should be alright, no?

0div: i think it's good practice for nested elements even if it's currently alright, might come back to bite you when updating event handlers of parent elements.

apps/web/src/components/Dashboard/Templates.tsx
jakubno: I see `Template Name` in CLI and in SDK it isn't really exposed

apps/web/src/components/Dashboard/Templates.tsx
jakubno: I see `Template Name` in CLI and in SDK it isn't really exposed. Also users can set only 1 alias, so plural doesn't make much sense (the code is little bit confusing for that)

apps/web/src/components/Dashboard/Templates.tsx
jakubno: It isn't in form, so it shouldnt be problem right?

0div: you're thinking of `preventDefault()`, the `preventPropagation()` is to avoid triggering other parent `onClick`s

jakubno (changes requested):
.

apps/web/src/components/Dashboard/Templates.tsx
mishushakov: have added it, but keep in mind there are no event listeners in parent

apps/web/src/components/Dashboard/Templates.tsx
mishushakov: Ok, reverted

apps/web/src/components/Dashboard/Templates.tsx
jakubno: I am not sure if I am not missing something, but as Mish says, there are no parent elements with a click handler and I wouldn't expect there to be any. Is there a reason why to put it here? Wouldn't you put it everywhere then?

apps/web/src/components/Dashboard/Templates.tsx
```diff
@@ -28,43 +59,128 @@ export function TemplatesContent({
...
+    setIsPublishTemplateDialogOpen(false)
+  }
+
   return (
     <div className="flex flex-col justify-center">
       <Table>
         <TableHeader>
-          <TableRow className="hover:bg-orange-500/10 dark:hover:bg-orange-500/10 border-b border-white/5 ">
+          <TableRow className="hover:bg-orange-500/10 dark:hover:bg-orange-500/10 border-b border-white/5">
+            <TableHead>Visibility</TableHead>
```
jakubno, line 148: https://e2b-team.slack.com/archives/C07UR0U4QD7/p1732745931793249?thread_ts=1732743673.013419&cid=C07UR0U4QD7

jakubno: I would probably rename to `Private` and do ✔️  or :x:

apps/web/src/components/Dashboard/Templates.tsx
0div: In this case all the `<Table...` elements and im not saying there are currently click handlers, i was saying it could avoid unexpected behavior were there to be event handlers added in the future.

apps/web/src/components/Dashboard/Templates.tsx
mishushakov: I have changed the style a little bit, let me know if it's better

apps/web/src/components/Dashboard/Templates.tsx
mishushakov: don't worry about it, the button is removed from there now and the table we'll be replacing anyways with a tan-stack one

0div: Sounds good 👍  
Example of what i meant previously: https://github.com/TanStack/table/discussions/2243

jakubno (approved):
Looks good, i will merge once the update API server is live

mishushakov:
Good to go.

PR #484, opened 2024-11-29
Added publish/unpublish commands to the CLI

mishushakov:
Adds publish/unpublish commands to the CLI to allow changing the visibility of templates. Publishing a template makes it visible to other teams using E2B, while setting it private (default) only makes it visible within your own team.

example usage:

``` 
e2b template publish <template>
```

```
e2b template unpublish <template>
```


mlejva (approved):
This looks good to me usage-wise. Let's wait for @jakubno or @ValentaTomas

ValentaTomas:
As @jakubno mentioned, this is waiting for the deployment of the new API, so the used endpoints are available.

jakubno (approved):
I will merge it after we deploy API today

PR #488, opened 2024-12-02
Add recursive directory watch

dobrac:
Adds recursive directory watch to the SDK and documentation section. Requires merging and deploying https://github.com/e2b-dev/infra/pull/210 first.

### Proposed API change:
**JS**
```js
await sandbox.files.watchDir(dirname, handler, {
  recursive: true
})
```

**Python**
```py
sandbox.files.watch_dir(dirname, recursive=True)
```

---

### Notes

Python SDK was generated using buf.build/protocolbuffers/python:v28.3. The runtime check was removed for now. Further tests might be required to make sure the runtime is at least the generated version before including the runtime check.
Changing the following in the file spec/envd/buf-python.gen.yaml (generator) allows using different protoc-gen-python version:
From: `plugin: python`
To: `plugin: buf.build/protocolbuffers/python:v28.3`


apps/web/src/app/(docs)/docs/filesystem/watch/page.mdx
```diff
@@ -39,3 +39,49 @@ for event in events: # $HighlightLine
...
+  if (event.type === FilesystemEventType.WRITE) {
+    console.log(`wrote to file ${event.name}`)
+  }
+}, {
+  recursive: true // $HighlightLine
+})
+
+// Trigger file write event
+await sandbox.files.write(`${dirname}/my-folder/my-file`, 'hello') // $HighlightLine
+```
```
0div, line 71: I think you need to have a newline, else the markdown formatting will break for these two distinct code snippets (use the github "view file" option to see for yourself)

packages/js-sdk/tests/sandbox/files/watch.test.ts
```diff
@@ -31,6 +31,71 @@ sandboxTest('watch directory changes', async ({ sandbox }) => {
...
+  const handle = await sandbox.files.watchDir(dirname, async (event) => {
+    if (event.type === FilesystemEventType.WRITE && event.name === expectedFileName) {
+      trigger()
+    }
+  }, {
+    recursive: true
+  })
+
+  await sandbox.files.makeDir(`${dirname}/${nestedDirname}`)
+  await sandbox.files.write(`${dirname}/${nestedDirname}/${filename}`, content)
```
0div: could you test if the directory being created is watched as well, after all a directory is a type of file in linux.

packages/js-sdk/tests/sandbox/files/watch.test.ts
```diff
@@ -31,6 +31,71 @@ sandboxTest('watch directory changes', async ({ sandbox }) => {
...
+
+  await eventPromise
+
+  await handle.stop()
+})
```
0div, line 109: if you want to cleanup you can use `vitest`'s `onTestFinished` helper.

0div (changes requested):
Overall this looks good from a code perspective 👍  

From a product perspective, do you think it would be better _not_ to make the recursive an option and just always watch recursively ? Also taking into account that the instructions said: 
> Make as few changes to the communication between envd and SDKs as possible

packages/js-sdk/tests/sandbox/files/watch.test.ts
dobrac: sure 👍

packages/js-sdk/tests/sandbox/files/watch.test.ts
dobrac: are you referencing the `handle.stop()` or the folder removal?

apps/web/src/app/(docs)/docs/filesystem/watch/page.mdx
dobrac: I did test it on the website locally, and it doesn't seem to make any difference, but I will add the newline for consistency

packages/js-sdk/tests/sandbox/files/watch.test.ts
0div: mostly had the folder removal in mind so that you can re-run the tests w/o a problem.

dobrac: updated

jakubno (changes requested):
The envd version check is missing in Python SDK

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -115,10 +116,13 @@ export class Sandbox extends SandboxApi {
...
-    this.envdApi = new EnvdApiClient({
-      apiUrl: this.envdApiUrl,
-      logger: opts?.logger,
-    })
+    this.envdApi = new EnvdApiClient(
+      {
+        apiUrl: this.envdApiUrl,
+        logger: opts?.logger,
+      },
+      opts?.envdVersion
```
jakubno: why not like this?
```suggestion
        apiUrl: this.envdApiUrl,
        logger: opts?.logger,
        envdVersion: opts?.envdVersion
      },
```

packages/js-sdk/tests/sandbox/files/watch.test.ts
```diff
@@ -31,6 +31,79 @@ sandboxTest('watch directory changes', async ({ sandbox }) => {
...
 
+sandboxTest('watch recursive directory changes', async ({ sandbox }) => {
+  const dirname = 'test_recursive_watch_dir'
+  const nestedDirname = 'test_nested_watch_dir'
+  const filename = 'test_watch.txt'
+  const content = 'This file will be watched.'
+  const newContent = 'This file has been modified.'
+
+  await sandbox.files.makeDir(`${dirname}/${nestedDirname}`)
+  onTestFinished(() => sandbox.files.remove(dirname))
```
jakubno: Technically correct, but not really necessary, the sandbox is killed after the test ends. This will only add one call to sandbox, but you could do this for local testing. Locally the same sandbox is reused
```suggestion
  if (isDebug) {
    onTestFinished(() => sandbox.files.remove(dirname))
  }
```

packages/js-sdk/tests/sandbox/files/watch.test.ts
```diff
@@ -31,6 +31,79 @@ sandboxTest('watch directory changes', async ({ sandbox }) => {
...
+})
+
+sandboxTest('watch recursive directory after nested folder addition', async ({ sandbox }) => {
+  const dirname = 'test_recursive_watch_dir_add'
+  const nestedDirname = 'test_nested_watch_dir'
+  const filename = 'test_watch.txt'
+  const content = 'This file will be watched.'
+
+  await sandbox.files.makeDir(dirname)
+  onTestFinished(() => sandbox.files.remove(dirname))
```
jakubno: same here

packages/js-sdk/src/sandbox/index.ts
dobrac: I was thinking it may be different scope than "ConnectionConfig" (maybe in the future part of some Metadata interface?), but If you want, I can move it there

packages/js-sdk/tests/sandbox/files/watch.test.ts
dobrac: 👍

packages/js-sdk/tests/sandbox/files/watch.test.ts
dobrac: 👍

packages/js-sdk/src/sandbox/index.ts
dobrac: Resolved by wrapping envdVersion to the metadata object

ValentaTomas:
I think the envd part is already merged, right?

0div:
> I think the envd part is already merged, right?

looks like it https://github.com/e2b-dev/infra/pull/233

jakubno (approved):

PR #493, opened 2024-12-06
Redirect the root URL route to new Webflow hosting

mlejva:
This PR changes the target hostname for the root route (`/`) for https://e2b.dev to reflect our new landing page hosting.

This is the first step towards removing the reverse proxy handling from our public web app.


mishushakov (commented):
looks good, but can we remove "made in Webflow"

mlejva:
> looks good, but can we remove "made in Webflow"

It shouldn't be there. Try reloading without cache pls

mishushakov:
another thing is: default is light mode, but my os in dark mode

mishushakov:
opened in incognito
Screenshot 2024-12-06 at 01 52 30

jakubno (approved):
LGTM

mlejva:
> opened in incognito Screenshot 2024-12-06 at 01 52 30

This is really weird because if you go to https://e2b-dev.webflow.io/ which this deployed route points to, it's not there

PR #497, opened 2024-12-06
Added developer panel

mishushakov:
- Click on team dropdown > Developer
- Setting alternative API URL
- Setting alternative Billing URL
- Resetting the above to defaults (provided by env variable)

### DIFF
--- a/apps/web/src/app/(dashboard)/dashboard/layout.tsx
+++ b/apps/web/src/app/(dashboard)/dashboard/layout.tsx
@@ -3,8 +3,8 @@ import { Toaster } from '@/components/ui/toaster'
 
 export default async function Layout({ children }) {
   return (
-    <div className="h-full w-full">
-      <main className="w-full h-full flex flex-col">
+    <div className="h-full w-full flex flex-col">
+      <main className="w-full flex flex-col flex-1">
         {children}
         <Toaster />
       </main>
--- a/apps/web/src/app/(dashboard)/dashboard/page.tsx
+++ b/apps/web/src/app/(dashboard)/dashboard/page.tsx
@@ -230,18 +230,16 @@ const MenuItem = ({
   onClick: () => void
 }) => (
   <div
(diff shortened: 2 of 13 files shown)


mlejva:
Looks like the auth for preview URLs stopped working because it redirects to the production URL


https://github.com/user-attachments/assets/c14fae5b-9e80-49b4-ae38-bd9cf93150cd

mishushakov:
try with username/password for now, not related to this

mlejva:
Thanks, works now.

1. Please fix the unnecessary scrolling and broken footer
2. What's the reason to have the "Developer" button in the team dropdown and not as one of the tabs on the left? I want to better understand the thinking here
3. Please change the developer icon to something else because this icon is associated with bugs

https://github.com/user-attachments/assets/287bc7cd-edf0-4698-8017-6aabf0733e7d

mishushakov:
1. on it
2. it's a hidden setting
3. on it

mishushakov:
removed scrolling and made footer not-so-ugly

mlejva:
> 2. it's a hidden setting

I get the hidden element of it it just doesn't fit well to me that it's in the team selection. Do we need to "hide" it?

mlejva:
Great, thanks! Let's go a little bit further in the UX and:

1. Add description of what each field is for like this:
Screenshot 2024-12-06 at 1 46 10 PM

2. My understanding is that if a user changes the `API URL`, it implies they're running on a separate cluster and they should set the `E2B_DOMAIN` env var when using our SDK. I want us to show a code snippet showing that the user should set the env var. For example if `API URL` is `api.domain.com` people should set `E2B_DOMAIN=domain.com` before using our SDK. Does that make sense?

3. I find it a bit confusing that no tab on the left is selected because the developer settings is in the dropdown menu. It's just a bit inconsistent. I think we don't need to hide it.
 
Screenshot 2024-12-06 at 1 51 38 PM

mishushakov:
updated
Screenshot 2024-12-06 at 23 31 45

mlejva:
I pushed some minor updates. I spoke with @jakubno that we can hide the billing URL for now and it should work for users as well.

mlejva (approved):

PR #503, opened 2024-12-11
Add multi-file upload capability to SDKs

0div:
# Description 
Update the js-sdk write method to work as follows and adapt tests
```js
import { Sandbox } from 'e2b'

const sbx = await Sandbox.create()

const infos = sbx.files.write([
    { path: "/path/to/a", data: "<blob>" },
    { path: "/path/to/b", data: "<blob>" },
], filesystemRequestOpts)

console.log('Files created', infos)
// [
//     { "name": "a", "path": "/path/to/a" },
//     { "name": "b", "path": "/path/to/b" }
//  ]
```

Upadte the python-sdk write method as well
```py
from e2b import Sandbox

sbx = Sandbox()

files = []
for i in range(10):
    path = f"test_write_{i}.txt"
    content = f"This is a test file {i}."
    files.append({"path": path, "data": content})

infos = sbx.files.write(files)
print('Files created', infos)
# [
#   { "name": "a", "path": "/path/to/a" },
#   { "name": "b", "path": "/path/to/b" }
#  ]
```

# Test
```sh
cd packages/js-sdk
pnpm test

cd packages/python-sdk
pnpm test
```

### DIFF
--- a/packages/js-sdk/src/sandbox/filesystem/index.ts
+++ b/packages/js-sdk/src/sandbox/filesystem/index.ts
@@ -246,7 +246,7 @@ export class Filesystem {
           // Important: RFC 7578, Section 4.2 requires that if a filename is provided,
           // the directory path information must not be used.
           // BUT in our case we need to use the directory path information with a custom
-          // muktipart part name getter in envd.
+          // multipart part name getter in envd.
           fd.append('file', blob, files[i].path)
 
           return fd
--- a/packages/js-sdk/src/sandbox/filesystem/index.ts
+++ b/packages/js-sdk/src/sandbox/filesystem/index.ts
@@ -22,6 +22,13 @@ import { FileType as FsFileType, Filesystem as FilesystemService } from '../../e
 
 import { WatchHandle, FilesystemEvent } from './watchHandle'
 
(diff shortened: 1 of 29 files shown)


packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -214,32 +221,36 @@ export class Filesystem {
...
-        const fd = new FormData()
-
-        fd.append('file', blob)
-
-        return fd
+        return blobs.reduce((fd, blob, i) => {
+          // Important: RFC 7578, Section 4.2 requires that if a filename is provided,
+          // the directory path information must not be used.
+          // BUT in our case we need to use the directory path information with a custom
+          // muktipart part name getter in envd.
```
mlejva:
```suggestion
          // multipart part name getter in envd.
```

mlejva:
@0div please also add/update docs to reflect the breaking change

0div:
> @0div please also add/update docs to reflect the breaking change once the SDK ergonomics are set

@jamesmurdza tested it with `js-sdk`, it doesn't throw, it uses replacement chars by default.

0div:
per discussion, moving towards overloading approach 
https://github.com/e2b-dev/E2B/pull/451

PR #505, opened 2024-12-12
Make stdout and stderr handle decode errors gracefully 

0div:
# Description
The `sbx.commands.run` method can generate a `stdout` with badly encoded `UTF-8` bytes, which can cause this kind of errors: 
```py
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe2 in position 8191: unexpected end of data
```

see more here: https://linear.app/e2b/issue/E2B-1291/unicodedecodeerror-in-python-sdk-when-using-curl-command

To address this we update the `decode` error handler to replace with the unicode replacement char. 

# Test
```sh
 poetry run pytest -s -n 4 -k "tests/sync/sandbox_sync/commands/test_run"
 ```

### DIFF
--- a/packages/python-sdk/e2b/sandbox_async/commands/command_handle.py
+++ b/packages/python-sdk/e2b/sandbox_async/commands/command_handle.py
@@ -112,11 +112,11 @@ class AsyncCommandHandle:
         async for event in self._events:
             if event.event.HasField("data"):
                 if event.event.data.stdout:
-                    out = event.event.data.stdout.decode()
+                    out = event.event.data.stdout.decode('utf-8', 'replace')
                     self._stdout += out
                     yield out, None, None
                 if event.event.data.stderr:
-                    out = event.event.data.stderr.decode()
+                    out = event.event.data.stderr.decode('utf-8', 'replace')
                     self._stderr += out
                     yield None, out, None
                 if event.event.data.pty:
--- a/packages/python-sdk/e2b/sandbox_sync/commands/command_handle.py
+++ b/packages/python-sdk/e2b/sandbox_sync/commands/command_handle.py
@@ -65,11 +65,11 @@ class CommandHandle:
         for event in self._events:
             if event.event.HasField("data"):
                 if event.event.data.stdout:
-                    out = event.event.data.stdout.decode()
+                    out = event.event.data.stdout.decode('utf-8', 'replace')
                     self._stdout += out
                     yield out, None, None
                 if event.event.data.stderr:
-                    out = event.event.data.stderr.decode()
+                    out = event.event.data.stderr.decode('utf-8', 'replace')
                     self._stderr += out
                     yield None, out, None
                 if event.event.data.pty:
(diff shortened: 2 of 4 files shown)


ValentaTomas:
1. Can this happen in the JS too?
2. Aren't we missing the change is async?

0div:
> 1. Can this happen in the JS too?
> 2. Aren't we missing the change is async?

1. @jamesmurdza tested it with js-sdk, it doesn't throw, it uses replacement chars by default.
2. Adding it to async 👍

ValentaTomas:
Do we know why it is happening? Can we improve the encoding in envd or are we doing something improper while encoding?

ValentaTomas:
My question is basically—why are there undecodable chars in the stdout/stderr? Is this because of the command output or are we messing something up in the envd?

0div:
> Do we know why it is happening? Can we improve the encoding in envd or are we doing something improper while encoding?

Essentially when reading data from a network socket, the client might encounter a scenario where a multi-byte UTF-8 character is split across network reads. UTF-8 uses multiple bytes to represent characters with codepoints above 127, and these characters can be interrupted mid-sequence when the buffer is filled.

0div:
> > Do we know why it is happening? Can we improve the encoding in envd or are we doing something improper while encoding?
> 
> Essentially when reading data from a network socket, the client might encounter a scenario where a multi-byte UTF-8 character is split across network reads. UTF-8 uses multiple bytes to represent characters with codepoints above 127, and these characters can be interrupted mid-sequence when the buffer is filled.

That's why this can happen random

ValentaTomas:
> > Do we know why it is happening? Can we improve the encoding in envd or are we doing something improper while encoding?
> 
> Essentially when reading data from a network socket, the client might encounter a scenario where a multi-byte UTF-8 character is split across network reads. UTF-8 uses multiple bytes to represent characters with codepoints above 127, and these characters can be interrupted mid-sequence when the buffer is filled.

Should this happen here when the messages we are getting from envd are actually structured and contain multiple fields in addition to the stdout/stderr?

ValentaTomas:
With that in mind, should we fix this when reading the messages here (https://github.com/e2b-dev/E2B/blob/5f022cc6deac06bc637ea195edf64b708ca1eb4e/packages/python-sdk/e2b_connect/client.py#L395) because we might not be reading the whole message?

0div:
> With that in mind, should we fix this when reading the messages here (
> 
> https://github.com/e2b-dev/E2B/blob/5f022cc6deac06bc637ea195edf64b708ca1eb4e/packages/python-sdk/e2b_connect/client.py#L395
> 
> ) because we might not be reading the whole message?

IIUC you don't get to control where the split happens when reading byte stream from the network, but at the same time im surprised this is not a solved problem given the ubiquity of utf8 encoded byte array streaming

mlejva:
> With that in mind, should we fix this when reading the messages here (
> 
> https://github.com/e2b-dev/E2B/blob/5f022cc6deac06bc637ea195edf64b708ca1eb4e/packages/python-sdk/e2b_connect/client.py#L395
> 
> ) because we might not be reading the whole message?

That sounds like a separate PR. This PR looks good to me!

mlejva (approved):

PR #514, opened 2024-12-19
Use billing subdomain to simplify dashboard for custom domains 

jakubno:
# Description

Billing service runs on `billing.` subdomain, so you can pass only 1 url now


mishushakov (commented):
I think domain sounds confusing, maybe use apiHost? or something that should tell it is hostname/url/address for the API

jakubno:
It isn't api url, it's the base url:
domain: `e2b.dev`
api url: `https://api.e2b.dev`

mlejva (approved):

PR #515, opened 2024-12-19
Handle 404 for not running/existent sandbox

dobrac:


jakubno:
We're handling it little bit differently, we are returning 502 but with custom message to distinguish between 404 for the resource and an unexpected not found for sandbox

PR #517, opened 2024-12-20
Add `getInfo/get_info` methods to the SDKs

mishushakov:
- Adds the ability to retrieve the current timeout of the sandbox via SDK (pre-requisite for events)

### DIFF
--- a/packages/js-sdk/src/sandbox/index.ts
+++ b/packages/js-sdk/src/sandbox/index.ts
@@ -318,11 +318,6 @@ export class Sandbox extends SandboxApi {
   }
 
   async getTimeout(opts?: Pick<SandboxOpts, 'requestTimeoutMs'>) {
-    if (this.connectionConfig.debug) {
-      // Skip killing in debug mode
-      return
-    }
-
     const { endAt } = await Sandbox.get(this.sandboxId, {
       ...this.connectionConfig,
       ...opts,
--- a/packages/js-sdk/src/sandbox/index.ts
+++ b/packages/js-sdk/src/sandbox/index.ts
@@ -317,6 +317,13 @@ export class Sandbox extends SandboxApi {
     await Sandbox.kill(this.sandboxId, { ...this.connectionConfig, ...opts })
   }
 
+  /**
+   * Get the timeout of the sandbox.
+   *
+   * @param opts connection options.
+   *
+   * @returns timeout of the sandbox.
+   */
   async getTimeout(opts?: Pick<SandboxOpts, 'requestTimeoutMs'>) {
     const { endAt } = await Sandbox.get(this.sandboxId, {
       ...this.connectionConfig,
--- a/packages/js-sdk/src/sandbox/index.ts
+++ b/packages/js-sdk/src/sandbox/index.ts
@@ -317,6 +317,20 @@ export class Sandbox extends SandboxApi {
     await Sandbox.kill(this.sandboxId, { ...this.connectionConfig, ...opts })
   }
 
+  async getTimeout(opts?: Pick<SandboxOpts, 'requestTimeoutMs'>) {
+    if (this.connectionConfig.debug) {
+      // Skip killing in debug mode
+      return
+    }
+
+    const { endAt } = await Sandbox.get(this.sandboxId, {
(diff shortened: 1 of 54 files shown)


packages/js-sdk/src/sandbox/index.ts
```diff
@@ -317,6 +317,22 @@ export class Sandbox extends SandboxApi {
     await Sandbox.kill(this.sandboxId, { ...this.connectionConfig, ...opts })
   }
 
+  /**
+   * Get the timeout of the sandbox.
+   *
+   * @param opts connection options.
+   *
+   * @returns timeout of the sandbox.
```
0div: This returns the "end time", no?? 
i understand you want the naming to be consistent with `setTimeout` but to me "timeout" means the "delay".

packages/js-sdk/src/sandbox/sandboxApi.ts
```diff
@@ -80,6 +85,52 @@ export class SandboxApi {
     return true
   }
 
+  /**
+   * Get running sandbox.
+   *
+   * @param sandboxId sandbox ID.
+   * @param opts connection options.
+   *
+   * @returns running sandbox.
```
0div, line 94: doesn't this just return `SanboxInfo` ? My understanding is that `connect()` returns an actual instance of `Sandbox`, in which case, shouldn't the method be called `getInfo` or something to avoid confusion?

packages/js-sdk/tests/sandbox/timeout.test.ts
```diff
@@ -10,14 +10,22 @@ sandboxTest.skipIf(isDebug)('shorten timeout', async ({ sandbox }) => {
...
 
-  await wait(6000)
+    await wait(6000)
+
+    await sandbox.isRunning()
+  }
+)
 
-  await sandbox.isRunning()
+sandboxTest.skipIf(isDebug)('get sandbox timeout', async ({ sandbox }) => {
```
0div, line 28: some assertions around the time left after a waiting for a period of time with reasonable margins could be insightful here?

0div (changes requested):
Same comments about js-sdk apply to python-sdk.

packages/js-sdk/tests/sandbox/timeout.test.ts
mishushakov: my test pass

0div: i meant assert that some of the expected time arithmetic is ball park what we would expect, ie `setTimeout(1000ms) <=> endAt == now() + ~1000ms`

0div (approved):

mlejva:
@mishushakov Please add sections/pages to docs so people know this method exists and how it can be used

ValentaTomas:
I was checking RFC about adding sandbox state getter (https://github.com/e2b-dev/RFCs/pull/6/files) and I think these are related—maybe the state should be part of the info?

It is possible that we can add the state later, but I would like to have a high-level agreement on where these things should be.

mishushakov:
State can be a part of the info, I agree, but maybe we don't need getInfo as it might be too opaque (what info?) and instead getState() and getEndDate() / getTimeout()

ValentaTomas:
Yes, I think having the state, endDate, etc., separated is better here.
We can keep the discussion about state in the RFC and focus on how the endData/timeout should look like here.

packages/js-sdk/src/sandbox/sandboxApi.ts
```diff
@@ -39,6 +39,11 @@ export interface SandboxInfo {
    * Sandbox start time.
    */
   startedAt: Date
```
ValentaTomas, line 41: With the introduction of pause/resume it might make sense to explicitly mention if this is the original start time of the paused sandbox or the time it was started now.

ValentaTomas:
With snapshots we might want to return also the state when adding this.
I'm still considering if returning info object vs having methods for getting each field make sense.

ValentaTomas:
Added comments—we should resolve and merge.

mishushakov:
I have rebased on main in #625, after it's merged I can add "state" as well

PR #519, opened 2024-12-23
Boilerplate for integration tests (within js-sdk)

0div:
# Description 
- [x] nextjs integration test template
- [x] `pnpm test:integration`
- [x] set of stress tests
- [x] adapt vitest workspace


packages/js-sdk/tests/setup.ts
```diff
@@ -1,7 +1,7 @@
 import { Sandbox } from '../src'
 import { test as base } from 'vitest'
 
-export const template = 'base'
+export const template = 'jonas_base'
```
jakubno:
```suggestion
export const template = 'base'
```

packages/js-sdk/package.json
```diff
@@ -35,7 +35,8 @@
     "update-deps": "ncu -u && pnpm i",
     "postPublish": "./scripts/post-publish.sh || true",
     "test:bun": "bun test tests/runtimes/bun --env-file=.env",
-    "test:deno": "deno test tests/runtimes/deno/ --allow-net --allow-read --allow-env --unstable-sloppy-imports --trace-leaks"
+    "test:deno": "deno test tests/runtimes/deno/ --allow-net --allow-read --allow-env --unstable-sloppy-imports --trace-leaks",
+    "test:integration": "E2B_INTEGRATION_TEST=1 vitest run tests/integration/**"
```
jakubno, line 39: can we do it only based on directory structure? :pray:

packages/js-sdk/tests/integration/template/e2b.Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM e2bdev/code-interpreter:latest
+
+# Clone the Next.js app repository
+RUN git clone https://github.com/ezesundayeze/basic-nextjs-app
```
jakubno: nothing against this, but wouldn't be something like this be better?
```
RUN npx -y create-next-app@latest test --yes --ts --use-npm
```

packages/js-sdk/tests/integration/stress.test.ts
```diff
@@ -0,0 +1,117 @@
...
+              console.log('fetching url', url)
+              fetch(url)
+            })
+
+            try {
+              sbx.kill()
+            } catch (e) {
+              console.error('error killing sbx', e)
+            }
+          })
```
jakubno: could you do async function, so it's easier to read?

packages/js-sdk/tests/integration/stress.test.ts
```diff
@@ -0,0 +1,117 @@
...
+              try {
+                resolve(sbx.getHost(8000))
+              } catch (e) {
+                console.error('error getting sbx host', e)
+                reject(e)
+              }
+            }).then((host) => {
+              const url = `https://${host}`
+              console.log('fetching url', url)
+              fetch(url)
```
jakubno: this only lists the files right? Is this testing something different than 1 or 3 does?

packages/js-sdk/package.json
0div: IIRC, I initially tried that, but the way vitest workspace works makes it impossible to exclude the integration tests when when running `pnpm test` , bc that just picks up everything — maybe i'm missing something tho!

packages/js-sdk/tests/integration/stress.test.ts
0div: oh this is not supposed to be here, it's a remnant of a little of both tests above and below, ill remove

packages/js-sdk/tests/integration/stress.test.ts
0div: oh this is not supposed to be here, it's a remnant of a little of both tests above and below, ill remove

jakubno (approved):

PR #524, opened 2024-12-27
Add get_metrics method to python-sdk

0div:


0div:
consolidated here: https://github.com/e2b-dev/E2B/pull/543

PR #532, opened 2025-01-10
Add filtering option for listing sandboxes

jakubno:
# Description

Allow users to filter out sandbox based on metadata

### DIFF
--- a/packages/js-sdk/src/sandbox/sandboxApi.ts
+++ b/packages/js-sdk/src/sandbox/sandboxApi.ts
@@ -11,6 +11,13 @@ export interface SandboxApiOpts
     Pick<ConnectionOpts, 'apiKey' | 'debug' | 'domain' | 'requestTimeoutMs'>
   > {}
 
+export interface SandboxListOpts extends SandboxApiOpts {
+  /**
+   * Filter the list of sandboxes by metadata, e.g. `{"key": "value"}`
+   */
+  filters?: Record<string, string>
+}
+
 /**
  * Information about a sandbox.
  */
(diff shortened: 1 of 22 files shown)


ValentaTomas:
(Blocked by https://github.com/e2b-dev/infra/pull/235 right now.)

mishushakov:
the only confusing part i find is that you call it "filters" instead of "metadata"

mlejva:
This looks like it's missing updates to docs. Please add it.

jakubno:
They are already here https://e2b.dev/docs/sandbox/list#filtering-sandboxes

packages/js-sdk/src/sandbox/sandboxApi.ts
```diff
@@ -99,15 +99,15 @@ export class SandboxApi {
...
+    let metadata = undefined
+    if (opts?.metadata) {
+      const encodedPairs: Record<string, string> = Object.fromEntries(Object.entries(opts.metadata).map(([key, value]) => [encodeURIComponent(key),encodeURIComponent(value)]))
+      metadata = new URLSearchParams(encodedPairs).toString()
     }
 
     const res = await client.api.GET('/sandboxes', {
         params: {
-          query: {query},
+          query: {metadata},
```
mishushakov, line 112: query should be metadata

mishushakov (changes requested):
Isn't the DX supposed to be like

```
Sandbox.list({ query: {metadata: {'a': 'b'}}})
```

Right now it is 

```
Sandbox.list({metadata: {'a': 'b'}})
```

Also, there should be updates to the (beta?) docs, since this is now changed

packages/python-sdk/e2b/sandbox_async/sandbox_api.py
```diff
@@ -28,6 +30,9 @@ async def list(
         List all running sandboxes.
 
         :param api_key: API key to use for authentication, defaults to `E2B_API_KEY` environment variable
+        :param filters: Filter the list of sandboxes by metadata, e.g. `{"key": "value"}`, if there are multiple filters they are combined with AND.
```
mishushakov: :param metadata

packages/js-sdk/src/sandbox/sandboxApi.ts
jakubno: that's query as a part of an url (the part after `?`)

mishushakov: ahh, yes

packages/python-sdk/e2b/sandbox_async/sandbox_api.py
jakubno: Fixed

mlejva:
Explicitly confirming what Mish said, it should be:

```
Sandbox.list({ query: {metadata: {'a': 'b'}}})
```

and similarly in Python

mishushakov (approved):

PR #533, opened 2025-01-10
Show badge for legacy documentation search results

ben-fornefeld:
This pr adds a `badge` attribute to documentation indices, which can represent a string literal for certain badge types defined in the `Result` type. 

Possible types at the time of writing: `"Legacy" | undefined`

On each documentation query, we sort all search results containing a badge value of `Legacy` at last and render a badge containing the string value in each search result item.

### DIFF
--- a/apps/web/src/components/Search.tsx
+++ b/apps/web/src/components/Search.tsx
@@ -253,6 +253,13 @@ function SearchResult({
           <HighlightQuery text={contextPreview} query={query} />
         </div>
       )}
+      {result.badge === 'Legacy' && (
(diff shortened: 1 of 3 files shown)


mlejva:
Search looks good. Can you also add a banner/some signal showing the page itself is legacy docs if a user visits it?

mlejva:
> Search looks good. Can you also add a banner/some signal showing the page itself is legacy docs if a user visits it?

Per our private convo - let's leave this for a separate PR

mlejva (approved):
Ready to go to prod!

PR #540, opened 2025-01-16
Test new error handling indicating when the sandbox is not found 

0div:
Added a test in `js-sdk` for https://github.com/e2b-dev/infra/pull/231


packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -21,3 +24,36 @@ sandboxTest('ping server in sandbox', async ({ sandbox }) => {
     }
   }
 })
+
+// TODO: unskip when we have this new error handling deployed
+sandboxTest.skip('ping server in non-running sandbox', async () => {
```
jakubno:
```suggestion
sandboxTest('ping server in non-running sandbox', async () => {
```

packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -21,3 +24,36 @@ sandboxTest('ping server in sandbox', async ({ sandbox }) => {
...
+
+    const host = sbx.getHost(8000)
+
+    const res = await fetch(`${isDebug ? 'http' : 'https'}://${host}`)
+
+    assert.equal(res.status, 200)
+
+    await sbx.kill()
+
+    const res2 = await fetch(`${isDebug ? 'http' : 'https'}://${host}`)
```
jakubno: We can use envd here, so we don't have to wait for the server to start, maybe we could do the same above.
```suggestion
  // Envd port
  const host = sbx.getHost(49983)

  const url = `${isDebug ? 'http' : 'https'}://${host}/health`
  const res = await fetch(url)

    assert.equal(res.status, 204)

    await sbx.kill()

    const res2 = await fetch(url)
```

jakubno (changes requested):
Could we do the same for Python?

packages/js-sdk/tests/sandbox/host.test.ts
0div: I don't think we would get the wanted `nginx` response with local envd

jakubno: You're right, skip this test if debug 🙏🏻

packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -21,3 +24,35 @@ sandboxTest('ping server in sandbox', async ({ sandbox }) => {
     }
   }
 })
+
+sandboxTest('ping server in non-running sandbox', async () => {
+  const sbx = await Sandbox.create(template, { timeoutMs: 60_000 })
```
jakubno: Use the sandbox from the test case
```suggestion
sandboxTest('ping server in non-running sandbox', async ({ sandbox }) => {
```

0div:
> Could we do the same for Python?

This is more of an integration test rather than a unit test, do you think it's necessary?

0div:
> > Could we do the same for Python?
> 
> This is more of an integration test rather than a unit test, do you think it's necessary?

@jakubno maybe relevant to that comment: https://github.com/e2b-dev/E2B/pull/519

packages/js-sdk/tests/sandbox/host.test.ts
0div: it's ok if i kill it?

jakubno: Is there a reason why to create it manually, the one from test case should be handled even though the test fails anywhere during execution, you don't want to care about such thing in the test

packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -1,18 +1,56 @@
...
 
     assert.equal(res.status, 200)
+
+    await sbx.kill()
+
+    const res2 = await fetch(`${isDebug ? 'http' : 'https'}://${host}`)
+    assert.equal(res2.status, 502)
+
+    const text = await res2.text()
+    assert.equal(text, 'Sandbox does not exist.')
```
jakubno: could you just ping envd as I suggested in previous comment, so we don't have to wait 10 seconds? I think it's nice we test other port as well, so I would keep it there, but no need for it here

0div: We would not get the expected "Sandbox does not exist." in that case, no?

.gitignore
```diff
@@ -292,4 +292,3 @@ cython_debug/
 # SDK reference artifacts
 sdk_ref/
 sdkRefRoutes.json
-spec/openapi_generated.yml
```
jakubno: Don't

packages/js-sdk/package.json
```diff
@@ -28,15 +28,14 @@
     "dev": "tsup --watch",
     "example": "tsx example.mts",
     "test": "vitest run",
-    "generate": "python ./../../spec/remove_extra_tags.py sandboxes templates && openapi-typescript ../../spec/openapi_generated.yml -x api_key --support-array-length --alphabetize --output src/api/schema.gen.ts",
+    "generate": "openapi-typescript ../../spec/openapi.yml -x api_key --support-array-length --alphabetize --output src/api/schema.gen.ts",
```
jakubno: Don't

packages/js-sdk/package.json
```diff
@@ -28,15 +28,14 @@
...
     "generate-envd-api": "openapi-typescript ../../spec/envd/envd.yaml -x api_key --support-array-length --alphabetize --output src/envd/schema.gen.ts",
     "generate-ref": "./scripts/generate_sdk_ref.sh",
     "check-deps": "knip",
     "update-deps": "ncu -u && pnpm i",
     "postPublish": "./scripts/post-publish.sh || true",
     "test:bun": "bun test tests/runtimes/bun --env-file=.env",
-    "test:deno": "deno test tests/runtimes/deno/ --allow-net --allow-read --allow-env --unstable-sloppy-imports --trace-leaks",
-    "test:integration": "E2B_INTEGRATION_TEST=1 vitest run tests/integration/**"
+    "test:deno": "deno test tests/runtimes/deno/ --allow-net --allow-read --allow-env --unstable-sloppy-imports --trace-leaks"
```
jakubno: Do you want to overwrite your changes?

packages/js-sdk/tests/sandbox/commands/sendStdin.test.ts
```diff
@@ -37,6 +37,8 @@ sandboxTest('send special characters to stdin', async ({ sandbox }) => {
 
   await sandbox.commands.sendStdin(cmd.pid, text)
 
+  await wait(5_000)
```
jakubno: This isn't probably needed, should be solved by the loop bellow, feel free to increase the loop range

packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -2,30 +2,60 @@ import { assert } from 'vitest'
...
+    try {
+      await wait(10_000)
 
-    const host = sandbox.getHost(8000)
+      const host = sandbox.getHost(8000)
 
-    let res = await fetch(`${isDebug ? 'http' : 'https'}://${host}`)
+      const res = await fetch(`${isDebug ? 'http' : 'https'}://${host}`)
 
-    for (let i = 0; i < 10; i++) {
```
jakubno: I would prefer this behavior, so we can pass the test faster

packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -2,30 +2,60 @@ import { assert } from 'vitest'
...
+      background: true,
+    })
+
     try {
-      await cmd.kill()
-    } catch (e) {
-      console.error(e)
+      await wait(10_000)
+
+      const host = sandbox.getHost(8000)
```
jakubno:
```suggestion
      const host = sandbox.getHost(49983)
```

packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -2,30 +2,60 @@ import { assert } from 'vitest'
...
+
+      const res = await fetch(`${isDebug ? 'http' : 'https'}://${host}`)
```
jakubno:
```suggestion
      const url = `${isDebug ? 'http' : 'https'}://${host}/health`
      const res = await fetch(url)
```

packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -2,30 +2,60 @@ import { assert } from 'vitest'
...
+
+      assert.equal(res.status, 200)
```
jakubno:
```suggestion
      assert.equal(res.status, 204)
```

packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -2,30 +2,60 @@ import { assert } from 'vitest'
...
+
+      await sandbox.kill()
+
+      const res2 = await fetch(`${isDebug ? 'http' : 'https'}://${host}`)
```
jakubno:
```suggestion
      const res2 = await fetch(url)
```

jakubno (changes requested):

packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -2,30 +2,60 @@ import { assert } from 'vitest'
...
+  async ({ sandbox }) => {
+    const cmd = await sandbox.commands.run('python -m http.server 8000', {
+      background: true,
+    })
+
     try {
-      await cmd.kill()
-    } catch (e) {
-      console.error(e)
+      await wait(10_000)
```
jakubno:
```suggestion
```

packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -2,30 +2,68 @@ import { assert } from 'vitest'
...
+)
+
+sandboxTest.skipIf(isDebug)(
+  'ping server in non-running sandbox',
+  async ({ sandbox }) => {
+    const cmd = await sandbox.commands.run('python -m http.server 8000', {
+      background: true,
+    })
+
     try {
```
jakubno:
```suggestion
```

packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -2,30 +2,68 @@ import { assert } from 'vitest'
...
+      assert.equal(res2.status, 502)
+
+      const text = await res2.text()
+      assert.equal(text, 'Sandbox does not exist.')
+    } finally {
+      try {
+        await cmd.kill()
+      } catch (e) {
+        console.error(e)
+      }
```
jakubno:
```suggestion
    const host = sandbox.getHost(49983)
    const url = `${isDebug ? 'http' : 'https'}://${host}/health`

    const res = await fetch(url)

    assert.equal(res.status, 204)

    await sandbox.kill()

    const res2 = await fetch(url)
    assert.equal(res2.status, 502)

    const text = await res2.text()
    assert.equal(text, 'Sandbox does not exist.')
```

jakubno (commented):
This is ready to merge, but probably not to main, because we're waiting to deploy API

ValentaTomas:
After we resolve the conflict this should be good to go—it is deployed on prod, right @jakubno ?

ValentaTomas:
Btw, what was the reason for not adding the Python version?

0div:
> Btw, what was the reason for not adding the Python version?

@ValentaTomas felt like we are doing more of an integration test here, asserting that the error handling on the backend works as expected, rather than how the SDKs handle this themselves.

packages/js-sdk/tests/sandbox/connect.test.ts
```diff
@@ -19,3 +19,25 @@ test('connect', async () => {
...
+    const isRunning = await sbx.isRunning()
+    assert.isTrue(isRunning)
+    await sbx.kill()
+    isKilled = true
+
+    const sbxConnection = await Sandbox.connect(sbx.sandboxId)
+    const isRunning2 = await sbxConnection.isRunning()
+    assert.isFalse(isRunning2)
+
+    await sbxConnection.commands.run('echo "hello"')
```
dobrac: shouldn't this run command always fail? It looks to me like there is a missing assert on the behavior what is expected here

packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -2,30 +2,56 @@ import { assert } from 'vitest'
...
     }
+  },
+  60_000
+)
+
+sandboxTest.skipIf(isDebug)(
+  'ping server in non-running sandbox',
+  async ({ sandbox }) => {
+    const host = sandbox.getHost(49983)
```
dobrac: the port would be better in a constant

dobrac (dismissed):
lg, just one small thing

following the conversation about this being more of an integration test, would it make sense now that it would be in the Infra package directly and not in SDK?

ValentaTomas:
@0div Is this still blocked by PR/deploy of main/prod?

0div:
> @0div Is this still blocked by PR/deploy of main/prod?

No blocked per se, as @dobrac points out it could be transposed to the new integration test suite in infra/

ValentaTomas:
@0div Let's just resolve the comments and merge—we can move to integration tests later.

packages/js-sdk/tests/sandbox/host.test.ts
0div: why exactly? it's only used here.

dobrac: The port 49983 is the envd port on which every sandbox can be accessed. I understand that you know it, but I don't consider it being quite obvious just from the number itself if you do have the whole infra context in mind

0div: i see, i'd rather not have to worry about this special port and explaining it; i went ahead and updated the test to make it independent of this.

jakubno: It looks like you are still using it

packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -2,30 +2,63 @@ import { assert } from 'vitest'
...
+    const res = await fetch(url)
+
+    assert.equal(res.status, 204)
+
+    await sandbox.kill()
+
+    const res2 = await fetch(url)
+    assert.equal(res2.status, 502)
+
+    const text = await res2.text()
```
jakubno: you can simplify here (there's already test for running sandbox)
```suggestion
    const host = sandbox.getHost(3000)
    const url = `https://{host}`

    await sandbox.kill()

    const res = await fetch(url)
    assert.equal(res.status, 502)

    const text = await res.text()
```

packages/js-sdk/tests/sandbox/connect.test.ts
```diff
@@ -19,3 +19,23 @@ test('connect', async () => {
...
+
+    const sbxConnection = await Sandbox.connect(sbx.sandboxId)
+    const isRunning2 = await sbxConnection.isRunning()
+    assert.isFalse(isRunning2)
+  } finally {
+    if (!isKilled) {
+      await sbx.kill()
+    }
+  }
+})
```
jakubno:
```suggestion
sandboxTest('connect to non-running sandbox', async ({ sandbox }) => {
  const isRunning = await sbx.isRunning()
  assert.isTrue(isRunning)
  await sbx.kill()

  const sbxConnection = await Sandbox.connect(sbx.sandboxId)
  const isRunning2 = await sbxConnection.isRunning()
  assert.isFalse(isRunning2)
})
```

packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -2,30 +2,63 @@ import { assert } from 'vitest'
 
 import { isDebug, sandboxTest, wait } from '../setup.js'
 
-sandboxTest('ping server in sandbox', async ({ sandbox }) => {
-  const cmd = await sandbox.commands.run('python -m http.server 8000', { background: true })
+sandboxTest.skipIf(isDebug)(
```
jakubno: this should work in debug right? maybe you could use nonstandard random port to prevent some port conflict 
```suggestion
sandboxTest(
```

jakubno (changes requested):

jakubno (approved):

packages/js-sdk/tests/sandbox/host.test.ts
0div: GH had some issues yesterday and wasn't getting my pushes to remote for some reason

PR #546, opened 2025-01-20
Fix default user for the user

jakubno:
# Description

The column was moved from `Team` to `UserTeam`, because one team can be the default for a specific user, but not to all of them. This is addressing the logic in dashboard


ben-fornefeld (approved):

mishushakov:
@ben-fornefeld can you check if this affects Fragments also?

PR #547, opened 2025-01-20
Change the default generated code for e2b.Dockerfile when initializing custom template

mlejva:
Everywhere in docs, we direct users to be using our code interpreter package. However when you run `e2b template init`, we still generate the `e2b.Dockerfile` with 
```
FROM ubuntu:22.04
```
which won't work with the code interpreter package.

This PR changes the generated code to
```
FROM e2bdev/code-interpreter:latest
```

### DIFF
--- a/.changeset/thick-trains-lick.md
+++ b/.changeset/thick-trains-lick.md
@@ -1,0 +1,5 @@
+---
+'@e2b/cli': minor
+---
+
+Change the default generated code for e2b.Dockerfile when initializing custom template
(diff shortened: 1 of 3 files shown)


jakubno (commented):
This still won't work, because of missing start_cmd

mlejva:
> This still won't work, because of missing start_cmd

Can you expand? This should work because users run `e2b template build -c "/root/.jupyter/start-up.sh"`.

A user just had this issue that they generated `e2b.Dockerfile` with `e2b template init` and then they ran `-c "/root/.jupyter/start-up.sh"`.

jakubno:
I see, in that case it's okay

jakubno:
I am thinking if we shouldn't do something in CLI, if it's only this Dockerfile and start command wasn't passed, add it automatically?

PR #548, opened 2025-01-23
Add envd version to SDKs

jakubno:
# Description

Add envd version when creating sandbox to allow for check if features are available


ValentaTomas:
What about when we are just reconnecting to the sandbox?

jakubno:
We need to add endpoint for that, but this is blocking and will end up in merge conflicts if we don't add it

0div (approved):
the  newly returned `-> SandboxCreateResponse:` didn't break any tests?

jakubno:
why should it?

0div:
> why should it?

from memory there are a few tests that re-use the sandbox ID to call other class methods, but maybe they use the property.

jakubno:
This was change only in internals

PR #550, opened 2025-01-24
Process time out returns incorrect exception in pythonsdk e2b 1476

jakubno:
Handle the exception properly

### DIFF
--- a/packages/python-sdk/e2b/sandbox_sync/commands/command_handle.py
+++ b/packages/python-sdk/e2b/sandbox_sync/commands/command_handle.py
@@ -62,25 +62,28 @@ class CommandHandle:
         None,
         None,
     ]:
-        for event in self._events:
-            if event.event.HasField("data"):
-                if event.event.data.stdout:
-                    out = event.event.data.stdout.decode('utf-8', 'replace')
-                    self._stdout += out
-                    yield out, None, None
-                if event.event.data.stderr:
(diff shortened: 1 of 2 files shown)


ValentaTomas:
I think the exception handling is also missing in the async python implementation.

packages/python-sdk/e2b/sandbox_sync/commands/command_handle.py
```diff
@@ -62,25 +62,28 @@ def _handle_events(
...
+                    if event.event.data.pty:
+                        yield None, None, event.event.data.pty
+                if event.event.HasField("end"):
+                    self._result = CommandResult(
+                        stdout=self._stdout,
+                        stderr=self._stderr,
+                        exit_code=event.event.end.exit_code,
+                        error=event.event.end.error,
+                    )
+        except Exception as e:
```
ValentaTomas, line 85: Are we sure the incorrect exception originated from here?

jakubno: There's a test, you can try. It's from the Generator, we have the same handling in watch handle

Without the try/except block, you can see it's raised in connect library in the generator 
image

jakubno:
> I think the exception handling is also missing in the async python implementation.

The events are processed differently in Async, you can't iterate over the handle in async version

PR #564, opened 2025-01-30
Add filtering sandboxes by state in the SDKs

mishushakov:
Adds a new optional param to filter sandboxes by state

PR #565, opened 2025-01-30
Added sandbox state column when running `sbx ls` command

mishushakov:


mishushakov:
moved the changes to #564

PR #568, opened 2025-01-31
Add auto pause on timeout to SDK

dobrac:
**Changes the following**
- Adds auto pause to SDK, requires having `autoPause: true` and `auto_pause=True`
- Replaces sandbox `resume` with `connect`
- Updates sandbox timeout on connect if it's already running

**SDK Change**
The autoPause option is required, and has to be set to true.

JS
```js
Sandbox.create({ autoPause: true })
Sandbox.connect(sandboxId, {autoPause: true})
```

Python - Async
```py
AsyncSandbox.create(auto_pause=True)
AsyncSandbox.connect(sandbox_id, auto_pause=True)
```

Python - Sync
```py
Sandbox(auto_pause=True)
Sandbox.connect(sandbox_id, auto_pause=True)
```

For both Python implementations, the __exit__ now calls .pause() instead of .kill()

**Tasks**
- [x] Update interface of connect, resume, etc
- [x] Add tests for JS SDK
- [x] Add tests for Python SDK
- [ ] ~~Update web documentation~~ (not for now, until we know the final SDK DX)

**Temporary solution for connect**
There is a setTimeout for every connect to resolve issues where paused sandbox can immediately timeout. 

Options discussed:
         1. No set - never sure how long the sandbox will be running
         2. Always set the timeout in code - the user can't just connect to the sandbox
               without changing the timeout, round trip to the server time
         3. Set the timeout in resume on backend - side effect on error
         4. Create new endpoint for connect


ValentaTomas (approved):

ValentaTomas:
Everything is ok, but we are keeping the branch unmerged because the changes are only in preview.

ValentaTomas:
Merged via https://github.com/e2b-dev/E2B/pull/854

PR #578, opened 2025-02-11
Fix CLI build issue

jakubno:
Load update notifier dynamically

We needed update the package because of vulnerabilities

If we don't load it asynchronously there's following error.
```
packages/cli generate-ref: /home/runner/work/E2B/E2B/packages/cli/dist/index.js:55406
packages/cli generate-ref: var updateNotifier = __toESM(require("update-notifier"));
packages/cli generate-ref:                              ^
packages/cli generate-ref: Error [ERR_REQUIRE_ESM]: require() of ES Module /home/runner/work/E2B/E2B/node_modules/.pnpm/update-notifier@6.0.2/node_modules/update-notifier/index.js from /home/runner/work/E2B/E2B/packages/cli/dist/index.js not supported.
packages/cli generate-ref: Instead change the require of /home/runner/work/E2B/E2B/node_modules/.pnpm/update-notifier@6.0.2/node
```

### DIFF
--- a/.changeset/stale-owls-own.md
+++ b/.changeset/stale-owls-own.md
@@ -1,0 +1,5 @@
+---
+'@e2b/cli': patch
+---
+
+Fix ESM import for older node versions
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -1,7 +1,5 @@
 #!/usr/bin/env -S node --enable-source-maps
 
-import * as updateNotifier from 'update-notifier'
-
 import * as commander from 'commander'
 import * as packageJSON from '../package.json'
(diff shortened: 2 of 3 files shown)


ValentaTomas:
Are we sure this async loading works?<br>Also it might be possible to tweak [https://github.com/e2b-dev/E2B/blob/main/packages/cli/tsup.config.js#L5](https://github.com/e2b-dev/E2B/blob/main/packages/cli/tsup.config.js#L5) which was previously necessary for the `update-notifier` packaging to fix this without changing the actual code.

ValentaTomas (approved):

PR #580, opened 2025-02-12
Added tests for "killing" paused sandboxes

mishushakov:
- Adds tests for removing paused sandboxes using kill

### DIFF
--- a/packages/js-sdk/tests/api/kill.test.ts
+++ b/packages/js-sdk/tests/api/kill.test.ts
@@ -16,6 +16,7 @@ sandboxTest.skipIf(isDebug)('kill non-existing sandbox', async () => {
 
 sandboxTest.skipIf(isDebug)('kill paused sandboxes', async ({ sandbox }) => {
   const pausedSandbox = await sandbox.pause()
+  await sandbox.kill()
   const pausedSandboxId = pausedSandbox.split('-')[0] + '-' + '00000000'
 
   const list = await Sandbox.list()
--- a/packages/js-sdk/tests/api/kill.test.ts
+++ b/packages/js-sdk/tests/api/kill.test.ts
@@ -16,8 +16,9 @@ sandboxTest.skipIf(isDebug)('kill non-existing sandbox', async () => {
 
 sandboxTest.skipIf(isDebug)('kill paused sandboxes', async ({ sandbox }) => {
   const pausedSandbox = await sandbox.pause()
-  await Sandbox.kill(pausedSandbox)
   const pausedSandboxId = pausedSandbox.split('-')[0] + '-' + '00000000'
+
+  await Sandbox.kill(pausedSandbox)
 
   const list = await Sandbox.list()
   expect(list.length).toBeGreaterThan(0)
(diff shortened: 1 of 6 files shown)


packages/js-sdk/tests/api/kill.test.ts
```diff
@@ -7,9 +7,18 @@ sandboxTest.skipIf(isDebug)('kill existing sandbox', async ({ sandbox }) => {
...
   const list = await Sandbox.list()
-  expect(list.map(s => s.sandboxId)).not.toContain(sandbox.sandboxId)
+  expect(list.map((s) => s.sandboxId)).not.toContain(sandbox.sandboxId)
 })
 
 sandboxTest.skipIf(isDebug)('kill non-existing sandbox', async () => {
   await expect(Sandbox.kill('non-existing-sandbox')).resolves.toBe(false)
 })
+
+sandboxTest.skipIf(isDebug)('kill paused sandboxes', async ({ sandbox }) => {
```
dobrac, line 17: I don't see any .kill() call, is that desired?

packages/js-sdk/tests/sandbox/kill.test.ts
```diff
@@ -7,5 +7,13 @@ sandboxTest.skipIf(isDebug)('kill', async ({ sandbox }) => {
   await sandbox.kill()
 
   const list = await Sandbox.list()
-  expect(list.map(s => s.sandboxId)).not.toContain(sandbox.sandboxId)
+  expect(list.map((s) => s.sandboxId)).not.toContain(sandbox.sandboxId)
+})
+
+sandboxTest.skipIf(isDebug)('kill paused sandbox', async ({ sandbox }) => {
```
dobrac, line 13: I don't see any .kill() call, is that desired?

packages/python-sdk/tests/async/api_async/test_sbx_kill.py
```diff
@@ -14,3 +14,14 @@ async def test_kill_existing_sandbox(async_sandbox: AsyncSandbox):
 @pytest.mark.skip_debug()
 async def test_kill_non_existing_sandbox():
     assert await AsyncSandbox.kill("non-existing-sandbox") == False
+
+
+@pytest.mark.skip_debug()
+async def test_kill_paused_sandbox(async_sandbox: AsyncSandbox):
+    paused_sandbox = await async_sandbox.pause()
+    paused_sandbox_id = paused_sandbox.split("-")[0] + "-" + "00000000"
```
dobrac: Why is the sandbox_id transformation necessary? Shouldn't the kill work for the id directly?

mishushakov: The reason is because .pause returns a sandbox id with a client id (for caching reasons when you call .resume alter), which we don't store in the snapshots table, so what we do instead is return zeroed-out client id on the backend

mishushakov: (but yeah, .kill should work with the id, there's no difference though as the /kill endpoint strips the client id anyways)

mishushakov:
keep in mind, this is blocked by https://github.com/e2b-dev/infra/pull/264 now

packages/python-sdk/tests/async/api_async/test_sbx_kill.py
dobrac: Right, if this just for the list verification, could you make the transformation here (or around) `assert paused_sandbox_id not in [s.sandbox_id for s in list]` and not before for the kill? As it invokes that you need to nullify the clientId in order for the kill to work

mishushakov: Done

mishushakov:
I have to update the PR still with some more test cases from infra

mishushakov:
Outdated

PR #600, opened 2025-03-05
CommandHandle onPty callback

Gioxxy:
This PR sets the **CommandHandle** `onPty` callback in the **Commands** `start` and `connect` methods for the JavaScript SDK. This allows streaming stdout using `onData` in real-time, rather than waiting for the process to complete.

### DIFF
--- a/packages/js-sdk/src/sandbox/commands/index.ts
+++ b/packages/js-sdk/src/sandbox/commands/index.ts
@@ -63,6 +63,10 @@ export interface CommandStartOpts extends CommandRequestOpts {
    */
   onStderr?: (data: string) => void | Promise<void>
   /**
+  * Callback for stdout output.
+  */
+  onData?: (data: Uint8Array) => void | Promise<void>
+  /**
    * Timeout for the command in **milliseconds**.
    * 
    * @default 60_000 // 60 seconds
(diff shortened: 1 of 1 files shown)


mlejva:
Hey @Gioxxy, thanks for the PR. 

In the PR's title you're talking about `onPty` but you added the `onData` callback. Is this intended?

Gioxxy:
> Hey @Gioxxy, thanks for the PR.
> 
> In the PR's title you're talking about `onPty` but you added the `onData` callback. Is this intended?

Hi @mlejva, I noticed that `onData` was already being used on the [Pty](https://github.com/e2b-dev/E2B/blob/ac87d9df35874d908828c58875a4b99edd028077/packages/js-sdk/src/sandbox/commands/pty.ts#L122-L130), so I kept the same naming for consistency. But I’m open to changing it!

ValentaTomas:
If we don't pass PTY information to the handle on server (https://github.com/e2b-dev/infra/blob/main/packages/envd/internal/services/process/handler/handler.go#L108) as we do in the Pty in SDK (https://github.com/Gioxxy/E2B/blob/feature/commands-on-data/packages/js-sdk/src/sandbox/commands/pty.ts#L100) the `onPty` won't ever return any data and there should be no reason for exposing this in the command options.

ValentaTomas:
Closing with the above reason for now.

PR #616, opened 2025-03-19
Include build ID output for debugging purposes

r33drichards:


packages/cli/src/commands/template/build.ts
```diff
@@ -397,7 +397,7 @@ export const buildCommand = new commander.Command('build')
         console.log(
           `> Triggered build for the sandbox template ${asFormattedSandboxTemplate(
             template
-          )} `
+          )} buildID: ${template.buildID}`
```
r33drichards:
```suggestion
          )} with build ID: ${template.buildID}`
```

ValentaTomas (approved):

PR #622, opened 2025-03-21
Add table column to show template `updatedAt` field

ziruihao:
When iterating on templates, our team often wanted to sanity check when a template was modified. So we end up opening the network log and reading the `templates` object resp manually.

If you've gotten this feedback from other users, here's a PR to show `updatedAt` as a column.


ValentaTomas:
@ben-fornefeld Dashboard feedback—this might be incorporated into the new one already?

ben-fornefeld:
> @ben-fornefeld Dashboard feedback—this might be incorporated into the new one already?

yes it will be incorporated. although tests are failing because of different reasons, this looks good. we might as well just merge this until the launch.

ben-fornefeld (approved):

PR #623, opened 2025-03-21
Add temporary docs redirects 

ben-fornefeld:
on our landing page & documentation we have references to outdated urls. for an upcoming event, redirects should be in place to avoid broken links. the redirects should only be temporary and the broken references should be fixed when resources are available to do so

### DIFF
--- a/apps/web/next.config.mjs
+++ b/apps/web/next.config.mjs
@@ -82,6 +82,20 @@ const nextConfig = {
       ],
     }
   },
+  async redirects() {
+    return [
+      {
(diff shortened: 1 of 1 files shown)


jamesmurdza:
@ben-fornefeld Did you mean to stage stress.test.js?

ben-fornefeld:
> @ben-fornefeld Did you mean to stage stress.test.js?

no thanks, for reviewing!

mishushakov (approved):

PR #632, opened 2025-03-22
fix: type improvement of React Component `children` prop

samuel871211:


mishushakov:
Hey, thanks for the PR - I believe the development of the dashboard has moved to https://github.com/e2b-dev/dashboard

PR #635, opened 2025-03-24
Change default domain in sdks from `e2b.dev` to `e2b.app`

sitole:


mishushakov (approved):
I have approved the code changes, but also make sure to check if there's anything needed to be changed in the documentation and CLI.

jakubno (approved):

jakubno:
There's no public url in CLI and in docs we only state how to get the url. There's  nothing about how it should look like.

PR #640, opened 2025-03-25
Allow writing BufferedReader

0div:
- [x] update `write` method in sync & async python SDKs to allow `io.BufferedReader` files
- [x] update tests

### DIFF
--- a/packages/python-sdk/e2b/sandbox_sync/filesystem/filesystem.py
+++ b/packages/python-sdk/e2b/sandbox_sync/filesystem/filesystem.py
@@ -1,4 +1,4 @@
-from io import TextIOBase
+from io import TextIOBase, BufferedReader
 from typing import IO, Iterator, List, Literal, Optional, overload, Union
 from e2b.sandbox.filesystem.filesystem import WriteEntry
 
@@ -202,7 +202,7 @@ class Filesystem:
             file_path, file_data = file['path'], file['data']
             if isinstance(file_data, str) or isinstance(file_data, bytes):
                 httpx_files.append(('file', (file_path, file_data)))
-            elif isinstance(file_data, TextIOBase):
+            elif isinstance(file_data, TextIOBase) or isinstance(file_data, BufferedReader):
                 httpx_files.append(('file', (file_path, file_data.read())))
             else:
                 raise ValueError(f"Unsupported data type for file {file_path}")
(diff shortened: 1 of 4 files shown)


packages/python-sdk/e2b/sandbox_sync/filesystem/filesystem.py
```diff
@@ -202,7 +202,7 @@ def write(
             file_path, file_data = file['path'], file['data']
             if isinstance(file_data, str) or isinstance(file_data, bytes):
                 httpx_files.append(('file', (file_path, file_data)))
-            elif isinstance(file_data, TextIOBase):
+            elif isinstance(file_data, TextIOBase) or isinstance(file_data, BufferedReader):
```
ValentaTomas: It is possible that we can actually accept `RawIOBase`, `IOBase`, `BinaryIO` or just plain `IO` (not sure if the instance check for this works well)?
This way it could be as wide as the `IO` we let users pass (if it works properly for the inputs).

Also, `isinstance` might be accepting tuples so the `or` might not be necessary (have not tested this).

packages/python-sdk/tests/async/sandbox_async/files/test_write.py
```diff
@@ -20,6 +22,21 @@ async def test_write_file(async_sandbox: AsyncSandbox):
     read_content = await async_sandbox.files.read(filename)
     assert read_content == content
 
+async def test_write_binary_file(async_sandbox: AsyncSandbox):
+    filename = "test_write.txt"
+    text = "This is a test binary file."
+    # equivalent to `open("path/to/local/file", "rb")`
+    content = io.BufferedReader(io.BytesIO(text.encode("utf-8")))
```
ValentaTomas: This is linted for me with 

```
Argument of type "BytesIO" cannot be assigned to parameter "raw" of type "RawIOBase" in function "__init__"
  "BytesIO" is incompatible with "RawIOBase"PylancereportArgumentType
```

0div: `isinstance` now checks for any inheritance from the parent `IOBase` 

see https://tedboy.github.io/python_stdlib/_modules/io.html#RawIOBase

packages/python-sdk/e2b/sandbox_sync/filesystem/filesystem.py
0div: `isinstance` now checks for any inheritance from the parent `io.IOBase`

see https://tedboy.github.io/python_stdlib/_modules/io.html#RawIOBase

jakubno (approved):

PR #649, opened 2025-04-01
Add optional depth parameter to SDK file listing methods. 

0div:
# Description

Do not merge before https://github.com/e2b-dev/infra/pull/524 is merged and deployed.

The DX would look like this:
```js
const files = await sandbox.files.list(dirName, { depth: 3 })
```
- [x] update `envd` pb spec 
- [x] generate for `js-sdk`
- [x] update `js-sdk` `file.list` method to include optional depth param
- [x] update `js-sdk` tests
- [x] generate for `python-sdk`
- [x] update `python-sdk` `file.list` method to include optional depth param
  - [x] sync
  - [x] async
- [x] update `python-sdk` tests
  - [x] sync
  - [x] async


ValentaTomas:
Also, add docstrings for the new `depth` parameter.

I'll document the spec update process so it is not blocking.

ValentaTomas:
This is blocked by the unpredictable Python client generation, right?

0div:
> This is blocked by the unpredictable Python client generation, right?

@ValentaTomas yes, per discussion, im on a side-quest to build a predictable codegen, will apply it here and commit.

packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -339,10 +339,16 @@ export class Filesystem {
    *
    * @returns list of entries in the sandbox filesystem directory.
    */
-  async list(path: string, opts?: FilesystemRequestOpts): Promise<EntryInfo[]> {
+  async list(
+    path: string,
+    opts?: FilesystemRequestOpts & { depth?: number }
```
mishushakov: why not:

```ts
path: string,
depth?: number,
opts?: FilesystemRequestOpts
```

packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -339,10 +339,16 @@ export class Filesystem {
...
+  ): Promise<EntryInfo[]> {
     try {
       const res = await this.rpc.listDir(
-        { path },
+        {
+          path,
+          depth: opts?.depth ?? 1,
```
mishushakov, line 358: I think it's fine to remove ?? 1 since it will be handled on backend?

packages/js-sdk/tests/integration/next.test.ts
mishushakov, line 1: explain changes here?

packages/js-sdk/tests/sandbox/files/list.test.ts
```diff
@@ -1,23 +1,62 @@
...
+  await sandbox.files.makeDir(`${parentDirName}/subdir2`)
+  await sandbox.files.makeDir(`${parentDirName}/subdir1/subdir1_1`)
+  await sandbox.files.makeDir(`${parentDirName}/subdir1/subdir1_2`)
+  await sandbox.files.makeDir(`${parentDirName}/subdir2/subdir2_1`)
+  await sandbox.files.makeDir(`${parentDirName}/subdir2/subdir2_2`)
 
-  const files = await sandbox.files.list(dirName)
-  assert.equal(files.length, 0)
+  // explicit depth 0 (should default to 1)
+  const files0 = await sandbox.files.list(parentDirName, { depth: 0 })
```
mishushakov: what happens if I pass depth -1?

packages/js-sdk/tests/sandbox/files/list.test.ts
```diff
@@ -1,23 +1,62 @@
...
+  const parentDirName = 'test_directory'
 
-  await sandbox.files.makeDir(dirName)
+  await sandbox.files.makeDir(parentDirName)
+  await sandbox.files.makeDir(`${parentDirName}/subdir1`)
+  await sandbox.files.makeDir(`${parentDirName}/subdir2`)
+  await sandbox.files.makeDir(`${parentDirName}/subdir1/subdir1_1`)
+  await sandbox.files.makeDir(`${parentDirName}/subdir1/subdir1_2`)
+  await sandbox.files.makeDir(`${parentDirName}/subdir2/subdir2_1`)
+  await sandbox.files.makeDir(`${parentDirName}/subdir2/subdir2_2`)
```
mishushakov, line 15: maybe you can make a loop here

packages/js-sdk/tests/sandbox/files/list.test.ts
```diff
@@ -1,23 +1,62 @@
...
 
-  const files = await sandbox.files.list(dirName)
-  assert.equal(files.length, 0)
+  // explicit depth 0 (should default to 1)
+  const files0 = await sandbox.files.list(parentDirName, { depth: 0 })
+  console.log(files0)
```
mishushakov: remove console.log

packages/js-sdk/tests/sandbox/files/list.test.ts
```diff
@@ -1,23 +1,62 @@
...
+  assert.equal(files0.length, 2)
+  assert.equal(files0[0].name, 'subdir1')
+  assert.equal(files0[1].name, 'subdir2')
 
-  await sandbox.files.write('test_directory4/test_file', 'test')
+  // default depth (1)
+  const files = await sandbox.files.list(parentDirName)
+  console.log(files)
```
mishushakov: remove console.log

packages/js-sdk/tests/sandbox/files/list.test.ts
```diff
@@ -1,23 +1,62 @@
...
+  assert.equal(files[1].name, 'subdir2')
 
-  const files1 = await sandbox.files.list(dirName)
-  assert.equal(files1.length, 1)
-  assert.equal(files1[0].name, 'test_file')
-  assert.equal(files1[0].type, 'file')
-  assert.equal(files1[0].path, `/home/user/${dirName}/test_file`)
+  // explicit depth 1
+  const files1 = await sandbox.files.list(parentDirName, { depth: 1 })
+  console.log(files1)
```
mishushakov: remove console.log

packages/js-sdk/tests/sandbox/files/list.test.ts
```diff
@@ -1,23 +1,62 @@
...
+  assert.equal(files2[0].name, 'subdir1')
+  assert.equal(files2[1].name, 'subdir1_1')
+  assert.equal(files2[2].name, 'subdir1_2')
+  assert.equal(files2[3].name, 'subdir2')
+  assert.equal(files2[4].name, 'subdir2_1')
+  assert.equal(files2[5].name, 'subdir2_2')
+
+  // explicit depth 3 (should be the same as depth 2)
+  const files3 = await sandbox.files.list(parentDirName, { depth: 3 })
+  console.log(files3)
```
mishushakov: remove console.log

packages/js-sdk/tests/sandbox/files/list.test.ts
```diff
@@ -1,23 +1,62 @@
...
+  console.log(files1)
+  assert.equal(files1.length, 2)
+  assert.equal(files1[0].name, 'subdir1')
+  assert.equal(files1[1].name, 'subdir2')
 
-  const exists = await sandbox.files.exists(dirName)
-  assert.isTrue(exists)
+  // explicit depth 2
+  const files2 = await sandbox.files.list(parentDirName, { depth: 2 })
+  console.log(files2)
```
mishushakov: remove console.log

packages/js-sdk/tests/sandbox/files/list.test.ts
```diff
@@ -1,23 +1,62 @@
...
+  // explicit depth 3 (should be the same as depth 2)
+  const files3 = await sandbox.files.list(parentDirName, { depth: 3 })
+  console.log(files3)
+  assert.equal(files3.length, 6)
+  assert.equal(files3[0].name, 'subdir1')
+  assert.equal(files3[1].name, 'subdir1_1')
+  assert.equal(files3[2].name, 'subdir1_2')
+  assert.equal(files3[3].name, 'subdir2')
+  assert.equal(files3[4].name, 'subdir2_1')
+  assert.equal(files3[5].name, 'subdir2_2')
```
mishushakov: can this be a for loop somehow?

packages/python-sdk/e2b/sandbox_async/filesystem/filesystem.py
```diff
@@ -254,21 +254,23 @@ async def write(
     async def list(
         self,
         path: str,
+        depth: Optional[int] = 1,
```
mishushakov, line 256: should 1 be explicit?

packages/python-sdk/e2b/sandbox_sync/filesystem/filesystem.py
```diff
@@ -250,21 +250,23 @@ def write(
     def list(
         self,
         path: str,
+        depth: Optional[int] = 1,
```
mishushakov, line 253: should 1 be explicit?

packages/python-sdk/tests/async/sandbox_async/files/test_files_list.py
```diff
@@ -4,15 +4,53 @@
...
+
+    # explicit depth 3 (should be the same as depth 2)
+    files3 = await async_sandbox.files.list(parent_dir_name, depth=3)
+    assert len(files3) == 6
+    assert files3[0].name == "subdir1"
+    assert files3[1].name == "subdir1_1"
+    assert files3[2].name == "subdir1_2"
+    assert files3[3].name == "subdir2"
+    assert files3[4].name == "subdir2_1"
+    assert files3[5].name == "subdir2_2"
```
mishushakov: any way we could loop this?

packages/python-sdk/tests/sync/sandbox_sync/files/test_files_list.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # explicit depth 3 (should be the same as depth 2)
+    files3 = sandbox.files.list(parent_dir_name, depth=3)
+    assert len(files3) == 6
+    assert files3[0].name == "subdir1"
+    assert files3[1].name == "subdir1_1"
+    assert files3[2].name == "subdir1_2"
+    assert files3[3].name == "subdir2"
+    assert files3[4].name == "subdir2_1"
+    assert files3[5].name == "subdir2_2"
```
mishushakov: any way we could loop this?

mishushakov (changes requested):
Looks fine, the only things I am questioning if we should hardcode the default depth (any other examples where we hardcode a parameter like this?) and if we can reduce some repetition in the tests (using loops?)

packages/js-sdk/tests/integration/next.test.ts
0div: not meant for this branch, removed.

packages/js-sdk/src/sandbox/filesystem/index.ts
0div: i'd rather be defensive and pass an explicit value rather than `undefined`

packages/js-sdk/tests/sandbox/files/list.test.ts
0div:
```
InvalidArgumentError: [invalid_argument] unmarshal message: unmarshal into *filesystem.ListDirRequest: proto: (line 1:34): invalid val
ue for uint32 field depth: -1
```

added some tests to assert this error indeed gets thrown

packages/python-sdk/e2b/sandbox_async/filesystem/filesystem.py
0div: I'd rather it be explicit here than passing undefined

packages/js-sdk/src/sandbox/filesystem/index.ts
ValentaTomas: Default depth—having the depth 1 should work well for compatibility. Example of something similar could be the recursive.

```
path: string,
depth?: number,
opts?: FilesystemRequestOpts
```
I think a possible problem with this is that if you want to modify any spec then you need to define the depth parameter that was until then default. For this reason I generally don't like having two or more optional parameters.

packages/js-sdk/tests/sandbox/files/list.test.ts
mishushakov: Can you make the error more user-friendly? Check in the SDK

packages/js-sdk/src/sandbox/filesystem/index.ts
mishushakov: I see that it's a positional parameter instead of object, so in that case adding another one would be a breaking change.

Might be better then to extend `FilesystemRequestOpts` similar to `WatchOpts`? The current solution doesn't look very clean DX-vise.

packages/js-sdk/tests/sandbox/files/list.test.ts
```diff
@@ -1,23 +1,101 @@
...
+      ],
+    },
+    {
+      name: 'negative depth should error',
+      depth: -1,
+      expectedLen: 0,
+      expectedFiles: [],
+      expectError: 'invalid_argument',
+    },
+  ]
```
mishushakov: I think previous version was better, right now it's bit difficult to digest

packages/python-sdk/tests/async/sandbox_async/files/test_files_list.py
```diff
@@ -4,15 +4,98 @@
...
+            ],
+        },
+        {
+            "name": "negative depth should error",
+            "depth": -1,
+            "expected_len": 0,
+            "expected_files": [],
+            "expect_error": "Value out of range",
+        },
+    ]
```
mishushakov: same as my previous

packages/python-sdk/tests/sync/sandbox_sync/files/test_files_list.py
```diff
@@ -0,0 +1,101 @@
...
+            ],
+        },
+        {
+            "name": "negative depth should error",
+            "depth": -1,
+            "expected_len": 0,
+            "expected_files": [],
+            "expect_error": "Value out of range",
+        },
+    ]
```
mishushakov: same as my previous

mishushakov (changes requested):
Once tests improved and JS DX is agreed on we can merge

packages/js-sdk/src/sandbox/filesystem/index.ts
ValentaTomas: Yes, let's extend the type like `WatchOpts`.

packages/js-sdk/tests/sandbox/files/list.test.ts
0div: I think your initial feedback was correct, i find it just as if not more readable like this

packages/python-sdk/tests/async/sandbox_async/files/test_files_list.py
```diff
@@ -4,15 +4,98 @@
...
+
+    for test_case in test_cases:
+        if "expect_error" in test_case:
```
mishushakov: We can keep the loop, but I find the expect_error part very confusing now, can you just do it like in go?

packages/python-sdk/tests/sync/sandbox_sync/files/test_files_list.py
```diff
@@ -0,0 +1,101 @@
...
+
+    for test_case in test_cases:
+        if "expect_error" in test_case:
```
mishushakov: same as above

packages/js-sdk/tests/sandbox/files/list.test.ts
```diff
@@ -1,23 +1,101 @@
...
 
-  await sandbox.files.write('test_directory4/test_file', 'test')
+  for (const testCase of testCases) {
+    if (testCase.expectError) {
```
mishushakov: same as above

mishushakov (changes requested):

packages/python-sdk/tests/async/sandbox_async/files/test_files_list.py
0div: what is confusing about having an expected error in a test? would you word it differently ?

mishushakov: the whole block is difficult to read with if/else

mishushakov:
Since you extended FilesystemRequestOps, can you check if it's used in other places? We want depth only for list

0div:
> Since you extended FilesystemRequestOps, can you check if it's used in other places? We want depth only for list

it's used in quite a few places, i made it separate

packages/js-sdk/tests/integration/next.test.ts
mishushakov, line 1: is this file supposed to be here?

mishushakov (changes requested):
- remove, if not needed: packages/js-sdk/tests/integration/next.test.ts
 - add equivalent of InvalidArgumentError in JS to Python SDK

mishushakov (approved):
Looks good with the latest changes, let's wait for CI/infra update to make sure the test pass

packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -339,10 +346,17 @@ export class Filesystem {
    *
    * @returns list of entries in the sandbox filesystem directory.
    */
-  async list(path: string, opts?: FilesystemRequestOpts): Promise<EntryInfo[]> {
+  async list(path: string, opts?: FilesystemListOpts): Promise<EntryInfo[]> {
+    if (typeof opts?.depth === 'number' && opts.depth < 0) {
```
jakubno: Shouldn't this be at least 1? If 0 and 1 is the same, it's rather confusing

packages/python-sdk/e2b/sandbox_sync/filesystem/filesystem.py
```diff
@@ -250,21 +250,26 @@ def write(
...
         List entries in a directory.
 
         :param path: Path to the directory
+        :param depth: Depth of the directory to list
         :param user: Run the operation as this user
         :param request_timeout: Timeout for the request in **seconds**
 
         :return: List of entries in the directory
         """
+        if depth is not None and depth < 0:
```
jakubno: same here

packages/python-sdk/e2b/sandbox_async/filesystem/filesystem.py
```diff
@@ -254,21 +253,26 @@ async def write(
...
         List entries in a directory.
 
         :param path: Path to the directory
+        :param depth: Depth of the directory to list
         :param user: Run the operation as this user
         :param request_timeout: Timeout for the request in **seconds**
 
         :return: List of entries in the directory
         """
+        if depth is not None and depth < 0:
```
jakubno: and same here

packages/python-sdk/tests/async/sandbox_async/files/test_files_list.py
```diff
@@ -4,15 +4,88 @@
...
+
+    await async_sandbox.files.make_dir(parent_dir_name)
+    await async_sandbox.files.make_dir(f"{parent_dir_name}/subdir1")
+    await async_sandbox.files.make_dir(f"{parent_dir_name}/subdir2")
+    await async_sandbox.files.make_dir(f"{parent_dir_name}/subdir1/subdir1_1")
+    await async_sandbox.files.make_dir(f"{parent_dir_name}/subdir1/subdir1_2")
+    await async_sandbox.files.make_dir(f"{parent_dir_name}/subdir2/subdir2_1")
+    await async_sandbox.files.make_dir(f"{parent_dir_name}/subdir2/subdir2_2")
+
+    test_cases = [
```
jakubno, line 19: can you check the whole path, as there was bug in envd, in merging the path correctly

jakubno (changes requested):
JS test are failing

packages/python-sdk/tests/async/sandbox_async/files/test_files_list.py
0div: i added the whole path in tests, they fail against foxtrot now—i dont think envd was updated yet.

mishushakov: Did you rebuild the base template?

packages/python-sdk/e2b/sandbox_async/filesystem/filesystem.py
jakubno: still not changed

packages/js-sdk/tests/sandbox/files/list.test.ts
```diff
@@ -1,23 +1,91 @@
...
+      depth: 3,
+      expectedLen: 6,
+      expectedFiles: [
+        `${parentDirName}/subdir1`,
+        `${parentDirName}/subdir1/subdir1_1`,
+        `${parentDirName}/subdir1/subdir1_2`,
+        `${parentDirName}/subdir2`,
+        `${parentDirName}/subdir2/subdir2_1`,
+        `${parentDirName}/subdir2/subdir2_2`,
+      ],
```
jakubno: Unify the expected format, sometimes you have parent directory and sometimes you don't.

Also can you check for both `path` and  `name`? :pray:

packages/python-sdk/e2b/sandbox_async/filesystem/filesystem.py
0div: wait what the heck, this is the second time i push to this branch and it doesn't make it to the remote

packages/python-sdk/tests/async/sandbox_async/files/test_files_list.py
0div: Still having issues with that.

packages/js-sdk/tests/sandbox/files/list.test.ts
0div: - I'm not sure I understand? it's listing them like `tree` would, DFS style. 
- What do you mean by `path` and `name`?

jakubno: Currently we return `name`, `path` and `type`, `path` is absolute path, so it should stay as well otherwise it would be breaking change

jakubno (approved):

PR #656, opened 2025-04-02
Make possible to pass custom headers in the SDKs

mishushakov:
- Required for custom Supabase Auth implementation

### DIFF
--- a/packages/js-sdk/src/api/index.ts
+++ b/packages/js-sdk/src/api/index.ts
@@ -58,6 +58,7 @@ class ApiClient {
         ...(config.accessToken && {
           Authorization: `Bearer ${config.accessToken}`,
         }),
+        ...config.headers,
       },
     })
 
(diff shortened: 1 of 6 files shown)


dobrac:
Just to clarify, why did you choose to be able to override existing default headers?

mishushakov:
for my personal DX, hehe, it should only override the specific ones

jakubno:
What's the use case? How is it related to Supabase Auth implementation? Either we want to support it natively or we don't right?

mishushakov:
This is an escape hatch how one can pass custom headers (in my case X-Supabase-Token and X-Supabase-Team)

ValentaTomas:
This is for the API client, right? Not for the request to sandboxes?

mishushakov:
@ValentaTomas correct, not touching envd or jupyter

ValentaTomas (approved):
Looks good to me

PR #663, opened 2025-04-03
Forward user metadata is_fragments_user to hubspot

ben-fornefeld:
This pr forwards the is_fragments_user metadata field to our hubspot integration in new_user edge function

### DIFF
--- a/supabase/functions/new_user/index.ts
+++ b/supabase/functions/new_user/index.ts
@@ -36,13 +36,13 @@ function sendToLoops(body: any) {
   })
 }
 
-
 function createHubspotContact(userData: {
   id: string
   email: string
   raw_user_meta_data?: {
     full_name?: string
     name?: string
+    is_fragments_user?: boolean
   }
 }) {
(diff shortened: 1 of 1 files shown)


supabase/functions/new_user/index.ts
```diff
@@ -62,13 +62,19 @@ function createHubspotContact(userData: {
...
+
   // Map Supabase user data to HubSpot properties
   const hubspotContactData = {
     properties: {
-      db_user_id: userData.id,
       email: userData.email,
       firstname: firstName,
       lastname: lastName,
+      db_user_id: userData.id,
+      db_is_fragments_user: isFragmentsUser,
```
mishushakov, line 77: any particular reason these are prefixed with db_?

ben-fornefeld: > any particular reason these are prefixed with db_?

yes, in hubspot these properties are custom created. for usability on our side all db values will have this prefix in hubspot

PR #664, opened 2025-04-04
Make codegen robust and deterministic for js and python SDKs

0div:
# Description 

This PR allows to codegen without worrying about your environment and dependencies for consistent outcomes in what code is generated from openapi and envd protobuf spec. Will generate files for `js-sdk` and `python-sdk` 

- [x] create Docker image with all the pinned deps needed to codegen
- [x] create codegen script using this container with mapped volumes
- [x] adapt Makefiles to this new approach
- [x] adapt where we install the connect-python protobuf binary for this to work

# Test

```sh
make codegen # this command is similar to `make generate` but w/o the hassle

Generating SDK code from openapi and envd spec
cd packages/js-sdk && pnpm generate

> e2b@1.2.1 generate /workspace/packages/js-sdk
> python ./../../spec/remove_extra_tags.py sandboxes templates && openapi-typescript ../../spec/openapi_generated.yml -x api_key --arr
ay-length --alphabetize --output src/api/schema.gen.ts

✨ openapi-typescript 7.6.1
🚀 ../../spec/openapi_generated.yml → src/api/schema.gen.ts [44.2ms]
cd packages/js-sdk && pnpm generate-envd-api

> e2b@1.2.1 generate-envd-api /workspace/packages/js-sdk
> openapi-typescript ../../spec/envd/envd.yaml -x api_key --array-length --alphabetize --output src/envd/schema.gen.ts

✨ openapi-typescript 7.6.1
🚀 ../../spec/envd/envd.yaml → src/envd/schema.gen.ts [22.7ms]

[...]

cd packages/python-sdk && make generate-api

All done! ✨ 🍰 ✨
```


mishushakov:
I don't like that it is now required to run a docker container for this

0div:
> I don't like that it is now required to run a docker container for this. Suggesting adding "make init-codegen" command which will initialize all the dependencies required for codegen.

I always appreciate an opinion although it would be helpful to know why you don't like running a docker container for this. 
Regardless, with these changes you can still run the legacy commands as you are used to doing for codegen,

ValentaTomas:
@mish I think you needed to install global deps+also specific python version which would make the "local" install modify a lot of things in the system.

sitole:
> I don't like that it is now required to run a docker container for this. Suggesting adding "make init-codegen" command which will initialize all the dependencies required for codegen.

Its actually not required. Docker just setups env for running classic makefile scripts, nothing changed here.

As @ValentaTomas said, you need to have many global packages for go, python installed locally and we build requires specific versions so it can be very painful to set it up.

sitole:
Tested, just forced protobuf version to align with needed one, regenerated clients from current version.

sitole (approved):

Issue #531, opened 2025-01-09
Error running code on Vercel edge environment


olefirenko:
I got this error when running the code on Vercel edge:

```

SandboxError: 2: [unknown] Invalid redirect value, must be one of "follow" or "manual" ("error" won't be implemented since it does not make sense at the edge; use "manual" and check the response status code).
    at (node_modules/e2b/dist/index.mjs:319:15)
    at (node_modules/e2b/dist/index.mjs:1421:14)
    at (node_modules/e2b/dist/index.mjs:32:29) {
  name: 'SandboxError'
}

```
When I run the code locally it works fine, no issues. The code is creating the Word docx file and then I am trying to download it.

The library version is 1.0.4.

mishushakov:
Would you be able to proceed without using the edge function? Is that a hard requirement?

olefirenko:
We use it to stream responses from the LLM when tools are in use. I believe it is a very common and recommended environment by Vercel for handling streaming. Thank you for support!

jakubno:
Hey @olefirenko, after looking into it, it seems the package we're using for RPC isn't compatible with the edge environment. This was discussed in this thread: https://github.com/connectrpc/connect-es/pull/1082. I've tried reaching out there, but haven't received a response yet. It looks like we might need to address this ourselves. How urgent is this issue for you?

mishushakov:
Meanwhile, if this becomes a blocker - you might consider switching to the regular (non-edge) runtime.

I will also verify and document runtimes that are working with our SDK and for the ones which don't, I will explain the cause and offer mitigation steps. Thank you for your help here!

rychlis:
I made a small PoC with patching the connect-es issue on my side here https://github.com/rychlis/e2b-edge-test/

Planned to make a proper PR out this but didn't have time yet :) 

I would just take this change https://github.com/rychlis/connect-es/pull/1/files and make a pnpm patch inside e2b repo

mishushakov:
@rychlis did you also make a PR on connect-es?

ValentaTomas:
[...]

PR #669, opened 2025-04-05
Fix edge runtime issue - use redirect: follow fetch option

rychlis:
Works well in my Cloudflare PoC https://github.com/rychlis/e2b-edge-test/blob/main/src/index.ts
PoC is using these APIs work well:
- `Sandbox.create`
- `sandbox.commands.run` with background ture

I use follow as per @ValentaTomas comment here https://github.com/e2b-dev/E2B/issues/531#issuecomment-2779492867

Few notes:
- I use beta release so I based this PR on beta, can do separate PR to main if needed
- Maybe I can only change this behavior based on detected runtime, but I think it might be safe
- Didn't look into E2B testing suite yet. My PoC only tests APIs I listed, maybe there are other edge runtime issues I didn't uncover yet

Should fix https://github.com/e2b-dev/E2B/issues/531

### DIFF
--- a/.changeset/serious-rocks-exist.md
+++ b/.changeset/serious-rocks-exist.md
@@ -1,0 +1,5 @@
+---
+'e2b': patch
+---
+
+Added support for edge workers on cloudflare
--- a/packages/js-sdk/src/sandbox/index.ts
+++ b/packages/js-sdk/src/sandbox/index.ts
@@ -114,6 +114,17 @@ export class Sandbox extends SandboxApi {
       baseUrl: this.envdApiUrl,
       useBinaryFormat: false,
       interceptors: opts?.logger ? [createRpcLogger(opts.logger)] : undefined,
+      fetch: (url, options) => {
+        // Patch fetch to always use redirect: "follow"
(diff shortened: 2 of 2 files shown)


rychlis:
Works well in my PoC https://github.com/rychlis/e2b-edge-test/blob/main/src/index.ts - can confirm these APIs work well:
- `Sandbox.create`
- `sandbox.commands.run` with background ture

I use follow as per @ValentaTomas comment here https://github.com/e2b-dev/E2B/issues/531#issuecomment-2779492867

Few notes:
- I use beta release so I based this on beta, can do separate PR to main if needed
- Maybe I can only change this behavior based on detected runtime, but I think it might be safe
- Didn't look into E2B testing suite yet. My PoC only tests few APIs, maybe there are other edge runtime issues I didn't uncover yet https://github.com/rychlis/e2b-edge-test/blob/main/src/index.ts

rychlis:
When editing the file, eslint reformatted the file (due `editor.formatOnSave` or `editor.codeActionsOnSave` config in .vscode/settings.json) but I'm not sure if I have my local dev setup correctly so I didn't include these formating issues

PR #677, opened 2025-04-10
Updates SandboxApiOpts type with added headers field

mishushakov:
- Minor type fix


mishushakov:
Related to: https://github.com/e2b-dev/E2B/pull/656

PR #678, opened 2025-04-11
Fix minor issue in python filesystem SDK

p-templeton:
One should now be able to use sb.filesystem.write(WriteEntry[]) in a typesafe way.


mishushakov (approved):

mishushakov:
great find thank you!

PR #694, opened 2025-04-18
Autopause beta with list v2

0div:
- [x] js-sdk
  - [x] successful build?
  - [x] successful unit tests?
- [x] python-sdk
  - [x] successful build?
  - [x] successful unit tests?

### DIFF
--- a/packages/js-sdk/tests/integration/stress.test.ts
+++ b/packages/js-sdk/tests/integration/stress.test.ts
@@ -1,6 +1,6 @@
-import { test } from 'vitest'
+import { assert, test } from 'vitest'
 
-import Sandbox from '../../src/index.js'
+import Sandbox, { components, CommandResult } from '../../src/index.js'
 import { isIntegrationTest, wait } from '../setup.js'
 
 const heavyArray = new ArrayBuffer(256 * 1024 * 1024) // 256 MiB = 256 * 1024 * 1024 bytes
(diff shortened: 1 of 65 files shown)


mishushakov:
keep in mind in the API client, I had to manually overwrite get_v2_sandboxes function so the state is passed correctly to the API

Screenshot 2025-04-21 at 17 23 30

packages/js-sdk/tests/integration/stress.test.ts
mishushakov, line 1: why commented out?

mishushakov (commented):
was this line changed in sandbox api .py error handling?

```diff
- if res.status_code >= 300
+ if res.status_code >= 300 or isinstance(res.parsed, Error)
```

can res.parsed even be an error?

0div:
> was this line changed in sandbox api .py error handling?
> 
> ```diff
> - if res.status_code >= 300
> + if res.status_code >= 300 or isinstance(res.parsed, Error)
> ```
> 
> can res.parsed even be an error?

This seems in line with the open API spec, reponse can be `[200] model` or `[> 300] Error`

mishushakov:
will `handle_api_exception` process it correctly though?

ValentaTomas:
Quick question here—is this a problem of the API client not being able to compose the state field properly, because it is too complicated?<br>Another question—the override is *added* in our code, not in the generated code, right?

0div:
@ValentaTomas, GTM?

PR #695, opened 2025-04-18
Add option for proxying SDK requests to API and sandbox

ValentaTomas:
PR that allows you to pass `proxy` (of the type ProxyTypes) when calling methods in the SDK that send requests. Also, if you pass `proxy` when creating a sandbox, all the sandbox requests (called on the Sandbox/AsyncSandbox instance) will automatically use the same proxy.

The PR also contains a small bugfix—adding missing passing of limits and passing these directly instead of through transport to be able to pass the proxy parameter properly.

This change should also be compatible with the code interpreter—it uses the client transport, which uses the passed proxy if there is any.

### DIFF
--- a/packages/python-sdk/e2b/sandbox_sync/sandbox_api.py
+++ b/packages/python-sdk/e2b/sandbox_sync/sandbox_api.py
@@ -61,7 +61,8 @@ class SandboxApi(SandboxApiBase):
                 metadata = urllib.parse.urlencode(quoted_metadata)
 
         with ApiClient(
-            config, transport=HTTPTransport(limits=SandboxApiBase._limits)
+            config,
+            limits=SandboxApiBase._limits,
         ) as api_client:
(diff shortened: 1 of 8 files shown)


0div:
at a glance this looks good, it's nice that we can leverage the `httpx` clietn and its `ProxyTypes`

packages/python-sdk/e2b/sandbox_sync/sandbox_api.py
```diff
@@ -39,6 +39,7 @@ def list(
         :param debug: Enable debug mode, all requested are then sent to localhost
         :param request_timeout: Timeout for the request in **seconds**
         :param headers: Additional headers to send with the request
+        :param proxy: Proxy to use for the request
```
0div, line 42: i'd prob specify the type `httpx.ProxyType` here as the SDK ref won't show it explicitly

0div (approved):
all python-sdk tests pass, i realize it's hard to test the proxy itself but have you tested manually?

PR #702, opened 2025-04-24
Allow for short sandbox id in `e2b sbx connect`

jakubno:
# Description

Allow for short sandbox ID in CLI when using `e2b sandbox connect <sandbox-id>`


mishushakov (commented):
is it just connect where the issue is happening?

jakubno:
Yes, only connect and logs use sandbox ID and logs already works with both

mishushakov (approved):

PR #704, opened 2025-04-28
Unify passing of template used in tests and log background commands output

ValentaTomas:
- Show errors from background commands
- Unify passing of template used in tests


packages/js-sdk/tests/sandbox/commands/run.test.ts
```diff
@@ -30,7 +30,7 @@ sandboxTest('run with multiline string', async ({ sandbox }) => {
 })
 
 sandboxTest('run with timeout', async ({ sandbox }) => {
-  const cmd = await sandbox.commands.run('echo "Hello, World!"', { timeoutMs: 1000 })
+  const cmd = await sandbox.commands.run('echo "Hello, World!"', { timeoutMs: 4000 })
```
mishushakov, line 33: any particular reason this was increased

packages/js-sdk/tests/sandbox/host.test.ts
```diff
@@ -7,6 +7,12 @@ sandboxTest(
   async ({ sandbox }) => {
     const cmd = await sandbox.commands.run('python -m http.server 8000', {
       background: true,
+      onStdout: (data) => {
+        console.log('stdout', data)
+      },
+      onStderr: (data) => {
+        console.log('stderr', data)
+      },
```
mishushakov: was this added on purpose?

packages/js-sdk/tests/setup.ts
```diff
@@ -25,7 +25,7 @@ export const sandboxTest = base.extend<SandboxFixture>({
         }
       }
     },
-    { auto: true },
+    { auto: false },
```
mishushakov, line 27: this was not needed?

mishushakov (commented):
Looks fine, can you add also the changes to the py version? Should be in a similar manner

packages/js-sdk/tests/setup.ts
ValentaTomas: Yes, I think it was not needed as with `false` the sandbox would be created only if the fixture was actually invoked.

packages/js-sdk/tests/sandbox/host.test.ts
ValentaTomas: Yes—because in several places we are using background commands where we never check the output it happened to me that there was no `python` in the template, but the test was timing out on checking open port in sandbox, without any error indicating that the command was not successful.
Maybe a better way would be checking if the command ended.

packages/js-sdk/tests/sandbox/commands/run.test.ts
ValentaTomas: It was flaky on a non-cached template.

packages/js-sdk/tests/sandbox/host.test.ts
ValentaTomas: Replaced with async assert for commands

mishushakov:
Python tests are erroring out with

```
e2b.exceptions.SandboxException: 500: error while getting template: Template not found
```

packages/js-sdk/tests/cmdHelper.ts
```diff
@@ -0,0 +1,16 @@
+import { CommandHandle, CommandExitError } from '../src/index.js'
+import { assert } from 'vitest'
+
+export function waitForFailure(cmd: CommandHandle) {
```
jakubno: the name is rather confusing, it isn't waiting for failure

packages/python-sdk/tests/async/sandbox_async/commands/test_env_vars.py
```diff
@@ -9,7 +9,7 @@ async def test_command_envs(async_sandbox: AsyncSandbox):
 
 
 @pytest.mark.skip_debug()
-async def test_sandbox_envs(template: str):
+async def test_sandbox_envs(template):
```
jakubno, line 12: Why?

packages/python-sdk/tests/sync/sandbox_sync/commands/test_env_vars.py
```diff
@@ -9,7 +9,7 @@ def test_command_envs(sandbox: Sandbox):
 
 
 @pytest.mark.skip_debug()
-def test_sandbox_envs(template: str):
+def test_sandbox_envs(template):
```
jakubno, line 12: why?

packages/python-sdk/tests/conftest.py
```diff
@@ -54,3 +62,40 @@ def skip_by_debug(request, debug):
     if request.node.get_closest_marker("skip_debug"):
         if debug:
             pytest.skip("skipped because E2B_DEBUG is set")
+
+
+class Helpers:
+    @staticmethod
+    def wait_for_async_failure(cmd: AsyncCommandHandle):
```
jakubno: same issue as in js with confussing naming

packages/python-sdk/tests/async/sandbox_async/commands/test_env_vars.py
ValentaTomas: It is automatically inferred and was present in some places and not in others.

packages/python-sdk/tests/conftest.py
ValentaTomas: Changed naming.

PR #712, opened 2025-05-03
Fix and update integration test template

0div:
These changes were made while testing https://github.com/e2b-dev/infra/pull/603
- fix and install [stress-ng](https://github.com/ColinIanKing/stress-ng) in integration test template
- fix integration test e2b.toml


packages/js-sdk/tests/integration/template/e2b.toml
jakubno, line 1: this is the e2b `base` image, we definitely don't want to use that one

jakubno (dismissed):

packages/js-sdk/tests/integration/template/e2b.toml
0div: i thought the different template name `"integration-test-v1"` distinguishes them?

jakubno: It's just an alias to make it easier to understand what template it is.

PR #714, opened 2025-05-04
Remove HubSpot contact creation

ben-fornefeld:
This pr removes the HubSpot contact creation from our `/new_user` edge function.


jakubno (approved):
remove it also from

PR #733, opened 2025-05-15
CLI: Improve auth login typing + browser sign in response usage

ben-fornefeld:
This pr improves the auth login code, to be more declarative and realistic to the actual response that is sent by the dashboard sign in flow.

### DIFF
--- a/packages/cli/src/commands/auth/login.ts
+++ b/packages/cli/src/commands/auth/login.ts
@@ -44,7 +44,7 @@ export const loginCommand = new commander.Command('login')
       const signal = connectionConfig.getSignal()
       const config = new e2b.ConnectionConfig({
         accessToken: process.env.E2B_ACCESS_TOKEN || userConfig?.accessToken,
-        apiKey: process.env.E2B_API_KEY || userConfig?.teamApiKey,
+        apiKey: process.env.E2B_API_KEY,
       })
       const client = new e2b.ApiClient(config)
       const res = await client.api.GET('/teams', { signal })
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/auth/login.ts
```diff
@@ -35,16 +36,16 @@ export const loginCommand = new commander.Command('login')
...
         console.info('Login aborted')
         return
       }
 
       const signal = connectionConfig.getSignal()
       const config = new e2b.ConnectionConfig({
-        accessToken: process.env.E2B_ACCESS_TOKEN || userConfig?.accessToken,
-        apiKey: process.env.E2B_API_KEY || userConfig?.teamApiKey,
+        accessToken: process.env.E2B_ACCESS_TOKEN || signInResponse.accessToken,
+        apiKey: process.env.E2B_API_KEY,
```
dobrac: do you need to set it here?

packages/cli/src/commands/auth/login.ts
```diff
@@ -20,7 +20,8 @@ import { handleE2BRequestError } from '../../utils/errors'
 export const loginCommand = new commander.Command('login')
   .description('log in to CLI')
   .action(async () => {
-    let userConfig
+    let userConfig: UserConfig | null = null
```
dobrac, line 23: why is the variable defined at the top here?

dobrac (changes requested):

packages/cli/src/commands/auth/login.ts
```diff
@@ -61,9 +62,13 @@ export const loginCommand = new commander.Command('login')
         process.exit(1)
       }
 
-      userConfig.teamName = defaultTeam.name
-      userConfig.teamId = defaultTeam.teamID
-      userConfig.teamApiKey = defaultTeam.apiKey
+      userConfig = {
+        ...signInResponse,
```
dobrac: I think it might be better to be explicit here

PR #738, opened 2025-05-19
Ignore unknown fields from received messages (protobuf)

mishushakov:
(We might actually not need this if we version the methods instead)

context: https://github.com/e2b-dev/infra/pull/655#pullrequestreview-2851059703
reference: https://googleapis.dev/python/protobuf/latest/google/protobuf/message.html#google.protobuf.message.Message.MergeFrom


ValentaTomas:
@mishushakov We want to merge this—do we need to do some additional testing or everything should be ok?

mishushakov:
Okay, let me run the tests with the new envd locally and I will report to you if changing this is good.

ValentaTomas (approved):

PR #743, opened 2025-05-26
(Jules) Linting TS/Python in SDK PRs

mishushakov:
- added workflow that uses Biome for JS and Ruff for Python to lint SDKs
- listed existing files

_PR is AI-generated_


mishushakov:
Closed because was incomplete.

PR #750, opened 2025-05-29
Add option for a template ready check

dobrac:
## Description
Add option for a template ready check. You can specify `ready_cmd` which will be run until it exits with code 0. Max timeout is 5 minutes. After exit code 0, the template will be marked as ready.

If no ready command is defined, `sleep 20` is assumed when the start command is present, and `sleep 0` when none is defined. It is also possible to define ready command without any start command. This might be helpful if the filesystem contains self-booted services (e.g. using systemd).

```
ready_cmd = "echo 'Starting ready check'; sleep 40; echo 'Template is ready'"
```

This adds possibility to configure the default 20 seconds wait time for a start command.

New documentation can be found at `/docs/sandbox-template/ready-cmd`.

## Checklist
- [x] Make sure the https://github.com/e2b-dev/infra/pull/719 is deployed

### DIFF
--- a/apps/web/src/app/(docs)/docs/sandbox-template/page.mdx
+++ b/apps/web/src/app/(docs)/docs/sandbox-template/page.mdx
@@ -106,17 +106,20 @@ print(execution.stdout)
 ```
 </CodeGroup>
 
+## How it works
+Every time you are building a sandbox template, we create a container based on the [`e2b.Dockerfile`](/docs/sandbox-template#3-customize-e2b-dockerfile) file you create in the process.
+We extract the container's filesystem, do provisioning and configuration (e.g. installing required dependencies), and start a sandbox.
+We call this sandbox a _sandbox template_.
+
+Then, these steps happen:
+
+    1. We take the running sandbox template.
+    2. (Only if you specified the [start command](/docs/sandbox-template/start-cmd), otherwise this step is skipped) **Execute the start command and wait for readiness (by default 20 seconds)**. Readiness check can be configured using [ready command](/docs/sandbox-template/ready-cmd).
+    3. Snapshot the sandbox and make it ready for you to spawn it with the SDK.
 
 
+<Note title="Sandbox Snapshot">
+  Snapshots are saved running sandboxes. We serialize and save the whole sandbox's filesystem together with all the running processes in a way that can be loaded later.
 
-
-
-
-
-
-
-
-
-
-
-
+  This allows us to load the sandbox in a few hundred milliseconds any time later with all the processes already running and the filesystem exactly as it was.
+</Note>
--- a/apps/web/src/app/(docs)/docs/sandbox-template/page.mdx
+++ b/apps/web/src/app/(docs)/docs/sandbox-template/page.mdx
@@ -109,15 +109,15 @@ print(execution.stdout)
 ## How it works
 Every time you are building a sandbox template, we create a container based on the [`e2b.Dockerfile`](/docs/sandbox-template#3-customize-e2b-dockerfile) file you create in the process.
 We extract the container's filesystem, do provisioning and configuration (e.g. installing required dependencies), and start a sandbox.
-We call this sandbox a _sandbox template_.
 
 Then, these steps happen:
 
-    1. We take the running sandbox template.
+    1. We take the running sandbox.
     2. (Only if you specified the [start command](/docs/sandbox-template/start-cmd), otherwise this step is skipped) Execute the start command.
     3. Wait for readiness (by default 20 seconds if start command is specified, otherwise immediately ready). Readiness check can be configured using [ready command](/docs/sandbox-template/ready-cmd).
     4. Snapshot the sandbox and make it ready for you to spawn it with the SDK.
 
+We call this sandbox snapshot a _sandbox template_.
 
 <Note title="Sandbox Snapshot">
   Snapshots are saved running sandboxes. We serialize and save the whole sandbox's filesystem together with all the running processes in a way that can be loaded later.
--- a/apps/web/src/app/(docs)/docs/sandbox-template/page.mdx
+++ b/apps/web/src/app/(docs)/docs/sandbox-template/page.mdx
@@ -114,8 +114,9 @@ We call this sandbox a _sandbox template_.
 Then, these steps happen:
 
     1. We take the running sandbox template.
-    2. (Only if you specified the [start command](/docs/sandbox-template/start-cmd), otherwise this step is skipped) **Execute the start command and wait for readiness (by default 20 seconds)**. Readiness check can be configured using [ready command](/docs/sandbox-template/ready-cmd).
-    3. Snapshot the sandbox and make it ready for you to spawn it with the SDK.
+    2. (Only if you specified the [start command](/docs/sandbox-template/start-cmd), otherwise this step is skipped) Execute the start command.
+    3. Wait for readiness (by default 20 seconds if start command is specified, otherwise immediately ready). Readiness check can be configured using [ready command](/docs/sandbox-template/ready-cmd).
+    4. Snapshot the sandbox and make it ready for you to spawn it with the SDK.
 
 
 <Note title="Sandbox Snapshot">
--- a/apps/web/src/app/(docs)/docs/sandbox-template/ready-cmd/page.mdx
+++ b/apps/web/src/app/(docs)/docs/sandbox-template/ready-cmd/page.mdx
@@ -6,7 +6,7 @@ This way you can control how long should we wait for the start command or any sy
 
 ## How to add ready command
 
-You can specify the ready command inside the `e2b.toml` in the same directory where you run `e2b build`.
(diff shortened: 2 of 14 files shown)


packages/cli/src/utils/errors.ts
```diff
@@ -34,6 +36,9 @@ export function handleE2BRequestError<T>(
     case 500:
       message = 'internal server error'
       break
+    default:
+      message = status(code) || 'unknown error'
```
dobrac, line 40: translate also all other error codes to text, so it doesn't show `unknown error`

packages/js-sdk/package.json
```diff
@@ -28,7 +28,7 @@
     "dev": "tsup --watch",
     "example": "tsx example.mts",
     "test": "vitest run",
-    "generate": "python ./../../spec/remove_extra_tags.py sandboxes templates && openapi-typescript ../../spec/openapi_generated.yml -x api_key --array-length --alphabetize --output src/api/schema.gen.ts",
+    "generate": "python ./../../spec/remove_extra_tags.py sandboxes templates auth && openapi-typescript ../../spec/openapi_generated.yml -x api_key --array-length --alphabetize --output src/api/schema.gen.ts",
```
dobrac, line 31: this is just to correctly regenerate the openapi spec with /teams endpoint

apps/web/src/app/(docs)/docs/sandbox-template/ready-cmd/page.mdx
```diff
@@ -0,0 +1,22 @@
+# Ready Command
+The ready command allows you to specify a command that will determine **template sandbox** readiness before creating the snapshot.
```
mlejva: > readiness before creating the snapshot

Unsure if we explain to the user what snapshot is and when it gets created during the build process. If not, we need  to explain it before we start mentioning it.

apps/web/src/app/(docs)/docs/sandbox-template/ready-cmd/page.mdx
```diff
@@ -0,0 +1,22 @@
...
+The ready command is executed in an infinite loop until it returns a successful **exit code&nbsp;0**.
+This way you can control how long should we wait for the start command or any system state.
```
mlejva:
```suggestion
This way you can control how long should we wait for the [start command](/docs/sandbox-template/start-cmd) or any system state.
```

apps/web/src/app/(docs)/docs/sandbox-template/ready-cmd/page.mdx
```diff
@@ -0,0 +1,22 @@
...
+You can specify the ready command inside the `e2b.toml` in the same directory where you run `e2b template build`.
+<CodeGroup isFileName title="e2b.toml" isRunnable={false}>
+  ```toml
+  # This is a config for E2B sandbox template
+  template_id = "1wdqsf9le9gk21ztb4mo"
+  dockerfile = "e2b.Dockerfile"
+  template_name = "my-agent-sandbox"
+  ready_cmd = "<your-ready-command>"  # $HighlightLine
+  ```
+</CodeGroup>
```
mlejva, line 25: Is the only way to set it through the config file? No option to do it via the CLI (similarly like we have start command option)?

apps/web/src/app/(docs)/docs/sandbox-template/ready-cmd/page.mdx
dobrac: We explain it in the start command. I'll reorganize this as it doesn't make sense to explain it in the start command and not in the sandbox customization (effectively template)

apps/web/src/app/(docs)/docs/sandbox-template/ready-cmd/page.mdx
dobrac: I've done it only through the config file as It didn't make any sense to me to have it configurable through command (as it's bound to the template, not specific build). But, the same applies for the start command too, so I'll add it there for parity. Hopefully this will be removed in the future.

jakubno: You can specify start command 
image

dobrac: Added `--ready-cmd` option, should I also add a short version? If so, maybe `-r`?

apps/web/src/app/(docs)/docs/sandbox-template/ready-cmd/page.mdx
dobrac: Moved the "How it works" to the sandbox template explanation rather than the start command + added link to the `snapshot` with the explanation

apps/web/src/app/(docs)/docs/sandbox-template/ready-cmd/page.mdx
```diff
@@ -0,0 +1,29 @@
+# Ready Command
+The ready command allows you to specify a command that will determine **template sandbox** readiness before a [snapshot](/docs/sandbox-template#how-it-works) is created.
+It is executed in an infinite loop until it returns a successful **exit code&nbsp;0**.
+This way you can control how long should we wait for the [start command](/docs/sandbox-template/start-cmd) or any system state.
+
+
+## How to add ready command
+
+When you are building a sandbox template you can specify the ready command by using the [`--ready-cmd`](/docs/sdk-reference/cli/v1.5.0/template#e2b-template-build) option:
```
dobrac, line 9: Not sure how to correctly approach this link, as it's going to properly exist only after this PR is merged

apps/web/src/app/(docs)/docs/sandbox-template/page.mdx
```diff
@@ -106,17 +106,21 @@ print(execution.stdout)
 ```
 </CodeGroup>
 
+## How it works
+Every time you are building a sandbox template, we create a container based on the [`e2b.Dockerfile`](/docs/sandbox-template#3-customize-e2b-dockerfile) file you create in the process.
+We extract the container's filesystem, do provisioning and configuration (e.g. installing required dependencies), and start a sandbox.
+We call this sandbox a _sandbox template_.
```
jakubno: I think this isn't the template yet. The snapshotted sandbox is the template.

apps/web/src/app/(docs)/docs/sandbox-template/ready-cmd/page.mdx
jakubno: It's annoying we can't specify something like `latest`

apps/web/src/app/(docs)/docs/sandbox-template/ready-cmd/page.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+```bash
```
jakubno, line 11: could you add some example, something like curl to your server returns 200? 🙏🏻

jakubno (commented):
Few small things and it's ready to go. Let's deploy it this week

apps/web/src/app/(docs)/docs/sandbox-template/page.mdx
dobrac: I've just copied what we have there currently, I'll fix it as you're correct, it's inaccurate

dobrac: updated

apps/web/src/app/(docs)/docs/sandbox-template/ready-cmd/page.mdx
dobrac: Added three examples

apps/web/src/app/(docs)/docs/sandbox-template/ready-cmd/page.mdx
```diff
@@ -0,0 +1,47 @@
...
+## Default values
+By default, the ready command is set to `sleep 0`, which means the sandbox template will be ready immediatelly.
+If the [start command](/docs/sandbox-template/start-cmd) is defined, the default is set to `sleep 20`, which means that the template sandbox will wait for 20 seconds before taking the snapshot.
+
+## Examples
+Here are some examples of the ready command you can use.
+
+### Wait for URL to return 200 status code
+```bash
+ready_cmd = 'curl -s -o /dev/null -w "%{http_code}" https://myurl.com | grep -q "200"'
```
jakubno:
```suggestion
ready_cmd = 'curl -s -o /dev/null -w "%{http_code}" http://localhost:3000 | grep -q "200"'
```

jakubno (approved):
Love the examples! Thanks

apps/web/src/app/(docs)/docs/sandbox-template/ready-cmd/page.mdx
mlejva: `--ready-cmd` is fine

mlejva (approved):

PR #753, opened 2025-06-01
Refactor: handle run command return based on opts.background

ben-fornefeld:
### What's Changed
Improved TypeScript type safety for the `Commands.run()` method by adding proper function overloads and conditional return types based on the `background` parameter, while maintaining full backwards compatibility.

### Key Improvements

**🔧 Better Type Safety**
- Added generic type parameter `<T extends boolean>` with conditional return type `RunCommandReturn<T>`
- TypeScript now correctly infers return types:
  - `background: true` → returns `CommandHandle`
  - `background: false | undefined` → returns `CommandResult`

**📝 Enhanced Developer Experience**
- Added comprehensive JSDoc overloads that clearly document each usage pattern
- IntelliSense now provides accurate type hints and autocompletion
- Eliminates the need for manual type casting when using the `run()` method

**🛡️ Prevents Runtime Errors**
- Developers can no longer accidentally call `CommandResult` methods on a `CommandHandle` or vice versa
- Compile-time type checking catches misuse before deployment

### Why This Matters for SDK Users

Before this change, developers had to manually cast types or deal with union types when using `commands.run()`:
```typescript
// Before: unclear return type, required manual casting
const result = await commands.run('ls', { background: true }) as CommandHandle
```

After this change, TypeScript automatically infers the correct type:
```typescript
// After: TypeScript knows this is CommandHandle
const handle = await commands.run('ls', { background: true })
await handle.wait() // ✅ TypeScript knows .wait() is available

// TypeScript knows this is CommandResult  
const result = await commands.run('ls')
console.log(result.stdout) // ✅ TypeScript knows .stdout is available
```

This makes the SDK more intuitive and prevents common developer mistakes while maintaining 100% backwards compatibility with existing codebases.

### DIFF
--- a/packages/js-sdk/src/sandbox/commands/index.ts
+++ b/packages/js-sdk/src/sandbox/commands/index.ts
@@ -6,7 +6,10 @@ import {
   Transport,
 } from '@connectrpc/connect'
 
-import { Signal, Process as ProcessService } from '../../envd/process/process_pb'
+import {
+  Signal,
+  Process as ProcessService,
+} from '../../envd/process/process_pb'
 import {
   ConnectionConfig,
   Username,
@@ -15,7 +18,7 @@ import {
   KEEPALIVE_PING_HEADER,
 } from '../../connectionConfig'
 import { authenticationHeader, handleRpcError } from '../../envd/rpc'
-import { CommandHandle, CommandResult } from './commandHandle'
+import { CommandResult, CommandHandle } from './commandHandle'
 import { handleProcessStartEvent } from '../../envd/api'
 export { Pty } from './pty'
 
@@ -23,7 +26,7 @@ export { Pty } from './pty'
  * Options for sending a command request.
  */
 export interface CommandRequestOpts
-  extends Partial<Pick<ConnectionOpts, 'requestTimeoutMs'>> { }
+  extends Partial<Pick<ConnectionOpts, 'requestTimeoutMs'>> {}
 
 /**
  * Options for starting a new command.
@@ -36,21 +39,21 @@ export interface CommandStartOpts extends CommandRequestOpts {
   background?: boolean
   /**
    * Working directory for the command.
-   * 
+   *
    * @default // home directory of the user used to start the command
    */
   cwd?: string
   /**
    * User to run the command as.
-   * 
+   *
    * @default `user`
    */
   user?: Username
   /**
    * Environment variables used for the command.
-   * 
+   *
    * This overrides the default environment variables from `Sandbox` constructor.
-   * 
+   *
    * @default `{}`
    */
   envs?: Record<string, string>
@@ -64,7 +67,7 @@ export interface CommandStartOpts extends CommandRequestOpts {
   onStderr?: (data: string) => void | Promise<void>
   /**
    * Timeout for the command in **milliseconds**.
-   * 
+   *
    * @default 60_000 // 60 seconds
    */
   timeoutMs?: number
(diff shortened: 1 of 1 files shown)


ben-fornefeld:
## RFC: Should we add runtime type discriminators to improve type safety?

### The Question
While this PR improves TypeScript inference, we should decide on also adding runtime type discriminators to make type casting more explicit and type-safe.

### Current Approach (This PR)
- Uses TypeScript overloads and conditional types
- Maintains 100% backwards compatibility
- Relies on compile-time type inference only

### Alternative Approach: Runtime Discriminators
Add a `type` property to distinguish between `CommandHandle` and `CommandResult`:

```typescript
// Option 1: Spread syntax (❌ Doesn't work)
{ type: "handle", ...commandHandle } // Loses all methods, only copies enumerable properties

// Option 2: Nested structure (⚠️ Breaking change)
{ type: "handle", handle: CommandHandle }
{ type: "result", result: CommandResult }
```

### Trade-offs

**Option 1 Problems:**
- JavaScript spread syntax only copies enumerable own properties
- All `CommandHandle` methods (`wait()`, `kill()`, `disconnect()`) would be lost
- Getters like `exitCode`, `stdout`, `stderr` would be lost
- Result would be essentially useless

**Option 2 Trade-offs:**
- ✅ Perfect type discrimination with `result.type === "handle"`
- ✅ All functionality preserved under nested structure
- ❌ **Breaking change**: `result.wait()` becomes `result.handle.wait()`
- ❌ All existing code needs updates
- ❌ More verbose API

```typescript
// Current way how to type cast:
const isBackground: boolean = Math.random() > 0.5
const result = await commands.run('ls', { background: isBackground }) // CommandHandle | CommandResult

// ✅ Is allowed, since the two return types share this property 
console.log(result.stdout)

// ❌ TypeScript complains, since we cannot be certain that `.disconnect()` exists here
console.log(result.disconnect())

if ("pid" in result) {
  // ✅ Is allowed, since TypeScript now is able to narrow down the type
  result.disconnect()
}
```

### Questions for Discussion
1. Is the improved TypeScript inference from this PR sufficient?
2. Would we be willing to accept breaking changes for better runtime type safety?
3. Are there other patterns we should consider?

packages/js-sdk/src/sandbox/commands/index.ts
```diff
@@ -292,33 +294,51 @@ export class Commands {
...
+   * @param opts options for starting the command.
+   *   - `opts.background: true` - runs in background, returns `CommandHandle`
+   *   - `opts.background: false | undefined` - waits for completion, returns `CommandResult`
+   *
+   * @returns Either a `CommandHandle` or a `CommandResult` (depending on `opts.background`).
+   */
+  async run(
+    cmd: string,
+    opts?: CommandStartOpts & { background?: boolean }
+  ): Promise<CommandHandle | CommandResult>
```
mishushakov, line 337: I don't understand why the overload is requires, can you make it work without pls.

mishushakov (changes requested):

mishushakov (approved):
Approved, the alternative would be to not use overloads but type inference, however we would be losing docstrings that way

mishushakov:
our convo for future reference:

> maybe i am blind, but aren't the two the same

> yes they are, but that is correct. this just states, that we also want to allow the user to use the function as it would be possible in runtime (with a boolean instead of only literals). the implementation has the same typing, just because it is the sum of the overloads that are enforced compile time. the implementation could also have any, but that would make our life harder working with the implementation

PR #755, opened 2025-06-03
Add: server side metadata generation in root layout

ben-fornefeld:
- adds `generateMetadata` in root layout to generate canonical and robots meta tags server side
- listens on a custom header which is set by our proxy before rewriting `/docs` routes, to decide whether the document should have no-index metadata or index metadata
- adds a middleware which appends a custom header to retrieve the current pathname for server side canonical tag generation


apps/web/src/app/layout.tsx
```diff
@@ -1,30 +1,42 @@
...
-  },
+    twitter: {
+      title: 'E2B - Code Interpreting for AI apps',
+      description: 'Open-source secure sandboxes for AI code execution',
+    },
+    openGraph: {
+      title: 'E2B - Code Interpreting for AI apps',
+      description: 'Open-source secure sandboxes for AI code execution',
+    },
+    alternates: pathname
```
dobrac: what if the pathname is ""?

apps/web/src/middleware.ts
```diff
@@ -0,0 +1,19 @@
+import { NextResponse } from 'next/server'
+import type { NextRequest } from 'next/server'
+
+export function middleware(request: NextRequest) {
+  const path = request.nextUrl.pathname
```
dobrac, line 5: is this conversion not necessary anymore?

`const pathWithoutTrailingSlash = pathname?.endsWith('/') ? pathname.slice(0, -1) : pathname`

apps/web/src/app/layout.tsx
ben-fornefeld: that case should not occur, since middleware only matches /docs/path* paths, but even if that occurs, the statement would be still correct afaik

dobrac: empty string is falsy in JS, so if it occurres, it will not set cannonical

apps/web/src/middleware.ts
ben-fornefeld: should be unnecessary according to: [https://nextjs.org/docs/app/api-reference/config/next-config-js/trailingSlash](https://nextjs.org/docs/app/api-reference/config/next-config-js/trailingSlash)

dobrac: Could you test it please? This is more for the canonical tag than functionality

ben-fornefeld: canonical tags don't really care in this regard. what is important to know is, that with or without trailing slash, the canonicals are treated as different pages. since we normalize to without trailing slash by default on the proxy and in this docs application, it does not matter.

dobrac: if we normalize them, then ok. Otherwise, as you said, it's quite important issue.

apps/web/src/app/layout.tsx
ben-fornefeld: i see, missed that here, good catch. it does not change something about the implementation though in this case

apps/web/src/middleware.ts
ben-fornefeld: yes, but that would be an important issue already one layer deeper. nextjs normalizes by default to not have users deal with these seo issues because of trailing slashes

dobrac: Ok, I was also asking, because it was there previously

apps/web/src/app/layout.tsx
dobrac: I think you should still fix it

ben-fornefeld: 1. the noindex would prevent the canonical from even being considered
2. we actually don't want to canonical on our landing page ever, so this acts as a second line of defense (although never reached)

ben-fornefeld: i've noticed a different bug though. when accessing a 404 page under `/docs`, this would still add the canonical for an invalid page:

`/docs/test `-> 404 under docs layout -> `<meta ... href='https://e2b.dev/docs/test"/>`

ben-fornefeld: resolved by [3964bce](https://github.com/e2b-dev/E2B/pull/755/commits/3964bcebca98dd8c6dd4eb67f2a6dcf8a4f61bc4)

dobrac: Two things:

1) The fix won't work for self hosting (if that's a concern)
2) About the pathname, I understand that with our using at the moment it's not an issue, but if you do few refactors and changes in the future, it might start being a silent issue. Just update the if condition to check for proper null/undefined value please, or make it explicit pathname !== "".

ben-fornefeld: > The fix won't work for self hosting (if that's a concern)

for self-hosting dashboard, this should be still fine, since the content would be the same + one would need to explicitly open a 404 /docs page. i made the redirect absolute on prod url, because i am not 100 sure if nextjs would redirect to the vercel (host) domain when accessing a 404 through the proxy. we definitely want to avoid that.

dobrac: I've meant the docs selfhosting, but it doesn't make much sense, so ok with me

PR #757, opened 2025-06-04
Fix: path validation for metadata tags resolution

ben-fornefeld:
This pr fixes validating pathnames for SEO metadata tag generation, by checking against already built time cached sitemap instead of doing request time glob resolution. path resolutions in vercels request-time are hard to optimize for & test, since the environment differs quite a bit from what could be tested locally + this should be faster since the fetch & sitemap are cached for the build.

### DIFF
--- a/apps/web/src/app/layout.tsx
+++ b/apps/web/src/app/layout.tsx
@@ -10,67 +10,28 @@ import { Layout } from '@/components/Layout'
 import { headers } from 'next/headers'
 
 async function isValidPath(pathname: string) {
-  const startTime = Date.now()
-  console.log('🔍 [isValidPath] Starting validation for:', pathname)
-  console.log('🌍 [isValidPath] Environment:', {
-    VERCEL_ENV: process.env.VERCEL_ENV,
-    VERCEL_URL: process.env.VERCEL_URL,
-    NODE_ENV: process.env.NODE_ENV,
-  })
-
   try {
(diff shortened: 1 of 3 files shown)


apps/web/src/app/layout.tsx
```diff
@@ -11,13 +11,27 @@ import { headers } from 'next/headers'
...
 async function isValidPath(pathname: string) {
   try {
-    const docsDirectory = await glob('**/*.mdx', {
-      cwd: `src/app/(docs)${pathname}`,
+    const baseUrl = process.env.VERCEL_URL
+      ? `https://${process.env.VERCEL_URL}`
+      : 'https://e2b.dev'
+    const sitemapUrl = `${baseUrl}/sitemap.xml`
+
+    const response = await fetch(sitemapUrl, {
```
dobrac, line 20: sitemap can consist of multiple files, could you please count with that?

dobrac (changes requested):

apps/web/src/app/layout.tsx
ben-fornefeld: as discussed per dm's: 

- making path validation resilient to handle the potential case of multiple sitemaps does not seem plausible in this application, since the porting the docs to our `e2b-dev/dashboard` repo will deprecate this application in the near future.
- added notation comments to inform engineers instead.

dobrac (approved):

PR #760, opened 2025-06-05
Refactor: prebuild sitemap generation instead of statically cached dynamic sitemap

ben-fornefeld:
Since we do not use dynamic data inside the sitemap and we base the sitemap generation on page files, we do want to avoid a re-build of the sitemap in request-time, since this environment differs significantly from a build-time environment. Hence we remove the dynamic sitemap generation and move sitemap generation logic inside our already present `prebuild.js` script


apps/web/prebuild.js
```diff
@@ -239,3 +240,120 @@ fs.writeFileSync(
...
+// Generate sitemap
+async function generateSitemap() {
+  try {
+    console.log('Generating sitemap...')
+    console.log('Current working directory:', process.cwd())
+
+    // Patterns to find all page.mdx files
+    // Account for different possible CWDs (main repo root vs apps/web)
+    const patterns = [
+      'src/app/\\(docs\\)/docs/**/page.mdx',  // Subdirectories
```
jakubno: wouldn't something like work here?
```
      '**/src/app/\\(docs\\)/docs/**/page.mdx',  // Subdirectories

File: apps/web/prebuild.js [hunk 2]
```diff
@@ -239,3 +240,120 @@ fs.writeFileSync(
...
+      .map((filePath) => {
+        try {
+          // Find the /docs/ segment and extract everything after it
+          const docsMatch = filePath.match(/\/app\/\(docs\)\/docs\/(.*)\/page\.mdx$/) ||
+                           filePath.match(/\/app\/\(docs\)\/docs\/page\.mdx$/)
+
+          if (!docsMatch) {
+            console.warn(`Unexpected file path format: ${filePath}`)
+            return null
+          }
```
jakubno: do we want silent fail here?

apps/web/prebuild.js
```diff
@@ -239,3 +240,120 @@ fs.writeFileSync(
...
+          return null
+        }
+      })
+      .filter(entry => entry !== null)
+      // Filter out legacy docs pages
+      .filter((entry) => !entry.url.includes('/docs/legacy'))
+
+    console.log(`Generated ${docsPages.length} sitemap entries`)
+
+    // Deduplicate URLs, keeping the entry with highest priority
```
jakubno: How can they duplicate if the path is unique?

apps/web/prebuild.js
```diff
@@ -239,3 +240,120 @@ fs.writeFileSync(
...
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+${finalEntries.map(entry => `  <url>
+    <loc>${entry.url}</loc>
+    <lastmod>${entry.lastModified}</lastmod>
+    <priority>${entry.priority}</priority>
+  </url>`).join('\n')}
+</urlset>`
+
+    // Write sitemap.xml to public directory (Next.js will serve it automatically)
+    const outputPath = path.join(__dirname, 'public', 'sitemap.xml')
```
jakubno: this won't work if you run it from root of the repo

apps/web/prebuild.js
```diff
@@ -242,99 +242,95 @@ fs.writeFileSync(
...
+    for (const filePath of mdxFiles) {
+      try {
+        // Find the /docs/ segment and extract everything after it
+        const docsMatch = filePath.match(/\/app\/\(docs\)\/docs\/(.*)\/page\.mdx$/) ||
+                         filePath.match(/\/app\/\(docs\)\/docs\/page\.mdx$/)
+
+        if (!docsMatch) {
+          processingErrors.push(`Unexpected file path format: ${filePath}`)
+          continue
+        }
```
jakubno: why not to fail the build, there's something wrong

apps/web/prebuild.js
```diff
@@ -242,99 +242,95 @@ fs.writeFileSync(
...
-
-    console.log(`Generated ${docsPages.length} sitemap entries`)
-
-    // Deduplicate URLs, keeping the entry with highest priority
-    const urlMap = new Map()
-    for (const entry of docsPages) {
-      const existing = urlMap.get(entry.url)
-      if (!existing || (existing.priority || 0) < (entry.priority || 0)) {
-        urlMap.set(entry.url, entry)
+    if (processingErrors.length > 0) {
```
jakubno: why won't you fail the build here, we're missing some pages, there isn't a reason why there should be any error

ben-fornefeld:
> Few things, check [here](https://e2b-web-git-fix-sitemap-generation.e2b-preview.dev/sitemap.xml).
> 
> 1. e2b.dev/docs is duplicated
> 2. could you remove api-reference folder, it's probably forgotten?

thanks for checking again, i've overseen this one. it would have been caught by duplication check in our proxy, but nice catch

jakubno (approved):

PR #765, opened 2025-06-07
RPC transport include default headers

mishushakov:
- Modified rpc transport to send default headers like user agent, client version, etc.
- We want this to be able to change backend logic in a non-breaking way for older clients

Browser tests failing due to https://github.com/e2b-dev/infra/pull/744


packages/js-sdk/src/sandbox/index.ts
```diff
@@ -171,7 +184,12 @@ export class Sandbox extends SandboxApi {
         apiUrl: this.envdApiUrl,
         logger: opts?.logger,
         accessToken: this.envdAccessToken,
-        headers: this.envdAccessToken ? { 'X-Access-Token': this.envdAccessToken } : { },
+        headers: {
+          ...defaultHeaders,
```
ValentaTomas: Would it make sense to mass the default header from inside of the ConnectionConfig by default instead?

jakubno:
Let's wait on https://github.com/e2b-dev/infra/pull/847 and https://github.com/e2b-dev/infra/pull/745 and decide what headers we want to set/update based on that

jakubno:
We are sending unique `User-Agent`(#830) for each version and SDK now, so this isn't needed at the end, closing

PR #766, opened 2025-06-07
Pass debug explicitly to Sandbox in tests

mishushakov:
- I don't know why, but I couldn't run Python tests otherwise for some reason


ValentaTomas:
This is really strange as the API key passing from env should also not work if this is not working. Do you think this might be just a problem with the passing inside of the Sandbox class instead?

ValentaTomas:

PR #767, opened 2025-06-09
Use keep-alive in all SDK HTTP requests 

0div:


0div:
@jakubno , is this what you envisioned wrt https://linear.app/e2b/issue/E2B-2185/ensure-we-are-using-keepalives-in-our-sdks ?

packages/js-sdk/src/api/index.ts
```diff
@@ -61,6 +61,12 @@ class ApiClient {
         }),
         ...config.headers,
       },
+      fetch: (request: Request) => {
```
ValentaTomas: It might be possible to just use `keepalive: true` here instead of overriding the whole fetch.


We might also to check adding `follow` here for the redirects.

Issue #746, opened 2025-05-28
api_key not fully working in sandbox_async connect


jimmystlr:
In packages/python-sdk/e2b/sandbox_async/main.py, line 293
response = await SandboxApi.get_info(sandbox_id)
should be changed to:
response = await SandboxApi.get_info(sandbox_id, api_key)
or it will get an exception of AuthenticationException even if the api_key is passed to the "connect" function.

Zhen3r:
same here, connect is currently not woking if apikey is passed (instead of set throuth env)

mishushakov:
This is now fixed

```
e2b==1.5.6
```

PR #772, opened 2025-06-12
Fixes Sandbox.connect not working when trying to pass API key via parameters

mishushakov:
fixes https://github.com/e2b-dev/E2B/issues/746

- Renamed SandboxApi.get_info() > SandboxApi._cls_get_info()
- Properly overloaded Sandbox.get_info()
- Passed connect parameters to SandboxApi._cls_get_info()

### DIFF
--- a/packages/python-sdk/e2b/sandbox_async/main.py
+++ b/packages/python-sdk/e2b/sandbox_async/main.py
@@ -463,6 +463,8 @@ class AsyncSandbox(SandboxSetup, SandboxApi):
         Get sandbox information like sandbox ID, template, metadata, started at/end at date.
         :param sandbox_id: Sandbox ID
         :param api_key: E2B API Key to use for authentication, defaults to `E2B_API_KEY` environment variable
+        :param domain: E2B domain to use for authentication, defaults to `E2B_DOMAIN` environment variable
+        :param debug: Whether to use debug mode, defaults to `E2B_DEBUG` environment variable
         :param request_timeout: Timeout for the request in **seconds**
         :param headers: Custom headers to use for the request
         :param proxy: Proxy to use for the request
--- a/packages/python-sdk/e2b/sandbox_sync/main.py
+++ b/packages/python-sdk/e2b/sandbox_sync/main.py
@@ -442,6 +442,8 @@ class Sandbox(SandboxSetup, SandboxApi):
         Get sandbox information like sandbox ID, template, metadata, started at/end at date.
         :param sandbox_id: Sandbox ID
         :param api_key: E2B API Key to use for authentication, defaults to `E2B_API_KEY` environment variable
+        :param domain: E2B domain to use for authentication, defaults to `E2B_DOMAIN` environment variable
+        :param debug: Whether to use debug mode, defaults to `E2B_DEBUG` environment variable
         :param request_timeout: Timeout for the request in **seconds**
         :param headers: Custom headers to use for the request
         :param proxy: Proxy to use for the request


packages/js-sdk/src/sandbox/index.ts
```diff
@@ -430,24 +448,36 @@ export class Sandbox extends SandboxApi {
    *
    * @returns URL for downloading file.
    */
-  downloadUrl(path: string, opts?: SandboxUrlOpts) { //path: string, useSignature?: boolean, signatureExpirationInSeconds?: number) {
+  downloadUrl(path: string, opts?: SandboxUrlOpts) {
+    //path: string, useSignature?: boolean, signatureExpirationInSeconds?: number) {
```
ValentaTomas: This looks redundant.

ValentaTomas:
@mishushakov The PR name "Fixes Sandbox.connect not working when trying to pass API key via parameters" is not what this PR does?

packages/js-sdk/src/sandbox/index.ts
mishushakov: This is line unchanged, it's just the formatter.

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -436,15 +450,23 @@ export class Sandbox extends SandboxApi {
...
```
sitole: Can we renmove comment here?

packages/python-sdk/tests/async/api_async/test_sbx_info.py
```diff
@@ -5,5 +5,5 @@
 
 @pytest.mark.skip_debug()
 async def test_get_info(async_sandbox: AsyncSandbox):
-    info = await async_sandbox.get_info()
+    info = await AsyncSandbox.get_info(async_sandbox.sandbox_id)
```
sitole, line 8: Not sure if someone is using it, but cannot we break someones code with calling it as static method now?

mishushakov: they can actually use both, so it should be non-breaking.
I have updated the test to match .set_timeout, .kill tests

packages/python-sdk/e2b/sandbox_async/main.py
```diff
@@ -430,6 +436,40 @@ async def set_timeout(  # type: ignore
...
+        headers: Optional[Dict[str, str]] = None,
+        proxy: Optional[ProxyTypes] = None,
+    ) -> SandboxInfo:
+        """
+        Get sandbox information like sandbox ID, template, metadata, started at/end at date.
+        :param sandbox_id: Sandbox ID
+        :param api_key: E2B API Key to use for authentication, defaults to `E2B_API_KEY` environment variable
+        :param request_timeout: Timeout for the request in **seconds**
+        :param proxy: Proxy to use for the request
+        :return: Sandbox info
```
sitole: Can we add missing docs for the params domain and headers here

packages/python-sdk/e2b/sandbox_sync/main.py
```diff
@@ -408,6 +415,40 @@ def set_timeout(  # type: ignore
...
+        headers: Optional[Dict[str, str]] = None,
+        proxy: Optional[ProxyTypes] = None,
+    ) -> SandboxInfo:
+        """
+        Get sandbox information like sandbox ID, template, metadata, started at/end at date.
+        :param sandbox_id: Sandbox ID
+        :param api_key: E2B API Key to use for authentication, defaults to `E2B_API_KEY` environment variable
+        :param request_timeout: Timeout for the request in **seconds**
+        :param proxy: Proxy to use for the request
+        :return: Sandbox info
```
sitole: Headers and domain are missing in docs

sitole (changes requested):

PR #775, opened 2025-06-13
Update default base image to use the code-interpreter one

FanaHOVA:
The [new docs](https://e2b.dev/docs/sandbox-template#1-install-e2b-cli) recommend using this image, but the `template init` was still setting Ubuntu as the default one. It should match the docs suggestion.

### DIFF
--- a/apps/web/src/app/(docs)/docs/sandbox-template/page.mdx
+++ b/apps/web/src/app/(docs)/docs/sandbox-template/page.mdx
@@ -86,7 +86,7 @@ import cowsay
 cowsay.say('Hello from E2B!')
 `)
 
-console.log(execution.stdout)
+console.log(execution.logs)
 ```
 ```python {{ language: 'python' }}
 from e2b_code_interpreter import Sandbox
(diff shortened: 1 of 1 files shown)


mishushakov:
Keep in mind, the Code Interpreter image won't work correctly if you don't supply start command to the `sandbox build` command (as outlined [in docs](https://e2b.dev/docs/sandbox-template#4-build-your-sandbox-template)). People often miss this.

ValentaTomas:
The docs [https://e2b.dev/docs/sandbox-template#3-customize-e2b-dockerfile](https://e2b.dev/docs/sandbox-template#3-customize-e2b-dockerfile) are saying we should use the code-interpreter docker image, because otherwise it was confusing, I think because of that it makes sense that the CLI template init command also creates the template with this base.

mishushakov:
@ValentaTomas didn't quite get you? should the CLI init be based on `ubuntu:22.04` (current) or `e2bdev/code-interpreter:latest` (proposed).

The only confusing bit here is that we generate e2b.Dockerfile but not e2b.toml on template init. Once you actually try building the template it won't do it correctly, as building `e2bdev/code-interpreter:latest` requires a start command.

I am thinking of closing the issue right now and revisiting, as we're working towards improved system.

PR #784, opened 2025-06-19
Option for providing system user during upload/download url creation

sitole:


packages/js-sdk/src/sandbox/index.ts
```diff
@@ -66,6 +66,11 @@ export interface SandboxUrlOpts {
    * Optional parameter to set the expiration time for the signature.
    */
   useSignatureExpiration?: number
+
+  /**
+   * User that will be used to access the file.
+   */
+  user?: string
```
dobrac: use the `Username` type

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -402,12 +407,12 @@ export class Sandbox extends SandboxApi {
     }
 
     const filePath = path ?? ''
-    const fileUrl = this.fileUrl(filePath, defaultUsername)
+    const fileUrl = this.fileUrl(filePath, opts.user ?? defaultUsername)
```
dobrac: nit: can you extract it to a separate var?

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -441,12 +446,12 @@ export class Sandbox extends SandboxApi {
       throw new Error('Signature expiration can be used only when signature is set to true.')
     }
 
-    const fileUrl = this.fileUrl(path, defaultUsername)
+    const fileUrl = this.fileUrl(path, opts.user ?? defaultUsername)
```
dobrac: nit: can you extract it to a separate var?

dobrac (changes requested):
Please add also changeset + tests (if possible)

dobrac (approved):
appreciate the test ❤️

PR #790, opened 2025-06-20
Use TextEncoder for crypto/signed URLs

mishushakov:
- Replaces node:crypto with cross-runtime TextEncoder/TextDecoder API
- The uploadUrl/downloadUrl methods are now async

### DIFF
--- a/packages/js-sdk/src/sandbox/signature.ts
+++ b/packages/js-sdk/src/sandbox/signature.ts
@@ -1,3 +1,5 @@
+import { sha256 } from '../utils'
+
 /**
  * Get the URL signature for the specified path, operation and user.
  *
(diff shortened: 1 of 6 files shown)


packages/js-sdk/src/sandbox/signature.ts
```diff
@@ -12,36 +10,62 @@ import crypto from 'node:crypto'
...
-    envdAccessToken?: string
+  path: string
+  operation: 'read' | 'write'
+  user: string
+  expirationInSeconds?: number
+  envdAccessToken?: string
+}
+
+// Cross-platform crypto implementation
+async function sha256(data: string): Promise<string> {
```
sitole: Let's move this into some helper functions, this will be nice to re-use in the future

packages/js-sdk/src/sandbox/signature.ts
```diff
@@ -12,36 +10,62 @@ import crypto from 'node:crypto'
...
+  // Use Node.js crypto if WebCrypto is not available
```
sitole: nit: do you think we can make the comment proper? We are using Node Crypto as the default option

sitole (changes requested):
- We are breaking functionality by making functions asynchronous, so we should release a proper version, not just a patch.
- We should probably force some linting during pull requests, os we are reviewing only functionality and not formatting (this is probably my bad, I may not have set up ESLint properly)

packages/js-sdk/src/sandbox/signature.ts
mishushakov: I moved the logic a bit other ways (use WebCrypto if available, node:crypto by default)

mishushakov:
> - We are breaking functionality by making functions asynchronous, so we should release a proper version, not just a patch.

done

> - We should probably force some linting during pull requests, os we are reviewing only functionality and not formatting (this is probably my bad, I may not have set up ESLint properly)

I thought of this, but then the problem is your local will always be out-of-date if we lint on the CI/CD. I think lint is fine, the issue is the formatting. You can install prettier extension, or configure your IDE to format on save using prettier.

mishushakov:
Offtopic, but I decided to go with ES Lint + Prettier for now as that's what we're already using here. Can updated this in the future, but no need to now.

PR #798, opened 2025-06-25
Fix broken example code in README

jamesmurdza:
This fixes the broken example code in the README.


0div (approved):

mishushakov:
I think we need to enable CI for PRs

PR #808, opened 2025-07-07
fix: Typing mismatch in filesystem write method

sharat87:
The function signature here is:

```python
    def write(
        self,
        path_or_files: Union[str, List[WriteEntry]],
        data_or_user: Union[str, bytes, IO, Username] = "user",
        user_or_request_timeout: Optional[Union[float, Username]] = None,
        request_timeout_or_none: Optional[float] = None,
    ) -> Union[EntryInfo, List[EntryInfo]]:
```

Notice that `path_or_files` is `str` or a list of `WriteEntry` objects. But further down, if it is a `str`, we're turning it into a `list` of `dict`s, instead of into a `list` of `WriteEntry` objects. We also then proceed to _use_ it like a `dict`. But this means if we pass a `list[WriteEntry]`, as the type hint suggests, we get the following error:

```
  File ".../spike.py", line 49, in write_files
    box.files.write(
          ~~~~~~~~~~~~~~~^
        [
        ^
    ...<2 lines>...
        ],
        ^^
    )
    ^
  File ".../.venv/lib/python3.13/site-packages/e2b/sandbox_sync/filesystem/filesystem.py", line 213, in write
    file_path, file_data = file["path"], file["data"]
                           ~~~~^^^^^^^^
TypeError: 'WriteEntry' object is not subscriptable
```

This PR fixes this, by instead only creating/assuming `WriteEntry` objects in the `list`, as the type hint suggests.

### DIFF
--- a/packages/python-sdk/e2b/sandbox_sync/filesystem/filesystem.py
+++ b/packages/python-sdk/e2b/sandbox_sync/filesystem/filesystem.py
@@ -193,7 +193,7 @@ class Filesystem:
                 )
             path, write_files, user, request_timeout = (
                 path_or_files,
-                [{"path": path_or_files, "data": data_or_user}],
+                [WriteEntry(path=path_or_files, data=data_or_user)],
                 user_or_request_timeout or "user",
                 request_timeout_or_none,
             )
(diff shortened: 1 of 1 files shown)


sharat87:
Looks like this is no longer an issue. Thanks folks, sorry for not being able to follow up. Closing now.

PR #811, opened 2025-07-07
Do not use user config team id for template build command 

0div:
# Description
The team ID is derived from the provided access token for this specific command, this PR lets you still explicitly specify the team id via opts but does not take it from the user config to avoid unwanted team does not exist errors when specifying access tokens for different clusters.

# Test
```sh
$ [E2B/packages/cli] pnpm build

# test while being logged in
$ [E2B/templates/base] e2b auth login # in prod
$ E2B/templates/base] E2B_DOMAIN=<your_dev_domain> E2B_ACCESS_TOKEN=<your_access_token> ../../packages/cli/dist/index.js template build

# test while being logged out 
$ [E2B/templates/base] e2b auth logout
$ E2B/templates/base] E2B_DOMAIN=<your_dev_domain> E2B_ACCESS_TOKEN=<your_access_token> ../../packages/cli/dist/index.js template build
```

### DIFF
--- a/packages/cli/src/commands/template/build.ts
+++ b/packages/cli/src/commands/template/build.ts
@@ -262,9 +262,6 @@ export const buildCommand = new commander.Command('build')
         }
 
         const userConfig = getUserConfig()
-        if (userConfig) {
-          teamID = teamID || userConfig.teamId
-        }
 
         if (config && templateID && config.template_id !== templateID) {
           // error: you can't specify different ID than the one in config


mishushakov:
1. Does it still work with no env vars
2. Does it still work when trying to override team id using --team?

mishushakov:
how does it know what team to upload the template to if the team id is not provided?

0div:
> how does it know what team to upload the template to if the team id is not provided?

it derives it from the access token

0div:
> But if it's a different one in the template.toml? Will it still be using my default team?

Down the backend call stack of the handler for the request made [here](https://github.com/e2b-dev/E2B/blob/main/packages/cli/src/commands/template/build.ts#L301-L322) in the CLI client code, this is what determines the team ID, in other words it seems like no team ID is used. 

[/infra/packages/api/internal/handlers/auth.go](https://github.com/e2b-dev/infra/blob/1016445b98f7931680436c7b03188c172c3d8a3b/packages/api/internal/handlers/auth.go#L4)
```go
func (a *APIStore) GetUserAndTeams(c *gin.Context) (*uuid.UUID, []queries.GetTeamsWithUsersTeamsWithTierRow, error) {
	userID := a.GetUserID(c)
	ctx := c.Request.Context()

	teams, err := a.sqlcDB.GetTeamsWithUsersTeamsWithTier(ctx, userID)
	if err != nil {
		return nil, nil, fmt.Errorf("error when getting default team: %w", err)
	}

	return &userID, teams, err
}
```

mishushakov:
Yeah but the problem here it will always use the default team id or the one supplied as a parameter but will ignore e2b.toml?

mishushakov (approved):

PR #819, opened 2025-07-15
Add image URI mask option for template build to allow custom image repository

sitole:
This will allow customers who are not using our Docker repository to fall back to their own during template build.

### DIFF
--- a/packages/cli/src/commands/template/build.ts
+++ b/packages/cli/src/commands/template/build.ts
@@ -29,6 +29,8 @@ import { buildWithProxy } from './buildWithProxy'
 
 const templateCheckInterval = 500 // 0.5 sec
 
+export const imageUriMask = process.env.E2B_IMAGE_URI_MASK
+
 async function getTemplateBuildLogs({
                                       templateID,
                                       buildID,
@@ -173,10 +175,6 @@ export const buildCommand = new commander.Command('build')
     'specify additional build arguments for the build command. The format should be <varname>=<value>.',
   )
   .option('--no-cache', 'skip cache when building the template.')
-  .option(
-      '--image-url-mask <mask>',
-      'specify image URL mask if you need to push to different from default Docker repository. Mask can contain {buildID} and {templateID} placeholders that will be resolved before calling.',
-  )
   .alias('bd')
   .action(
     async (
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/template/build.ts
```diff
@@ -173,6 +173,10 @@ export const buildCommand = new commander.Command('build')
     'specify additional build arguments for the build command. The format should be <varname>=<value>.',
   )
   .option('--no-cache', 'skip cache when building the template.')
+  .option(
+      '--image-url-mask <mask>',
```
sitole: Maybe we should call it just `--image-mask`

mishushakov: is --image-url-mask required when using custom domain?

packages/cli/src/commands/template/build.ts
```diff
@@ -340,20 +345,23 @@ export const buildCommand = new commander.Command('build')
...
-              stdio: 'inherit',
-              cwd: root,
-            },
-          )
-        } catch (err: any) {
-          console.error(
-            'Docker login failed. Please try to log in with `e2b auth login` and try again.',
-          )
-          process.exit(1)
+        if (opts.imageUrlMask == undefined) {
```
mishushakov: why skip auth when imageUrlMask is provided?

sitole: Because if a mask is provided,  you will never use this auth; actually, it can break stuff. Imagine that I am authorized to aws container registry, then I will run cli with the mask attribute, and here Docker will authorize to our default registry. Then docker push will fail, because iam authorized to a different registry.

packages/cli/src/commands/template/build.ts
sitole: Technically what user is providing is called image URI so iam to simplifiy to one of thease `--image-uri-mask`, `--image-mask`, `--docker-image-uri-mask`

packages/cli/src/commands/template/build.ts
```diff
@@ -616,3 +625,11 @@ async function triggerBuild(templateID: string, buildID: string) {
 
   return
 }
+
+function dockerImageUrl(templateID: string, buildID: string, defaultDomain: string, imageUrlMask?: string): string {
+  if (imageUrlMask == undefined) {
```
mishushakov, line 632: why is this condition dependent on imageUrlMask?

PR #829, opened 2025-07-25
Fix generating files with docker

jakubno:
# Description

Fixes an issue in generating files in Docker. There has been an incompatibility of `buf` (version `29.5`) and `protoc-gen-es` (version `2.2.2`).

I updated `protoc-gen-es@` to `2.6.2`

Also refactored the code a little so it's easier to read
Added a CI pipeline job to check all files are properly generated


Makefile
```diff
@@ -1,28 +1,16 @@
-update-api-spec:
-	@echo "Updating API spec"
-	@./scripts/update-api-spec.sh
-	@echo "Done"
-
```
sitole, line 5: How is the spec updated now?

jakubno: it didn't work and it isn't used anyway, so I removed it to prevent issues

PR #830, opened 2025-07-28
Update `User-Agent` header in SDKs based on the SDK version

jakubno:
# Description

The bug where extra fields in envd response caused an exception was fixed in #738. As we're also ignoring new fields in envd (implemented in https://github.com/e2b-dev/infra/pull/847). We now have to update the `user-agent` header.

### DIFF
--- a/.changeset/plenty-foxes-press.md
+++ b/.changeset/plenty-foxes-press.md
@@ -1,4 +1,5 @@
 ---
+'e2b': patch
 '@e2b/python-sdk': patch
 ---
 
--- a/.changeset/plenty-foxes-press.md
+++ b/.changeset/plenty-foxes-press.md
@@ -1,0 +1,5 @@
+---
+'@e2b/python-sdk': patch
+---
+
+Update connect header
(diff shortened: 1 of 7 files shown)


packages/python-sdk/e2b_connect/client.py
```diff
@@ -161,7 +161,7 @@ def __init__(
         self._codec = JSONCodec if json else ProtobufCodec
         self._response_type = response_type
         self._compressor = compressor
-        self._headers = {**{"user-agent": "connect-python"}, **headers}
+        self._headers = {**{"user-agent": "connect-python-v2"}, **headers}
```
dobrac: don't we want to set the proper python version there right away?

dobrac (changes requested):

.changeset/plenty-foxes-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'@e2b/python-sdk': patch
```
dobrac, line 3: release also the JS sdk?

packages/js-sdk/src/api/metadata.ts
```diff
@@ -2,6 +2,8 @@ import platform from 'platform'
 
 import { version } from '../../package.json'
 
+export { version } from '../../package.json'
```
dobrac: why not use just `export { version }`

dobrac (approved):

PR #845, opened 2025-08-04
Secured sandbox auto signature

sitole:
Some users report that JS/Python types lack a `useSignature` attribute, which only works when `secure=true` is set for the sandbox. This change removes the `useSignatrue` parameter and makes all upload/download URL requests signed if `secure` was provided, making sure the function will work for all possible cases.

### DIFF
--- a/packages/js-sdk/src/sandbox/index.ts
+++ b/packages/js-sdk/src/sandbox/index.ts
@@ -65,7 +65,7 @@ export interface SandboxOpts extends ConnectionOpts {
 export interface SandboxUrlOpts {
   /**
    * Use signature expiration for the URL.
-   * Optional parameter to set the expiration time for the signature.
+   * Optional parameter to set the expiration time for the signature in seconds.
    */
   useSignatureExpiration?: number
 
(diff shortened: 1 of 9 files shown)


sitole:
This is also a change needed for safe migration to allow `secure=true` as default when creating a sandbox.

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -63,14 +63,6 @@ export interface SandboxOpts extends ConnectionOpts {
...
-   * Use signature for the URL.
-   * This needs to be used in case of using secured envd in sandbox.
-   *
-   * @default false
-   */
-  useSignature?: true
-
   /**
    * Use signature expiration for the URL.
    * Optional parameter to set the expiration time for the signature.
```
ben-fornefeld:
```suggestion
   * Optional parameter to set the expiration time for the signature in seconds.
```

ben-fornefeld (approved):
small usability enhancement, otherwise LGTM.

PR #846, opened 2025-08-04
Allow more protobuf versions

ValentaTomas:
- Allow protobuf `4.21.0`+ (we previously allowed this, and some users in beta require this)
- Allow protobuf `6.0.0`+ (some other users need to use the SDK with the latest versions)

The SDK should work with both of these, but we may want to smoke test it somehow.


jakubno (changes requested):
It doesn't work with `3.20.0`, also it's long time outdated https://protobuf.dev/support/version-support/#python
```
ImportError while loading conftest '/Users/i/e2b/E2B/packages/python-sdk/tests/conftest.py'.
tests/conftest.py:9: in <module>
    from e2b import (
e2b/__init__.py:55: in <module>
    from .sandbox.filesystem.watch_handle import (
e2b/sandbox/filesystem/watch_handle.py:4: in <module>
    from e2b.envd.filesystem.filesystem_pb2 import EventType
e2b/envd/filesystem/filesystem_pb2.py:9: in <module>
    from google.protobuf import runtime_version as _runtime_version
E   ImportError: cannot import name 'runtime_version' from 'google.protobuf' (/Users/i/Library/Caches/pypoetry/virtualenvs/e2b-smSWSkw9-py3.11/lib/python3.11/site-packages/google/protobuf/__init__.py)
```

PR #849, opened 2025-08-06
Simplify connection config handling in Python SDK

jakubno:
Several improvements and refactors to the Python SDK's sandbox connection and configuration logic. 

The main focus is on making header management more flexible, improving API parameter handling, and simplifying class structures for better maintainability and usability. 

Also unifies doc strings

### DIFF
--- a/packages/python-sdk/e2b/connection_config.py
+++ b/packages/python-sdk/e2b/connection_config.py
@@ -150,6 +150,9 @@ class ConnectionConfig:
 
     @property
     def sandbox_headers(self):
+        """
+        # We need this separate as we use the same header for E2B access token to API and envd access token to sandbox.
+        """
         return {
             **self.headers,
             **self.__extra_sandbox_headers,
--- a/packages/python-sdk/e2b/connection_config.py
+++ b/packages/python-sdk/e2b/connection_config.py
@@ -1,7 +1,9 @@
 import os
 
 from typing import Literal, Optional, Dict
+from dataclasses import dataclass
 from httpx._types import ProxyTypes
+from typing_extensions import Unpack
 
 from e2b.api.metadata import package_version
 
@@ -9,6 +11,33 @@ REQUEST_TIMEOUT: float = 60.0  # 60 seconds
 
 KEEPALIVE_PING_INTERVAL_SEC = 50  # 50 seconds
 KEEPALIVE_PING_HEADER = "Keepalive-Ping-Interval"
+
+
+@dataclass
+class ApiParams:
+    """
+    Parameters for a request.
+
+    In the case of a sandbox, it applies to all **requests made to the returned sandbox**.
+    """
+
+    request_timeout: Optional[float]
+    """Timeout for the request in **seconds**, defaults to 60 seconds."""
+    
+    headers: Dict[str, str]
+    """Additional headers to send with the request."""
+    
+    api_key: Optional[str]
+    """E2B API Key to use for authentication, defaults to `E2B_API_KEY` environment variable."""
+    
+    domain: str
+    """E2B domain to use for authentication, defaults to `E2B_DOMAIN` environment variable."""
+    
+    debug: bool
+    """Whether to use debug mode, defaults to `E2B_DEBUG` environment variable."""
+    
+    proxy: Optional[ProxyTypes]
+    """Proxy to use for the request. In case of a sandbox it applies to all **requests made to the returned sandbox**."""
 
 
 class ConnectionConfig:
@@ -40,6 +69,7 @@ class ConnectionConfig:
         access_token: Optional[str] = None,
         request_timeout: Optional[float] = None,
         headers: Optional[Dict[str, str]] = None,
+        extra_sandbox_headers: Optional[Dict[str, str]] = None,
         proxy: Optional[ProxyTypes] = None,
     ):
         self.domain = domain or ConnectionConfig._domain()
@@ -48,6 +78,7 @@ class ConnectionConfig:
         self.access_token = access_token or ConnectionConfig._access_token()
         self.headers = headers or {}
         self.headers["User-Agent"] = f"e2b-python-sdk/{package_version}"
+        self.__extra_sandbox_headers = extra_sandbox_headers or {}
 
         self.proxy = proxy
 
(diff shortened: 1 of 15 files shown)


packages/python-sdk/e2b/sandbox/sandbox_api.py
```diff
@@ -28,7 +28,7 @@ class SandboxInfo:
     """Sandbox expiration date."""
     envd_version: Optional[str]
     """Envd version."""
-    _envd_access_token: Optional[str]
+    envd_access_token: Optional[str]
     """Envd access token."""
```
ValentaTomas: I really don't think we want to expose these now/yet.

I want to unify the sandboxInfo and listed sandbox types as in (https://github.com/e2b-dev/E2B/blob/no-beta/packages/python-sdk/e2b/sandbox/sandbox_api.py#L41) and the envd_version/envd_access_token can be missing on these, so for now I would hide the fields.

packages/python-sdk/e2b/connection_config.py
```diff
@@ -11,6 +12,32 @@
 KEEPALIVE_PING_HEADER = "Keepalive-Ping-Interval"
 
 
+class ApiParams(TypedDict):
```
ValentaTomas: As mentioned in private convo—for internal passing the `ApiParams` seems worth it, but we need to see if the `Unpack` behaves properly when used for autocomplete/docstrigns everywhere.

packages/python-sdk/e2b/sandbox/main.py
```diff
@@ -9,7 +8,7 @@
 from httpx import Limits
 
 
-class SandboxSetup(ABC):
+class SandboxSetup:
```
ValentaTomas: Really considering merging this with `SandboxApiBase`.

packages/python-sdk/e2b/sandbox/sandbox_api.py
ValentaTomas: Related also to changes in JS SDK (https://github.com/e2b-dev/E2B/blob/670b3b8b1ed491079ce42501765b6d03f32509a8/packages/js-sdk/src/sandbox/sandboxApi.ts#L221), but we can add that separately.

ValentaTomas:
One detail:
- https://github.com/e2b-dev/E2B/blob/1395b6e9e1a3d9ed7fea414b5f78b98ce56d3e0c/packages/python-sdk/e2b/sandbox_sync/sandbox_api.py#L249
- https://github.com/e2b-dev/E2B/blob/1395b6e9e1a3d9ed7fea414b5f78b98ce56d3e0c/packages/python-sdk/e2b/sandbox_async/sandbox_api.py#L264

Here we can use `SandboxApiBase._limits` as everywhere else too.

packages/python-sdk/e2b/sandbox/sandbox_api.py
jakubno: I think we want to expose envd version as it is used to check if some feature is available and it's used e.g. in dashboard. I will make `envd_access_token` private though

packages/python-sdk/e2b/connection_config.py
jakubno: Works well with both cursor and jet brains
image
image

packages/python-sdk/e2b/connection_config.py
```diff
@@ -40,6 +67,7 @@ def __init__(
         access_token: Optional[str] = None,
         request_timeout: Optional[float] = None,
         headers: Optional[Dict[str, str]] = None,
+        extra_sandbox_headers: Optional[Dict[str, str]] = None,
```
mishushakov, line 70: I find it looks weird that you have both headers and extra_sandbox_headers
what about?

```py
default_headers = {'User-Agent': 'xxx'}

def __init__(
    headers: Dict[str, str] = {}
):
    self.headers = {**default_headers, **headers}
```

packages/python-sdk/e2b/sandbox/sandbox_api.py
ValentaTomas: Do we always send the envd version—even if the sandbox is paused and we retrieve the data from BD?
If so, we can make it non-optional and public.

ValentaTomas:
> @jakubno one question - do we still need this? maybe there's a way to better handle this
> 
> https://github.com/e2b-dev/E2B/blob/main/packages/python-sdk/e2b/sandbox_async/main.py#L507-L512

@mishushakov I think this is already removed in this PR as it was both duplicated and there was a bug.

packages/python-sdk/e2b/sandbox/sandbox_api.py
jakubno: Yes, it wasn't present in list (not even for running) before, we added it quite recently

packages/python-sdk/e2b/sandbox/main.py
jakubno: Removed

jakubno:
> @jakubno one question - do we still need this? maybe there's a way to better handle this
> 
> https://github.com/e2b-dev/E2B/blob/main/packages/python-sdk/e2b/sandbox_async/main.py#L507-L512

It's basically removed, right?

jakubno:
> Can we add docstring for options also?

They are there: https://github.com/e2b-dev/e2b/blob/8ea3448e18edd64c01d7e723cea0ce561a8db633/packages/python-sdk/e2b/connection_config.py#L15

packages/python-sdk/e2b/sandbox/sandbox_api.py
ValentaTomas: I'm okay with keeping the *version optional* right now though and changing the backend in a separate PR later (to always return it as we should always have it).

packages/python-sdk/e2b/connection_config.py
ValentaTomas: @mishushakov Are you asking why we need two sets of headers?

packages/python-sdk/e2b/sandbox/sandbox_api.py
jakubno: Yea, let's move to separate PR, it's marked as optional in backed (it should be enough ad it as required)

packages/python-sdk/e2b/sandbox_async/main.py
```diff
@@ -203,54 +164,45 @@ async def create(
...
                 timeout=timeout or cls.default_sandbox_timeout,
                 metadata=metadata,
-                domain=domain,
                 debug=debug,
-                request_timeout=request_timeout,
                 env_vars=envs,
                 secure=secure,
-                proxy=proxy,
                 allow_internet_access=allow_internet_access,
+                **opts,
```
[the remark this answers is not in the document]
jakubno: Good boy

packages/python-sdk/e2b/connection_config.py
mishushakov: @ValentaTomas yes

ValentaTomas: We don't want to be sending the "api" headers (like the supabase token) to the sandbox endpoints and we also don't want to send the "sandbox" headers (like the envd access token) to the API, so to prevent mixing them on the requests made from the methods on the Sandbox instance we need two separate sets.

It is possible that it can be handled more elegantly internally, but we are mostly making sure it behaves sanely to the outside now.

PR #854, opened 2025-08-08
SDK v2 - Add beta submodule and update Sandbox `list`

jakubno:
Introduce beta submodule with beta features - pause and resume
Update Sandbox list to also return paused sandboxes

### DIFF
--- a/apps/web/src/app/(docs)/docs/api-key/page.mdx
+++ b/apps/web/src/app/(docs)/docs/api-key/page.mdx
@@ -16,7 +16,7 @@ const sandbox = await Sandbox.create({ apiKey: 'YOUR_API_KEY' })
 ```python {{ language: 'python' }}
 from e2b_code_interpreter import Sandbox
 
-sbx = Sandbox.create(api_key="YOUR_API_KEY")
+sbx = Sandbox(api_key="YOUR_API_KEY")
 ```
 </CodeGroup>
 
--- a/apps/web/src/app/(docs)/docs/code-interpreting/analyze-data-with-ai/page.mdx
+++ b/apps/web/src/app/(docs)/docs/code-interpreting/analyze-data-with-ai/page.mdx
@@ -82,7 +82,7 @@ load_dotenv()
 from e2b_code_interpreter import Sandbox
 
 # Create sandbox
-sbx = Sandbox.create()
+sbx = Sandbox()
 
 # Upload the dataset to the sandbox
 dataset_path_in_sandbox = ""
@@ -453,7 +453,7 @@ from e2b_code_interpreter import Sandbox
 from anthropic import Anthropic
 
 # Create sandbox
-sbx = Sandbox.create()
+sbx = Sandbox()
 
 # Upload the dataset to the sandbox
 with open("../dataset.csv", "rb") as f:
--- a/apps/web/src/app/(docs)/docs/code-interpreting/create-charts-visualizations/interactive-charts/page.mdx
+++ b/apps/web/src/app/(docs)/docs/code-interpreting/create-charts-visualizations/interactive-charts/page.mdx
@@ -68,7 +68,7 @@ plt.tight_layout()
 plt.show()
 """
 
-sandbox = Sandbox.create()
+sandbox = Sandbox()
 execution = sandbox.run_code(code)
 chart = execution.results[0].chart
 
--- a/apps/web/src/app/(docs)/docs/code-interpreting/create-charts-visualizations/static-charts/page.mdx
+++ b/apps/web/src/app/(docs)/docs/code-interpreting/create-charts-visualizations/static-charts/page.mdx
@@ -47,7 +47,7 @@ plt.ylabel('some numbers')
 plt.show()
 """
 
-sandbox = Sandbox.create()
+sandbox = Sandbox()
 
 # Run the code inside the sandbox
 execution = sandbox.run_code(code_to_run)
--- a/apps/web/src/app/(docs)/docs/code-interpreting/streaming/page.mdx
+++ b/apps/web/src/app/(docs)/docs/code-interpreting/streaming/page.mdx
@@ -40,7 +40,7 @@ code_to_run = """
   print("This goes last")
 """
 
-sandbox = Sandbox.create()
+sandbox = Sandbox()
 sandbox.run_code(
   code_to_run,
   # Use `on_error` to handle runtime code errors
@@ -131,7 +131,7 @@ plt.title('Values by Category')
 plt.show()
 """
 
-sandbox = Sandbox.create()
+sandbox = Sandbox()
 sandbox.run_code(
   code_to_run,
   on_result=lambda result: print('result:', result), # $HighlightLine
--- a/apps/web/src/app/(docs)/docs/code-interpreting/supported-languages/bash/page.mdx
+++ b/apps/web/src/app/(docs)/docs/code-interpreting/supported-languages/bash/page.mdx
@@ -12,7 +12,7 @@ console.log(execution)
 ```python
 from e2b_code_interpreter import Sandbox
 
-sbx = Sandbox.create()
+sbx = Sandbox()
 execution = sbx.run_code("echo 'Hello, world!'", language="bash")
 print(execution)
 ```
--- a/apps/web/src/app/(docs)/docs/code-interpreting/supported-languages/java/page.mdx
+++ b/apps/web/src/app/(docs)/docs/code-interpreting/supported-languages/java/page.mdx
@@ -12,7 +12,7 @@ console.log(execution)
 ```python
 from e2b_code_interpreter import Sandbox
 
-sbx = Sandbox.create()
+sbx = Sandbox()
 execution = sbx.run_code('System.out.println("Hello, world!");', language="java")
 print(execution)
 ```
--- a/apps/web/src/app/(docs)/docs/code-interpreting/supported-languages/javascript/page.mdx
+++ b/apps/web/src/app/(docs)/docs/code-interpreting/supported-languages/javascript/page.mdx
@@ -46,7 +46,7 @@ console.log(execution);
 from e2b_code_interpreter import Sandbox
 
 # Create a new sandbox
-sbx = Sandbox.create()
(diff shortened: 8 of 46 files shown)


packages/python-sdk/e2b/sandbox_async/main.py
```diff
@@ -45,7 +64,210 @@ class AsyncSandboxOpts(TypedDict):
...
+        query: Optional[SandboxQueryBeta] = None,
+        limit: Optional[int] = None,
+        next_token: Optional[str] = None,
+        **opts: Unpack[ApiParams],
+    ) -> AsyncSandboxPaginator:
+        """
+        List all running sandboxes.
+
+        :param query: Filter the list of sandboxes by metadata or state, e.g. `SandboxListQuery(metadata={"key": "value"})` or `SandboxListQuery(state=[SandboxState.RUNNING])`
+        :param limit: Maximum number of sandboxes to return
```
ValentaTomas: We should clarify that this is a maximum per page.

packages/python-sdk/e2b/sandbox_async/sandbox_beta.py
ValentaTomas, line 1: The new list won't be beta.

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -209,6 +209,69 @@ export class Sandbox extends SandboxApi {
...
+       * @param sandboxId sandbox ID.
+       * @param opts connection options.
+       *
+       * @returns sandbox ID that can be used to resume the sandbox.
+       */
+      pause: Sandbox.pauseSandbox.bind(Sandbox),
+    }
+  }
+
+  get beta() {
```
ValentaTomas: I would:
1. Try returning a class/instance so the methods are not properies (also for th static)
2. Add docstring to the beta property itself, so it is clear why you would call this.

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -209,6 +210,74 @@ export class Sandbox extends SandboxApi {
...
+        })
+      },
+      /**
+       * Pause a sandbox by its ID.
+       *
+       * @param opts connection options.
+       *
+       * @returns sandbox ID that can be used to resume the sandbox.
+       */
+      pause: async (
```
ValentaTomas: I think this is just `ConnectionOpts` instead of Omitted SandboxOpts.
Maybe we can simplify this in more places where we essentially just call API on instance.

packages/js-sdk/tests/setup.ts
```diff
@@ -4,13 +4,25 @@ import { template } from './template'
 
 interface SandboxFixture {
   sandbox: Sandbox
+  template: string
+  sandboxType: string
```
ValentaTomas: Calling this "type" is a little confusing.

packages/python-sdk/e2b/sandbox_async/sandbox_api.py
```diff
@@ -11,70 +10,51 @@
...
+        next_token: Optional[str] = None,
         **opts: Unpack[ApiParams],
-    ) -> List[SandboxInfo]:
+    ) -> AsyncSandboxPaginator:
         """
         List all running sandboxes.
 
-        :param query: Filter the list of sandboxes, e.g. by metadata `SandboxQuery(metadata={"key": "value"})`, if there are multiple filters, they are combined with AND.
+        :param query: Filter the list of sandboxes by metadata or state, e.g. `SandboxListQuery(metadata={"key": "value"})` or `SandboxListQuery(state=[SandboxState.RUNNING])`
+        :param limit: Maximum number of sandboxes to return
```
ValentaTomas: This limit is per page—would add that detail to the docstring.

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -295,8 +365,14 @@ export class Sandbox extends SandboxApi {
     sandboxId: string,
     opts?: Omit<SandboxOpts, 'metadata' | 'envs' | 'timeoutMs'>
   ): Promise<InstanceType<S>> {
-    const config = new ConnectionConfig(opts)
-    const info = await this.getInfo(sandboxId, opts)
+    const info = await this.getFullInfo(sandboxId, opts)
+
+    const config = new ConnectionConfig({
+      ...opts,
+      // We don't want to pass headers to the connection config as they would then be inherited by all requests,
```
ValentaTomas: We need to decide and unify if passing headers to `create`, `connect`, and `resume` should store the headers for all requests made on the sandbox instance from then on.

We have the sandbox headers separated, so that does not play a role, so we can disregard that.

(For passing proxy in the Python that makes sense to me to then always use the proxy for requests from that sandbox instance.)

I might be for not saving the "API" headers and just exposing the `headers` on each request (even instance) to the API here.

packages/js-sdk/src/sandbox/sandboxApi.ts
```diff
@@ -39,11 +44,36 @@ export interface SandboxCreateOpts extends SandboxApiOpts {
...
+    metadata?: Record<string, string>
+    /**
+     * Filter the list of sandboxes by state.
+     * @default ['running', 'paused']
+     */
+    state?: Array<SandboxState>
+  }
+
+  /**
+   * Number of sandboxes to return.
```
ValentaTomas: This was also just number of returned in each page, right?

packages/python-sdk/e2b/sandbox_sync/sandbox_api.py
```diff
@@ -11,67 +10,51 @@
...
+        next_token: Optional[str] = None,
         **opts: Unpack[ApiParams],
-    ) -> List[SandboxInfo]:
+    ) -> SandboxPaginator:
         """
         List all running sandboxes.
 
-        :param query: Filter the list of sandboxes, e.g. by metadata `SandboxQuery(metadata={"key": "value"})`, if there are multiple filters, they are combined with AND.
+        :param query: Filter the list of sandboxes by metadata or state, e.g. `SandboxListQuery(metadata={"key": "value"})` or `SandboxListQuery(state=[SandboxState.RUNNING])`
+        :param limit: Maximum number of sandboxes to return
```
ValentaTomas: Is this limit per page specification

packages/python-sdk/e2b/sandbox_sync/main.py
```diff
@@ -37,7 +37,127 @@ def pool(self):
...
+        self._sandbox = sandbox
+
+    @overload
+    def pause(
+        self,
+        **opts: Unpack[ApiParams],
+    ) -> str:
+        """
+        Pause the sandbox.
+
```
ValentaTomas: Can the docstrings of params be somehow unpacked to the method? I think you can manipulate that with decorators, just not sure if that would work for showing them in the IDEs.

ValentaTomas:
Also, one note—if we are squash merging, is there a point in force pushing? When I'm then pulling the updates it is a little tedious as now it tries to merge and I need to cancel and delete.

ValentaTomas:
Screenshot 2025-08-10 at 19 29 50

Also some errors around possible `Unset`/`Error` that might also be a VSCode more primitive linter fault.

packages/python-sdk/e2b/sandbox_sync/main.py
jakubno: Not really, but you should get validation, recommendations in IDEs, it works basically the same as in TS, doesn't it?

packages/js-sdk/src/sandbox/index.ts
jakubno: Based on our discussion, I am passing the opts to the Sandbox isntance

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -307,6 +354,80 @@ export class Sandbox extends SandboxApi {
     }) as InstanceType<S>
   }
 
+  /**
+   * [BETA] This feature is in beta and may change in the future.
```
ValentaTomas: I'm thinking if replacing the `connect` as it is might make more sense—otherwise we will need to do another breaking update when we replace it.

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -307,6 +354,80 @@ export class Sandbox extends SandboxApi {
...
+   * ```ts
+   * const sandbox = await Sandbox.create()
+   * const sandboxId = sandbox.sandboxId
+   * await Sandbox.betaPause()
+   *
+   * // Connect to the same sandbox.
+   * const sameSandbox = await Sandbox.betaConnect(sandboxId)
+   * ```
+   */
+  static async betaConnect<S extends typeof Sandbox>(
```
ValentaTomas: Still considering the `.beta<Method>` approach. @jakubno Any alternatives that we have though?

jakubno: The new beta class has issues with inheritance. You have to create something like `BetaCodeInterpreter`, and since `BetaSandbox` isn't a mixin, it becomes a bit tricky with `BetaCodeInterpreter(CodeInterpreter, BetaSandbox)`.

Plus, it's not clear which methods are beta and which aren't.

--- 

I also tried out a Factory/Manager class and an instance class, but inheritance is complicated as well, because you end up having to create two classes, one for the manager and one for the instance class.

Manager has only class methods:
Something like `SandboxManager.create()` and it creates instance of class `Sandbox`, it has only instance methods.

packages/js-sdk/src/sandbox/index.ts
jakubno: Make sense, I will replace it

packages/js-sdk/src/sandbox/index.ts
ValentaTomas: Ok, I cannot come up with a working better solution too here now—the `.beta<Method>` is actually good that it both "clusters" the beta methods on autocomplete, is simple, and really identified the methods as beta.

ValentaTomas: Btw, there is https://typedoc.org/documents/Tags._beta.html, will add.

PR #869, opened 2025-08-22
Fix API key check in JS SDK & remove deprecated fields

jakubno:
Fix the check and remove deprecated fields


packages/js-sdk/src/api/index.ts
```diff
@@ -41,7 +41,7 @@ class ApiClient {
     } = { requireAccessToken: false, requireApiKey: true }
   ) {
     // FIXME: This statement makes no sense
```
jakubno:
```suggestion
```

PR #871, opened 2025-08-25
Add Templates SDK

mishushakov:
Merge progress

- [ ] JavaScript SDK
- [ ] Python SDK

...to be continued

### DIFF
--- a/packages/js-sdk/src/api/index.ts
+++ b/packages/js-sdk/src/api/index.ts
@@ -13,6 +13,16 @@ export function handleApiError(
     return
   }
 
+  if (response.response.status === 401) {
+    const message = 'Unauthorized, please check your credentials.'
+    const content = response.error?.message ?? response.error
+
+    if (content) {
+      return new AuthenticationError(`${message} - ${content}`)
+    }
+    return new AuthenticationError(message)
+  }
+
   if (response.response.status === 429) {
     const message = 'Rate limit exceeded, please try again later'
     const content = response.error?.message ?? response.error
--- a/packages/js-sdk/src/errors.ts
+++ b/packages/js-sdk/src/errors.ts
@@ -94,3 +94,17 @@ export class RateLimitError extends SandboxError {
     this.name = 'RateLimitError'
   }
 }
+
+export class BuildError extends Error {
+  constructor(message: string) {
+    super(message)
+    this.name = 'BuildError'
+  }
+}
+
+export class FileUploadError extends BuildError {
+  constructor(message: string) {
+    super(message)
+    this.name = 'FileUploadError'
+  }
+}
--- a/packages/js-sdk/src/index.ts
+++ b/packages/js-sdk/src/index.ts
@@ -58,3 +58,5 @@ export type {
 export { Sandbox }
 import { Sandbox } from './sandbox'
 export default Sandbox
+
+export { Template } from './template'
--- a/packages/js-sdk/src/template/buildApi.ts
+++ b/packages/js-sdk/src/template/buildApi.ts
@@ -1,0 +1,167 @@
+import { ApiClient, paths, handleApiError } from '../api'
+import { BuildError, FileUploadError } from '../errors'
+import { tarFileStreamUpload } from './utils'
+
+type RequestBuildInput = {
+  alias: string
+  cpuCount: number
+  memoryMB: number
+}
+
+type GetFileUploadLinkInput = {
+  templateID: string
+  filesHash: string
+}
+
+type TriggerBuildInput = {
+  templateID: string
+  buildID: string
+  template: TriggerBuildTemplate
+}
+
+type GetBuildStatusInput = {
+  templateID: string
+  buildID: string
+  logsOffset: number
+}
+
+export type GetBuildStatusResponse =
+  paths['/templates/{templateID}/builds/{buildID}/status']['get']['responses']['200']['content']['application/json']
+
+export type TriggerBuildTemplate =
+  paths['/v2/templates/{templateID}/builds/{buildID}']['post']['requestBody']['content']['application/json']
+
+export async function requestBuild(
+  client: ApiClient,
+  { alias, cpuCount, memoryMB }: RequestBuildInput
+) {
+  const requestBuildRes = await client.api.POST('/v2/templates', {
+    body: {
+      alias,
+      cpuCount,
+      memoryMB,
+    },
+  })
+
+  const error = handleApiError(requestBuildRes)
+  if (error) {
+    throw error
+  }
+
+  if (!requestBuildRes.data) {
+    throw new BuildError('Failed to request build')
+  }
+
+  return requestBuildRes.data
+}
+
+export async function getFileUploadLink(
+  client: ApiClient,
+  { templateID, filesHash }: GetFileUploadLinkInput
+) {
+  const fileUploadLinkRes = await client.api.GET(
+    '/templates/{templateID}/files/{hash}',
+    {
+      params: {
+        path: {
+          templateID,
+          hash: filesHash,
+        },
+      },
+    }
+  )
+
+  const error = handleApiError(fileUploadLinkRes)
+  if (error) {
+    throw error
+  }
+
+  if (!fileUploadLinkRes.data) {
+    throw new BuildError('Failed to get file upload link')
+  }
+
+  return fileUploadLinkRes.data
+}
+
+export async function uploadFile(options: {
+  fileName: string
+  fileContextPath: string
+  url: string
+}) {
+  const { fileName, url, fileContextPath } = options
+  const { contentLength, uploadStream } = await tarFileStreamUpload(
+    fileName,
+    fileContextPath
+  )
+
+  // The compiler assumes this is Web fetch API, but it's actually Node.js fetch API
+  const res = await fetch(url, {
+    method: 'PUT',
+    // @ts-expect-error
+    body: uploadStream,
+    headers: {
+      'Content-Length': contentLength.toString(),
+    },
+    duplex: 'half',
+  })
+
+  if (!res.ok) {
+    throw new FileUploadError(
+      `Failed to upload file: ${res.statusText} ${res.status}`
+    )
+  }
+}
+
+export async function triggerBuild(
+  client: ApiClient,
+  { templateID, buildID, template }: TriggerBuildInput
+) {
+  const triggerBuildRes = await client.api.POST(
+    '/v2/templates/{templateID}/builds/{buildID}',
+    {
+      params: {
+        path: {
+          templateID,
+          buildID,
+        },
+      },
+      body: template,
+    }
+  )
+
+  const error = handleApiError(triggerBuildRes)
+  if (error) {
+    throw error
+  }
+}
+
+export async function getBuildStatus(
+  client: ApiClient,
+  { templateID, buildID, logsOffset }: GetBuildStatusInput
+) {
+  const buildStatusRes = await client.api.GET(
+    '/templates/{templateID}/builds/{buildID}/status',
+    {
+      params: {
+        path: {
+          templateID,
+          buildID,
+        },
+        query: {
+          logsOffset,
+        },
+      },
+    }
+  )
+
+  const error = handleApiError(buildStatusRes)
+  if (error) {
+    throw error
+  }
+
+  if (!buildStatusRes.data) {
+    throw new BuildError('Failed to get build status')
+  }
+
+  return buildStatusRes.data
+}
--- a/packages/js-sdk/src/template/dockerfileParser.ts
+++ b/packages/js-sdk/src/template/dockerfileParser.ts
@@ -1,0 +1,254 @@
+import { Instructions } from './types'
+import { DockerfileParser, Instruction, Argument } from 'dockerfile-ast'
+import fs from 'node:fs'
+
+export interface DockerfileParseResult {
+  baseImage: string
+  instructions: Instructions[]
+}
+
+export interface DockerfileParserInterface {
+  setWorkdir(workdir: string): void
+  setUser(user: string): void
+  setEnvs(envs: Record<string, string>): void
+  runCmd(command: string): void
+  copy(src: string, dest: string): void
+  setStartCmd(startCommand: string, readyCommand: string): void
+}
+
+/**
+ * Parse a Dockerfile and convert it to Template SDK format
+ *
+ * @param dockerfileContentOrPath Either the Dockerfile content as a string,
+ *                                or a path to a Dockerfile file
+ * @param templateBuilder Interface providing template builder methods
+ * @returns Parsed Dockerfile result with base image and instructions
+ */
+export function parseDockerfile(
+  dockerfileContentOrPath: string,
+  templateBuilder: DockerfileParserInterface
+): DockerfileParseResult {
+  // Check if input is a file path that exists
+  let dockerfileContent: string
+  try {
+    if (
+      fs.existsSync(dockerfileContentOrPath) &&
+      fs.statSync(dockerfileContentOrPath).isFile()
+    ) {
+      // Read the file content
+      dockerfileContent = fs.readFileSync(dockerfileContentOrPath, 'utf-8')
+    } else {
+      // Treat as content directly
+      dockerfileContent = dockerfileContentOrPath
+    }
+  } catch {
+    // If there's any error checking the file, treat as content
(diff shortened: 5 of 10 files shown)


mishushakov:
JavaScript SDK - Browser SDK tests are no longer passing, I believe due to some node-only dependencies.

packages/python-sdk/e2b/api/__init__.py
```diff
@@ -100,16 +107,20 @@ def __init__(
         kwargs.pop("auth_header_name", None)
         kwargs.pop("prefix", None)
 
+        # Only include limits in httpx_args if it's not None
```
dobrac: why?

packages/js-sdk/src/api/index.ts
```diff
@@ -7,12 +7,23 @@ import { AuthenticationError, RateLimitError, SandboxError } from '../errors'
...
 export function handleApiError(
-  response: FetchResponse<any, any, any>
+  response: FetchResponse<any, any, any>,
+  errorClass: new (message: string) => Error = SandboxError
 ): Error | undefined {
   if (!response.error) {
     return
   }
 
+  if (response.response.status === 401) {
```
dobrac, line 17: Is this a breaking change?

packages/js-sdk/src/template/index.ts
```diff
@@ -0,0 +1,639 @@
...
+
+  fromPythonImage(version: string = '3.13'): TemplateBuilder {
+    return this.fromImage(`python:${version}`)
+  }
+
+  fromNodeImage(variant: string = 'lts'): TemplateBuilder {
+    return this.fromImage(`node:${variant}`)
+  }
+
+  // maybe unnecessary
```
dobrac: todo? I think it's nice helper

packages/js-sdk/src/template/index.ts
```diff
@@ -0,0 +1,639 @@
...
+}
+
+export type BuildOptions = BasicBuildOptions & {
+  apiKey?: string
+  domain?: string
+}
+
+export class TemplateClass
+  implements TemplateFromImage, TemplateBuilder, TemplateFinal
+{
```
dobrac: could this implementation be moved out of the index file?

packages/js-sdk/src/index.ts
```diff
@@ -58,3 +60,5 @@ export type {
 export { Sandbox }
 import { Sandbox } from './sandbox'
 export default Sandbox
+
+export { Template } from './template'
```
dobrac: exporting also some type that will justify the builder/final types is necessary

packages/python-sdk/e2b/template/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+    TemplateBase,
+)
+from e2b.template_sync import Template
+from e2b.template_async import AsyncTemplate
+
+__all__ = [
+    "Template",
+    "AsyncTemplate",
+    "TemplateBuilder",
+    "TemplateFinal",
```
dobrac: can we maybe also export some time, that will be an umbrella for both of these? This type should be the same as the type that is passed to the build

packages/python-sdk/e2b/template/utils.py
```diff
@@ -0,0 +1,80 @@
...
+    return hash_obj.hexdigest()
+
+
+def strip_ansi_escape_codes(text: str) -> str:
+    # Valid string terminator sequences are BEL, ESC\, and 0x9c
+    st = r"(?:\u0007|\u001B\u005C|\u009C)"
+    pattern = [
+        rf"[\u001B\u009B][\[\]()#;?]*(?:(?:(?:(?:;[-a-zA-Z\d/#&.:=?%@~_]+)*|[a-zA-Z\d]+(?:;[-a-zA-Z\d/#&.:=?%@~_]*)*)?{st})",
+        r"(?:(?:\d{1,4}(?:;\d{0,4})*)?[\dA-PR-TZcf-nq-uy=><~]))",
+    ]
```
dobrac, line 64: is there any library that would have this? Or at least add a source of the regex

packages/python-sdk/e2b/template_async/__init__.py
```diff
@@ -0,0 +1,3 @@
+from e2b.template_async.main import AsyncTemplate
+
+__all__ = ["AsyncTemplate"]
```
dobrac, line 3: same as in JS

packages/python-sdk/e2b/template_sync/__init__.py
```diff
@@ -0,0 +1,3 @@
+from e2b.template_sync.main import Template
+
+__all__ = ["Template"]
```
dobrac, line 3: same as others

dobrac (changes requested):
\+ the Bugbot comment

Also, did you check the SDK reference for now?


\+ we should add proper documentation comments to all methods. We can do it in subsequent PR

packages/js-sdk/src/template/buildApi.ts
```diff
@@ -0,0 +1,167 @@
...
+  fileContextPath: string
+  url: string
+}) {
+  const { fileName, url, fileContextPath } = options
+  const { contentLength, uploadStream } = await tarFileStreamUpload(
+    fileName,
+    fileContextPath
+  )
+
+  // The compiler assumes this is Web fetch API, but it's actually Node.js fetch API
```
jakubno, line 99: is there something wrong with the config or why does it happen?

packages/js-sdk/src/template/dockerfileParser.ts
```diff
@@ -0,0 +1,254 @@
...
+        // VOLUME is not directly supported in our SDK, so we'll skip it
+        break
+
+      case 'CMD':
+      case 'ENTRYPOINT':
+        handleCmdEntrypointInstruction(instruction, templateBuilder)
+        break
+
+      default:
+        // For unsupported instructions, we could log a warning or add as raw commands
```
jakubno: is there a reason why not to print a warning for all not directly supported keywords?

packages/python-sdk/e2b/api/__init__.py
mishushakov: was causing an error because the API Client in the Templates SDK didn't provide the request/connection limits.
I've set the same ones now as the Sandbox, but maybe we'll have to revisit this.

packages/js-sdk/tests/template/folder/test.txt
jakubno, line 1: you could probably create it during the test, but it's fine

packages/js-sdk/src/api/index.ts
mishushakov: yes, before it was raising SandboxError (which was incorrect btw.,), now it will raise AuthenticationError. Keep in mind that I have also changed the AuthenticationError to extend Error class, not SandboxError anymore as it's now thrown by both Sandbox and Template.

packages/python-sdk/e2b/template/dockerfile_parser.py
```diff
@@ -0,0 +1,258 @@
...
+            elif instruction in [
+                "EXPOSE",
+                "VOLUME",
+                "LABEL",
+                "HEALTHCHECK",
+                "ONBUILD",
+                "STOPSIGNAL",
+                "SHELL",
+            ]:
+                # Unsupported instructions - skip silently
```
jakubno: maybe warning?

packages/python-sdk/e2b/template/utils.py
mishushakov: There was no package similar to strip-ansi in Python, so I vibe-coded myself.

@dobrac do you think we actually need this in the SDK? can it be done on the backend for the structured logs?

Thinking:

```py
class LogEntry:
    timestamp: datetime
    level: Literal["debug", "info", "warn", "error"]
    message: str # removed ansii chars
    message_raw: str # raw message
```

packages/js-sdk/src/errors.ts
```diff
@@ -68,7 +68,7 @@ export class NotFoundError extends SandboxError {
 /**
  * Thrown when authentication fails.
  */
-export class AuthenticationError extends SandboxError {
+export class AuthenticationError extends Error {
```
mishushakov, line 71: this is a breaking change

packages/python-sdk/e2b/exceptions.py
```diff
@@ -63,7 +63,7 @@ class NotFoundException(SandboxException):
     pass
 
 
-class AuthenticationException(SandboxException):
+class AuthenticationException(Exception):
```
mishushakov, line 66: this is a breaking change

packages/python-sdk/e2b/template/main.py
```diff
@@ -0,0 +1,419 @@
...
+
+    @classmethod
+    def wait_for_file(cls, filename: str) -> str:
+        """Generate a command to wait for a file to exist."""
+        return f"[ -f {filename} ]"
+
+    @classmethod
+    def wait_for_timeout(cls, timeout: Duration) -> str:
+        """Generate a command to wait for a specified duration."""
+        return f"sleep {timeout}"
```
jakubno: This doesn't have to work on all platforms, should be okay for debian based images, but if we want to use this I would prefer to convert it in SDK

packages/python-sdk/e2b/template/utils.py
dobrac: I think link to here (with the regex from here) might be enough: [https://github.com/chalk/ansi-regex/blob/main/index.js](https://github.com/chalk/ansi-regex/blob/main/index.js)

packages/python-sdk/e2b/template_async/main.py
```diff
@@ -0,0 +1,197 @@
...
+                        message="Waiting for logs...",
+                    )
+                )
+
+            await AsyncTemplate.__wait_for_build_finish(
+                api_client, template_id, build_id, on_build_logs
+            )
+
+    @staticmethod
+    async def __wait_for_build_finish(
```
jakubno: does it have to be on the template, wouldn't it be better to have it just as a separate function?

packages/python-sdk/tests/async/template_async/folder/test.txt
jakubno, line 1: same as in JS

packages/python-sdk/tests/sync/template_sync/folder/test.txt
jakubno, line 1: same

packages/python-sdk/e2b/template/main.py
mishushakov: I think we can make it work, because @dobrac will inject some common packages like curl, sudo, etc. into the sandbox in the provisioning script

packages/js-sdk/src/template/buildApi.ts
mishushakov: because we use global fetch, the TS compiler thinks it's frontend fetch
https://github.com/e2b-dev/E2B/blob/main/packages/js-sdk/tsconfig.json#L6

packages/python-sdk/e2b/template/dockerfile_parser.py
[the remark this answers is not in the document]
dobrac: this exact library should be there used

packages/python-sdk/e2b/template/main.py
jakubno: Leaving this up to @dobrac, it should be okay for most of distribution, but not all have to support it. I still think it's safer to resolve it already in SDK

dobrac: By SDK, do you mean the Infra?

dobrac: irl discussion, it's not about the `sleep` command not being available, but that the `sleep` command doesn't have to support the format `3s` or `1m` in other distributions. So we should convert it always to seconds (just a number) in the SDK already

(same applies to JS)

@mishushakov

packages/js-sdk/src/template/index.ts
mishushakov: can you elaborate?

packages/python-sdk/e2b/template_async/__init__.py
mishushakov: you meant exporting the template class?

packages/python-sdk/e2b/template_sync/__init__.py
mishushakov: you meant exporting the template class?

packages/js-sdk/src/template/index.ts
dobrac: meaning that the index will be just reexport of the implementations, instead of the actual implementation<br><br>e.g. at the bottom you're defining some of the helpers, etc<br><br>just simple barrel file

packages/python-sdk/e2b/template_sync/__init__.py
dobrac: exporting the types

packages/python-sdk/e2b/template_async/__init__.py
dobrac: exporting the types

packages/js-sdk/src/template/index.ts
mishushakov: @dobrac it's currently same pattern as with sandbox

packages/js-sdk/src/index.ts
```diff
@@ -61,4 +61,4 @@ export { Sandbox }
 import { Sandbox } from './sandbox'
 export default Sandbox
 
-export { Template, TemplateClass } from './template'
+export { Template, type TemplateClass, TemplateBase } from './template'
```
dobrac: TemplateBase is also type only

packages/js-sdk/src/template/index.ts
```diff
@@ -69,18 +69,15 @@ export class TemplateClass
   }
 
   static toJSON(template: TemplateBuilder | TemplateFinal): Promise<string> {
-    return (template as TemplateClass).toJSON()
+    return (template as TemplateBase).toJSON()
```
dobrac, line 72: why is this necessary? Can't we avoid it with proper types?

packages/js-sdk/src/template/index.ts
```diff
@@ -69,18 +69,15 @@ export class TemplateClass
   }
 
   static toJSON(template: TemplateBuilder | TemplateFinal): Promise<string> {
```
dobrac: The type in the function should be TemplateClass, otherwise it's confusing what type needs to be passed

packages/js-sdk/src/template/index.ts
```diff
@@ -591,3 +588,5 @@ Template.waitForFile = function (filename: string) {
 Template.waitForTimeout = function (timeout: number) {
   return `sleep ${Math.floor(timeout / 1000)}`
 }
+
+export type TemplateClass = TemplateBuilder
```
dobrac: will it be possible to do this?

const template: TemplateClass = tmpl.setStartCmd(...)

packages/python-sdk/e2b/template/main.py
```diff
@@ -0,0 +1,432 @@
...
+        self._template._ready_cmd = ready_cmd
+        return TemplateFinal(self._template)
+
+    def set_ready_cmd(self, ready_cmd: str) -> "TemplateFinal":
+        self._template._ready_cmd = ready_cmd
+        return TemplateFinal(self._template)
+
+
+class TemplateFinal:
+    """Final stage - only start/ready commands and utility methods available"""
```
dobrac: outdated comment

packages/js-sdk/src/template/index.ts
mishushakov: this is so annoying and I see the point. I will make a follow-up PR and rewrite the builder to follow same structure with multiple classes as the Python SDK

packages/js-sdk/src/index.ts
mishushakov: this was needed explicitly for TemplateClass so it only exports its type not class

packages/js-sdk/src/template/index.ts
mishushakov: no

Property 'setStartCmd' does not exist on type 'TemplateFromImage'.

but 

```ts
const template: TemplateClass = Template().fromImage('ubuntu:22.04').setStartCmd('echo "Hello, world!"', Template.waitForTimeout(10_000))

```

should be possible

dobrac (approved):

PR #880, opened 2025-08-27
add template migration to CLI

dobrac:
Add template migration to CLI


packages/cli/src/commands/template/generators/handlebars.ts
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * Transform template data for Handlebars
+ */
+export async function transformTemplateData(template: TemplateClass) {
+  // Extract JSON structure from parsed template
+  const jsonString = await Template.toJSON(template, false)
+  const json = JSON.parse(jsonString) as TemplateJSON
+
+  const transformedSteps: any[] = []
```
mishushakov: no any

packages/cli/src/commands/template/generators/handlebars.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  const templateTemplate = Handlebars.compile(templateSource)
+  const buildTemplate = Handlebars.compile(buildSource)
+
+  // Generate content
+  const templateData = {
+    ...transformedData,
+  }
+
+  const templateContent = templateTemplate(templateData)
```
mishushakov: templateTemplate is confusing

packages/cli/src/commands/template/generators/handlebars.ts
```diff
@@ -0,0 +1,177 @@
...
+  Handlebars.registerHelper('escapeQuotes', function (str) {
+    return str ? str.replace(/'/g, "\\'") : str
+  })
+
+  Handlebars.registerHelper('escapeDoubleQuotes', function (str) {
+    return str ? str.replace(/"/g, '\\"') : str
+  })
+
+  helpersRegistered = true
+}
```
mishushakov: maybe it's better to make it a class that extends Handlebars?

packages/cli/src/commands/template/generators/handlebars.ts
```diff
@@ -0,0 +1,177 @@
...
+  const buildSource = fs.readFileSync(
+    path.join(templatesDir, `python-build-${isAsync ? 'async' : 'sync'}.hbs`),
+    'utf8'
+  )
+
+  const templateTemplate = Handlebars.compile(templateSource)
+  const buildTemplate = Handlebars.compile(buildSource)
+
+  // Generate content
+  const templateContent = templateTemplate({
```
mishushakov: same as above

packages/cli/src/commands/template/generators/handlebars.ts
mishushakov: maye generateTemplateSource

packages/cli/src/commands/template/generators/handlebars.ts
```diff
@@ -0,0 +1,177 @@
...
+  const buildTemplate = Handlebars.compile(buildSource)
+
+  // Generate content
+  const templateData = {
+    ...transformedData,
+  }
+
+  const templateContent = templateTemplate(templateData)
+
+  const buildContent = buildTemplate({
```
mishushakov: maybe generateTemplateBuildSource

packages/cli/src/commands/template/generators/handlebars.ts
```diff
@@ -0,0 +1,177 @@
...
+  const templateTemplate = Handlebars.compile(templateSource)
+  const buildTemplate = Handlebars.compile(buildSource)
+
+  // Generate content
+  const templateContent = templateTemplate({
+    ...transformedData,
+    isAsync,
+  })
+
+  const buildContent = buildTemplate({
```
mishushakov: same as above

packages/cli/src/commands/template/generators/file-utils.ts
```diff
@@ -0,0 +1,38 @@
+import * as fs from 'fs'
+import * as path from 'path'
+
+/**
+ * Generate unique file names to avoid overwriting existing files
+ */
+export function getUniqueFileName(
```
mishushakov: I think we should abort if files already exist

packages/cli/src/commands/template/generators/template-generator.ts
```diff
@@ -0,0 +1,84 @@
...
+      template,
+      alias,
+      cpuCount,
+      memoryMB
+    )
+
+    const extension = '.ts'
+    const templateFile = getUniqueFileName(root, 'template', extension)
+    const buildDevFile = getUniqueFileName(root, 'build.dev', extension)
+    const buildProdFile = getUniqueFileName(root, 'build.prod', extension)
```
mishushakov: abort if exists

packages/cli/src/commands/template/generators/template-generator.ts
```diff
@@ -0,0 +1,84 @@
...
+ */
+export async function generateAndWriteTemplateFiles(
+  root: string,
+  alias: string,
+  language: Language,
+  template: TemplateClass,
+  cpuCount?: number,
+  memoryMB?: number
+): Promise<GeneratedFiles> {
+  if (language === Language.TypeScript) {
```
mishushakov: maybe it should be switch/case

packages/cli/src/commands/template/generators/template-generator.ts
```diff
@@ -0,0 +1,84 @@
...
+      alias,
+      cpuCount,
+      memoryMB,
+      isAsync
+    )
+
+    const extension = '.py'
+    const templateFile = getUniqueFileName(root, 'template', extension)
+    const buildDevFile = getUniqueFileName(root, 'build_dev', extension)
+    const buildProdFile = getUniqueFileName(root, 'build_prod', extension)
```
mishushakov: abort if exists

packages/cli/src/commands/template/migrate.ts
```diff
@@ -0,0 +1,212 @@
+import { select } from '@inquirer/prompts'
+import * as commander from 'commander'
+import { Template, TemplateBuilder, TemplateClass } from 'e2b'
+import * as fs from 'fs'
+import * as path from 'path'
+import { E2BConfig, getConfigPath, loadConfig } from 'src/config'
+import { defaultDockerfileName } from 'src/docker/constants'
+import { pathOption } from 'src/options'
+import { getRoot } from 'src/utils/filesystem'
+import { asLocal, asLocalRelative, asPrimary } from 'src/utils/format'
```
mishushakov: these should be relative

packages/cli/src/commands/template/generators/template-generator.ts
```diff
@@ -0,0 +1,84 @@
+import * as path from 'path'
+import { asLocalRelative, asPrimary } from 'src/utils/format'
```
mishushakov: use relative import

packages/cli/src/commands/template/migrate.ts
```diff
@@ -0,0 +1,212 @@
...
+    )} to new Template SDK format`
+  )
+  .option(
+    '-d, --dockerfile <file>',
+    `specify path to Dockerfile. Defaults to ${asLocal('e2b.Dockerfile')}`
+  )
+  .option(
+    '-c, --config <file>',
+    `specify path to config file. Defaults to ${asLocal('e2b.toml')}`
+  )
```
mishushakov: I am not sure if we want them as separate options, is it for cases where you have 2 different .toml but using same Dockerfile?

packages/cli/src/commands/template/migrate.ts
```diff
@@ -0,0 +1,212 @@
...
+    (value) => {
+      if (!validLanguages.includes(value as Language)) {
+        throw new Error(
+          `Invalid language. Must be one of: ${validLanguages.join(', ')}`
+        )
+      }
+      return value as Language
+    }
+  )
+  .addOption(pathOption)
```
mishushakov, line 112: what is pathOption?

packages/cli/src/commands/template/migrate.ts
```diff
@@ -0,0 +1,212 @@
...
+  generateAndWriteTemplateFiles,
+  Language,
+  languageDisplay,
+  validLanguages,
+} from './generators'
+
+/**
+ * Migrate Dockerfile to a specific target language using SDK
+ */
+async function migrateToLanguage(
```
mishushakov, line 21: I think this should be somewhere in utils, here is the only logic for the command

packages/cli/tests/commands/template/migrate.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    async function runMigrationTest(testCaseName: string, language: string) {
+      const fixtureDir = path.join(fixturesDir, testCaseName)
+
+      // Copy fixture files to test directory
+      await copyFixtureFiles(fixtureDir, testDir)
+
+      // Run migration
+      execSync(`node ${cliPath} template migrate --language ${language}`, {
+        cwd: testDir,
+      })
```
mishushakov, line 90: there must be a way to run the CLI logic without exec

packages/cli/tests/commands/template/migrate.test.ts
```diff
@@ -0,0 +1,177 @@
+import { describe, test, expect, beforeEach, afterEach } from 'vitest'
+import * as fs from 'fs/promises'
+import * as path from 'path'
+import { execSync } from 'child_process'
+
+describe('Template Migration', () => {
+  let testDir: string
+  const cliPath = path.join(process.cwd(), 'dist', 'index.js')
```
mishushakov, line 37: you'll have to recompile the CLI before running the tests, which might be confusing

packages/cli/tests/commands/template/migrate.test.ts
```diff
@@ -0,0 +1,177 @@
...
+      // Run migration
+      execSync(`node ${cliPath} template migrate --language ${language}`, {
+        cwd: testDir,
+      })
+
+      // Determine file extensions and names based on language
+      const isTypescript = language === 'typescript'
+      const templateExt = isTypescript ? '.ts' : '.py'
+      const buildDevName = isTypescript ? 'build.dev.ts' : 'build_dev.py'
+      const buildProdName = isTypescript ? 'build.prod.ts' : 'build_prod.py'
```
mishushakov: maybe make it function (s), looks confusing otherwise

packages/cli/tests/commands/template/migrate.test.ts
```diff
@@ -0,0 +1,177 @@
...
+
+      // Load expected outputs
+      const expectedDir = path.join(fixtureDir, 'expected', language)
+      const expectedTemplate = await fs.readFile(
+        path.join(expectedDir, `template${templateExt}`),
```
mishushakov: make `template${templateExt}` into a variable

packages/cli/tests/commands/template/migrate.test.ts
```diff
@@ -0,0 +1,177 @@
...
+        'utf-8'
+      )
+      const expectedBuildProd = await fs.readFile(
+        path.join(expectedDir, buildProdName),
+        'utf-8'
+      )
+
+      // Check generated files
+      const templateFile = await fs.readFile(
+        path.join(testDir, `template${templateExt}`),
```
mishushakov: same as above

packages/cli/tests/commands/template/migrate.test.ts
```diff
@@ -0,0 +1,177 @@
...
+      // Copy Dockerfile and config
+      await fs.copyFile(
+        path.join(fixtureDir, 'e2b.Dockerfile'),
+        path.join(targetDir, 'e2b.Dockerfile')
+      )
+      await fs.copyFile(
+        path.join(fixtureDir, 'e2b.toml'),
+        path.join(targetDir, 'e2b.toml')
+      )
+    }
```
mishushakov, line 140: it's used at the top but defined here

packages/cli/tests/commands/template/migrate.test.ts
```diff
@@ -0,0 +1,177 @@
...
+      // Run migration and expect it to fail
+      expect(() => {
+        execSync(`node ${cliPath} template migrate --language typescript`, {
+          cwd: testDir,
+        })
+      }).toThrow()
+    })
+  })
+
+  describe('File Collision Handling', () => {
```
mishushakov, line 172: this should just check if there's an error

packages/cli/package.json
```diff
@@ -40,24 +40,28 @@
     "dev": "tsup --watch",
     "lint": "eslint src",
     "format": "prettier --write src",
-    "test": "pnpm build && cd testground/demo-basic && ../../dist/index.js template build",
-    "test:interactive": "pnpm build && ./dist/index.js",
```
mishushakov, line 44: this should remain for manual tests

packages/cli/tsup.config.js
```diff
@@ -19,4 +19,6 @@ export default defineConfig({
     options.legalComments = 'none'
     return options
   },
+  // Copy template files to dist
+  onSuccess: 'cp -r src/templates dist/',
```
mishushakov, line 23: there should be some way to include them in dist without cp

packages/python-sdk/e2b/template_async/main.py
```diff
@@ -22,8 +22,8 @@ class AsyncTemplate(TemplateBase):
     async def build(
         template: TemplateClass,
         alias: str,
-        cpu_count: int,
-        memory_mb: int,
+        cpu_count: int = 2,
+        memory_mb: int = 512,
```
mishushakov: this should be defined in generated files, but no default here

packages/python-sdk/e2b/template_sync/main.py
```diff
@@ -22,8 +22,8 @@ class Template(TemplateBase):
     def build(
         template: TemplateClass,
         alias: str,
-        cpu_count: int,
-        memory_mb: int,
+        cpu_count: int = 2,
+        memory_mb: int = 512,
```
mishushakov: same as above

packages/cli/src/commands/template/generators/types.ts
```diff
@@ -0,0 +1,38 @@
...
+  TypeScript = 'typescript',
+  PythonSync = 'python-sync',
+  PythonAsync = 'python-async',
+}
+
+export const validLanguages: Language[] = [
+  Language.TypeScript,
+  Language.PythonSync,
+  Language.PythonAsync,
+]
```
mishushakov: you can just check if enum includes value

```js
Object.values(Language).includes(value as T[keyof T]);
```

packages/cli/src/commands/template/migrate.ts
```diff
@@ -0,0 +1,212 @@
...
+  )
+  .option(
+    '-c, --config <file>',
+    `specify path to config file. Defaults to ${asLocal('e2b.toml')}`
+  )
+  .option(
+    '-l, --language <language>',
+    `specify target language: ${validLanguages.join(', ')}`,
+    (value) => {
+      if (!validLanguages.includes(value as Language)) {
```
mishushakov: you can just check if enum includes value

```
Object.values(Language).includes(value as T[keyof T]);
```

packages/cli/src/commands/template/migrate.ts
dobrac: "change root directory where command is executed"

packages/cli/src/commands/template/generators/file-utils.ts
dobrac: why do you think so?

packages/cli/src/commands/template/generators/template-generator.ts
dobrac: duplicit

packages/cli/tests/commands/template/migrate.test.ts
dobrac: that should be fine, functions defined like this are hoisted to the top

packages/cli/tests/commands/template/migrate.test.ts
dobrac: right, any suggestions how to handle it better?

packages/cli/src/commands/template/migrate.ts
dobrac: I'll remove the option to define e2b.toml, we don't have it in the build command

dobrac: actually, we do, so I'll keep it to have parity

packages/python-sdk/e2b/template_async/main.py
dobrac: We should have a default for CPU and memory. We have the default in JS

packages/cli/tsup.config.js
dobrac: nope: https://github.com/egoist/tsup/issues/278?utm_source=chatgpt.com

dobrac: but we can migrate to [tsdown](https://tsdown.dev/guide/migrate-from-tsup) which supports `copy`

packages/cli/tests/commands/template/migrate.test.ts
dobrac: I wanted to test it in a way how user would use it

packages/cli/tests/commands/template/migrate.test.ts
dobrac: added build before running tests (it's fairly fast)

packages/cli/src/commands/template/migrate.ts
dobrac: is this actually the migrate command implementation more than any util?

packages/cli/tsup.config.js
mishushakov: okay, leave as is for now then

packages/cli/src/commands/template/generators/template-generator.ts
```diff
@@ -0,0 +1,95 @@
...
+      )
+
+      const extension = '.ts'
+      const templateFile = errorIfExists(root, 'template', extension)
+      const buildDevFile = errorIfExists(root, 'build.dev', extension)
+      const buildProdFile = errorIfExists(root, 'build.prod', extension)
+
+      await writeFileContent(path.join(root, templateFile), templateContent)
+      await writeFileContent(path.join(root, buildDevFile), buildDevContent)
+      await writeFileContent(path.join(root, buildProdFile), buildProdContent)
```
mishushakov: can you move the errorIfExists to writeFileContent

packages/cli/tests/commands/template/migrate.test.ts
mishushakov: I think it's fine with the build

packages/cli/src/commands/template/generators/handlebars.ts
```diff
@@ -0,0 +1,197 @@
+import { Template, TemplateClass } from 'e2b'
+import * as fs from 'fs'
+import HandlebarsLib from 'handlebars'
+import * as path from 'path'
+import { TemplateJSON, TemplateWithStepsJSON } from './types'
+
+// Track if helpers are registered to avoid duplicate registration
+let helpersRegistered = false
```
mishushakov: this should be a class property

PR #881, opened 2025-08-28
feat: add template init for new SDK

dobrac:
## 🚀 Add template init for new SDK

This PR updates the `e2b template init-v2` (or `e2b template create`) command to leverage the new E2B Templates SDK, providing a more streamlined and developer-friendly experience for creating sandbox templates.

The command will be migrated to just `init` once the Template SDK becomes the default to create new templates.

#### **Template Structure**
```
my-template/
├── README.md            # Information about the template
├── template.ts          # Template definition using E2B SDK
├── build.dev.ts         # Development build script
└── build.prod.ts        # Production build script
```

#### **Generated Build Scripts**
- `npm run e2b:build:dev` - Build development template
- `npm run e2b:build:prod` - Build production template

#### **Template Definition Example**
```typescript
const template = Template()
  .fromImage('ubuntu:22.04')
  .runCmd('echo Hello World E2B!')
```

### DIFF
--- a/packages/cli/src/commands/template/init-v2.ts
+++ b/packages/cli/src/commands/template/init-v2.ts
@@ -111,19 +111,13 @@ async function addPackageJsonScripts(
       pkgJson = await PackageJson.create(createRoot)
     }
 
-    switch (files.language) {
-      case Language.TypeScript:
-        pkgJson.update({
-          scripts: {
-            ...pkgJson.content.scripts,
-            'e2b:build:dev': `${cdPrefix}npx tsx ${files.buildDevFile}`,
-            'e2b:build:prod': `${cdPrefix}npx tsx ${files.buildProdFile}`,
-          },
-        })
-        break
-      default:
-        throw new Error('Unsupported language for package.json scripts')
-    }
+    pkgJson.update({
+      scripts: {
+        ...pkgJson.content.scripts,
+        'e2b:build:dev': `${cdPrefix}npx tsx ${files.buildDevFile}`,
+        'e2b:build:prod': `${cdPrefix}npx tsx ${files.buildProdFile}`,
+      },
+    })
 
     // Save the changes
     await pkgJson.save()


packages/cli/src/commands/template/generators/handlebars.ts
```diff
@@ -195,3 +200,40 @@ export async function generatePythonCode(
...
+    templateDir,
+    templateFile: generatedFiles.templateFile,
+    buildDevFile: generatedFiles.buildDevFile,
+    buildProdFile: generatedFiles.buildProdFile,
+    isTypeScript: generatedFiles.language === Language.TypeScript,
+    isPython:
+      generatedFiles.language === Language.PythonSync ||
+      generatedFiles.language === Language.PythonAsync,
+    isPythonSync: generatedFiles.language === Language.PythonSync,
+    isPythonAsync: generatedFiles.language === Language.PythonAsync,
```
mishushakov, line 229: this should be just language

packages/cli/src/commands/template/init-v2.ts
```diff
@@ -0,0 +1,271 @@
...
+ */
+async function generateTemplateFiles(
+  root: string,
+  alias: string,
+  language: Language,
+  cpuCount?: number,
+  memoryMB?: number
+): Promise<GeneratedFiles> {
+  const template = Template()
+    .fromImage('ubuntu:22.04')
```
mishushakov: .fromBaseImage()

packages/cli/src/commands/template/init-v2.ts
```diff
@@ -0,0 +1,271 @@
...
+): Promise<void> {
+  try {
+    // Use @npmcli/package-json for robust handling
+    // The library expects the directory path, not the full file path
+    const pkgJson = await PackageJson.load(root)
+
+    // Generate script commands based on language and directory structure
+    const cdPrefix = templateDirName ? `cd ${templateDirName} && ` : ''
+
+    if (files.language === Language.TypeScript) {
```
mishushakov: switch case

packages/cli/src/commands/template/init-v2.ts
```diff
@@ -0,0 +1,271 @@
...
+        },
+      })
+    } else {
+      pkgJson.update({
+        scripts: {
+          ...pkgJson.content.scripts,
+          'e2b:build:dev': `${cdPrefix}python ${files.buildDevFile}`,
+          'e2b:build:prod': `${cdPrefix}python ${files.buildProdFile}`,
+        },
+      })
```
mishushakov: in Python Makefiles are more common than package.json. Also to run scripts in package.json you'll need to have Node installed.

packages/cli/src/commands/template/init-v2.ts
```diff
@@ -0,0 +1,271 @@
...
+  .option('-n, --name <name>', 'template name (alias)')
+  .option(
+    '-l, --language <language>',
+    `target language: ${validLanguages.join(', ')}`,
+    (value) => {
+      if (!validLanguages.includes(value as Language)) {
+        throw new Error(
+          `Invalid language. Must be one of: ${validLanguages.join(', ')}`
+        )
+      }
```
mishushakov: should not throw, but console.error and process.exit with 1 error code instead (please make sure it's same in migrate)

packages/cli/src/commands/template/init-v2.ts
```diff
@@ -0,0 +1,271 @@
...
+ * Validate template name format
+ */
+function validateTemplateName(name: string): boolean {
+  return /^[a-z0-9][a-z0-9_-]*[a-z0-9]$|^[a-z0-9]$/.test(name)
+}
+
+export const initV2Command = new commander.Command('init-v2')
+  .description('initialize a new sandbox template using the SDK')
+  .addOption(pathOption)
+  .option('-n, --name <name>', 'template name (alias)')
```
mishushakov: validation logic here?

packages/cli/src/commands/template/init-v2.ts
```diff
@@ -0,0 +1,271 @@
...
+        // Step 1: Get template name (from CLI or prompt)
+        let templateName: string
+        if (opts.name) {
+          if (!validateTemplateName(opts.name)) {
+            throw new Error(
+              'Template name must contain only lowercase letters, numbers, hyphens, and underscores, and cannot start or end with a hyphen or underscore'
+            )
+          }
+          templateName = opts.name
+          console.log(`Using template name: ${templateName}`)
```
mishushakov: you can move the validation to args

packages/cli/src/commands/template/init-v2.ts
```diff
@@ -0,0 +1,271 @@
...
+        const readmeFilePath = path.join(templateDir, 'README.md')
+        await fs.promises.writeFile(readmeFilePath, readmeContent, 'utf8')
+
+        console.log('\n🎉 Template initialized successfully!')
+        console.log(
+          `\nTemplate created in: ${asPrimary(`./${templateDirName}/`)}`
+        )
+        console.log('\nYou can now build your template using:')
+
+        if (language === Language.TypeScript) {
```
mishushakov: switch case

packages/cli/src/commands/template/init-v2.ts
```diff
@@ -0,0 +1,271 @@
...
+          console.log(
+            `   ${asPrimary(
+              `cd ${templateDirName} && python ${generatedFiles.buildDevFile}`
+            )} (for development)`
+          )
+          console.log(
+            `   ${asPrimary(
+              `cd ${templateDirName} && python ${generatedFiles.buildProdFile}`
+            )} (for production)`
+          )
```
mishushakov: make

packages/cli/src/templates/readme.hbs
```diff
@@ -0,0 +1,109 @@
...
+// Your sandbox is ready to use!
+console.log('Sandbox created successfully')
+```
+{{/if}}
+{{#if isPythonSync}}
+```python
+from e2b import Sandbox
+
+# Create a new sandbox instance
+sandbox = Sandbox('{{alias}}')
```
mishushakov: Sandbox.create (since E2B SDK v2)

packages/cli/tests/commands/template/init.test.ts
```diff
@@ -0,0 +1,305 @@
...
+import * as path from 'path'
+import { afterEach, beforeEach, describe, expect, test } from 'vitest'
+
+describe('Template Init', () => {
+  let testDir: string
+  const cliPath = path.join(process.cwd(), 'dist', 'index.js')
+
+  beforeEach(async () => {
+    // Use Node.js built-in temp directory handling
+    testDir = await fs.mkdtemp('e2b-init-test-')
```
mishushakov, line 14: test- ?

packages/cli/tests/commands/template/init.test.ts
```diff
@@ -0,0 +1,305 @@
...
+  })
+
+  afterEach(async () => {
+    // Clean up test directory
+    if (testDir) {
+      await fs.rm(testDir, { recursive: true, force: true })
+    }
+  })
+
+  const languages = ['typescript', 'python-sync', 'python-async'] as const
```
mishushakov: don't we have an enum for that?

packages/cli/tests/commands/template/init.test.ts
```diff
@@ -0,0 +1,305 @@
...
+        // Verify files were created in the template directory
+        const expectedFiles = getExpectedFiles('typescript')
+        for (const file of expectedFiles) {
+          expect(existsSync(path.join(templateDir, file))).toBe(true)
+        }
+      }
+    })
+  })
+
+  describe('Package.json Integration', () => {
```
mishushakov, line 133: make for Python

packages/cli/tests/commands/template/init.test.ts
```diff
@@ -0,0 +1,305 @@
...
+      const updatedPackageJson = JSON.parse(
+        await fs.readFile(packageJsonPath, 'utf8')
+      )
+
+      expect(updatedPackageJson.scripts).toHaveProperty('e2b:build:dev')
+      expect(updatedPackageJson.scripts).toHaveProperty('e2b:build:prod')
+      expect(updatedPackageJson.scripts.test).toBe('echo "test"') // existing script preserved
+    })
+
+    test('should work without package.json', async () => {
```
mishushakov, line 162: same as above

packages/cli/src/commands/template/init-v2.ts
```diff
@@ -0,0 +1,271 @@
...
+ */
+async function addPackageJsonScripts(
+  root: string,
+  files: GeneratedFiles,
+  templateDirName?: string
+): Promise<void> {
+  try {
+    // Use @npmcli/package-json for robust handling
+    // The library expects the directory path, not the full file path
+    const pkgJson = await PackageJson.load(root)
```
mishushakov: if you can't load package.json you should make a new one instead

packages/cli/src/commands/template/generators/handlebars.ts
mishushakov: or can't Handlebars handle enums/string?

packages/cli/src/templates/readme.hbs
```diff
@@ -0,0 +1,109 @@
...
+{{/if}}
+{{#if isPython}}
+```bash
+# For development
+cd {{templateDir}} && python {{buildDevFile}}
+
+# For production  
+cd {{templateDir}} && python {{buildProdFile}}
+```
+{{/if}}
```
mishushakov, line 34: maybe switch/case here also if possible and pass the template variant from generateReadmeContent

packages/cli/tests/commands/template/init.test.ts
dobrac: it's prefix

packages/cli/src/templates/readme.hbs
dobrac: switch case is not supported in the handlebars language, but I've added `else if`

packages/cli/src/commands/template/generators/handlebars.ts
dobrac: it can handle strings, but I think it will be more stable when you for example change the string value in the enum

(it doesn't know enums)

packages/cli/src/commands/template/init-v2.ts
```diff
@@ -0,0 +1,332 @@
...
+            break
+          default:
+            throw new Error('Unsupported language for instructions')
+        }
+
+        console.log(
+          `\nLearn more about Sandbox Templates: ${asPrimary(
+            'https://e2b.dev/docs'
+          )}\n`
+        )
```
mishushakov, line 320: this can be moved to some utility function like we have in other places

packages/cli/src/commands/template/init-v2.ts
```diff
@@ -0,0 +1,332 @@
...
+    async (opts: { path?: string; name?: string; language?: Language }) => {
+      try {
+        process.stdout.write('\n')
+
+        const root = getRoot(opts.path)
+
+        console.log('🚀 Initializing Sandbox Template...\n')
+
+        // Step 1: Get template name (from CLI or prompt)
+        let templateName: string = opts.name ?? DEFAULT_TEMPLATE_NAME
```
mishushakov, line 198: I don't think you need `?? DEFAULT_TEMPLATE_NAME` because it will be set in the input? and if you clear the pre-defined name it should throw a validation error anyways

packages/cli/src/commands/template/init-v2.ts
```diff
@@ -0,0 +1,332 @@
...
+          scripts: {
+            ...pkgJson.content.scripts,
+            'e2b:build:dev': `${cdPrefix}npx tsx ${files.buildDevFile}`,
+            'e2b:build:prod': `${cdPrefix}npx tsx ${files.buildProdFile}`,
+          },
+        })
+        break
+      default:
+        throw new Error('Unsupported language for package.json scripts')
+    }
```
mishushakov: this condition can never occur? it will always be TypeScript

mishushakov (changes requested):
minor nits

packages/cli/src/commands/template/init-v2.ts
dobrac: I need it just for the type safety (to have `string`, not `string | undefined`)

packages/cli/src/commands/template/init-v2.ts
dobrac: there is no utility function in other places (but it would be nice). We can do a refactoring of this in a follow up PR

PR #889, opened 2025-09-03
Synced spec + generated API client

mishushakov:
Changelog

- Synced latest spec from e2b-dev/infra
- Regenerated API clients

### DIFF
--- a/packages/cli/src/utils/format.ts
+++ b/packages/cli/src/utils/format.ts
@@ -29,10 +29,9 @@ export function asFormattedTeam(
 }
 
 export function asFormattedSandboxTemplate(
-  template: Pick<
-    e2b.components['schemas']['Template'],
-    'templateID' | 'aliases'
-  >,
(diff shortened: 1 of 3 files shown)


packages/cli/src/utils/format.ts
```diff
@@ -29,10 +29,9 @@ export function asFormattedTeam(
 }
 
 export function asFormattedSandboxTemplate(
-  template: Pick<
-    e2b.components['schemas']['Template'],
-    'templateID' | 'aliases'
-  >,
+  template: Pick<e2b.components['schemas']['Template'], 'templateID'> & {
+    aliases?: e2b.components['schemas']['Template']['aliases']
```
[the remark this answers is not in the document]
dobrac: what would it break?

mishushakov, line 33: in this function aliases can be optional, because some methods don't pass the aliases

mishushakov: @dobrac clarified

PR #891, opened 2025-09-05
Templates SDK JS - Use classes instead of interface

mishushakov:
Change 

In JavaScript specifically this is implemented using interfaces unlike classes in Python. I've found some issues with the interfaces-based approach - you're able to pass empty TemplateBase to Template.build when you are not supposed to and also there was also much duplication in types.ts.

It's also probably not safe at all when using pure JS without TypeScript as you won't see compiler errors.

it was working until this point, because we had other methods on the TemplateFinal interface, but it's now empty so it doesn't match anything anymore.


packages/js-sdk/src/template/dockerfileParser.ts
```diff
@@ -70,10 +74,10 @@ export function parseDockerfile(
     baseImage = argumentsData[0].getValue()
   }
 
-  const resultInstructions: Instructions[] = []
+  const resultInstructions: Instruction[] = []
 
   // Process all other instructions
-  for (const instruction of instructions) {
+  for (const instruction of instructions as any[]) {
```
dobrac: why as any?

packages/js-sdk/src/template/index.ts
```diff
@@ -45,121 +37,24 @@ export type BuildOptions = BasicBuildOptions & {
...
   copy(
     src: string,
     dest: string,
-    options?: { forceUpload?: true; user?: string; mode?: number }
+    options?: { forceUpload?: boolean; user?: string; mode?: number }
   ): TemplateBuilder
   copy(
     items: CopyItem[],
-    options?: { forceUpload?: true; user?: string; mode?: number }
+    options?: { forceUpload?: boolean; user?: string; mode?: number }
```
dobrac: why this change? The idea was to not forget the true there for prod

packages/js-sdk/src/template/index.ts
```diff
@@ -241,44 +138,43 @@ export class TemplateBase
     return this
   }
 
-  runCmd(command: string, options?: { user?: string }): TemplateBuilder
-  runCmd(commands: string[], options?: { user?: string }): TemplateBuilder
   runCmd(
-    commandOrCommands: string | string[],
+    command: string | string[],
```
dobrac: why this change?

packages/js-sdk/src/template/index.ts
```diff
@@ -344,18 +240,183 @@ export class TemplateBase
...
+    this.template.readyCmd = readyCmd
+    return new TemplateFinal(this.template)
+  }
+}
+
+export class TemplateFinal {
+  constructor(private template: TemplateBase) {}
+}
+
+export class TemplateBase implements DockerfileParserInterface {
```
dobrac: Maybe the DockerfileParserInterface should be now common type, and call it differently

dobrac (changes requested):

packages/js-sdk/src/template/index.ts
mishushakov: I have removed `implements DockerfileParserInterface`, it shouldn't be there (same for Python)

packages/js-sdk/src/template/index.ts
```diff
@@ -241,44 +142,45 @@ export class TemplateBase
     return this
   }
 
-  runCmd(command: string, options?: { user?: string }): TemplateBuilder
-  runCmd(commands: string[], options?: { user?: string }): TemplateBuilder
```
dobrac: why this change?

packages/js-sdk/src/template/index.ts
```diff
@@ -504,17 +467,143 @@ export class TemplateBase
...
+      force: this.forceNextLayer,
+    })
+    return this
+  }
+
+  setStartCmd(startCommand: string, readyCommand: string): TemplateFinal {
+    this.startCmd = startCommand
+    this.readyCmd = readyCommand
+    return new TemplateFinal(this)
+  }
```
dobrac: why are these methods on the base?

dobrac (changes requested):

packages/js-sdk/src/template/index.ts
```diff
@@ -504,17 +468,97 @@ export class TemplateBase
...
+    }
+
+    return templateBuilder
+  }
+
+  setStartCmd(startCommand: string, readyCommand: string): TemplateFinal {
+    this.startCmd = startCommand
+    this.readyCmd = readyCommand
+    return new TemplateFinal(this)
+  }
```
dobrac: why is this here?

packages/js-sdk/src/template/index.ts
```diff
@@ -525,70 +569,52 @@ export class TemplateBase
...
-Template.waitForTimeout = function (timeout: number) {
-  // convert to seconds, but ensure minimum of 1 second
-  const seconds = Math.max(1, Math.floor(timeout / 1000))
-  return `sleep ${seconds}`
-}
+Template.waitForPort = TemplateBase.waitForPort
+Template.waitForURL = TemplateBase.waitForURL
+Template.waitForProcess = TemplateBase.waitForProcess
+Template.waitForFile = TemplateBase.waitForFile
+Template.waitForTimeout = TemplateBase.waitForTimeout
```
dobrac: this was better approach

packages/python-sdk/e2b/template/main.py
```diff
@@ -314,8 +320,8 @@ def from_dockerfile(self, dockerfile_content_or_path: str) -> TemplateBuilder:
         return builder
 
     @staticmethod
-    def to_json(template: "TemplateBase") -> str:
-        return template.__to_json()
+    def to_json(template: "TemplateClass") -> str:
+        return template.get_template_base().__to_json()
```
dobrac: why the to JSON instance method?

packages/python-sdk/e2b/template/main.py
```diff
@@ -324,8 +330,8 @@ def __to_json(self) -> str:
         )
 
     @staticmethod
-    def to_dockerfile(template: "TemplateBase") -> str:
-        return template.__to_dockerfile()
+    def to_dockerfile(template: "TemplateClass") -> str:
+        return template.get_template_base().__to_dockerfile()
```
dobrac: why is the to_dockerfile instance method?

packages/python-sdk/e2b/template/main.py
```diff
@@ -14,7 +14,10 @@
 
 class TemplateBuilder:
     def __init__(self, template: "TemplateBase"):
-        self._template = template
+        self.__template = template
+
+    def get_template_base(self) -> "TemplateBase":
+        return self.__template
```
dobrac, line 20: why do you need this?

dobrac (changes requested):

packages/python-sdk/e2b/template/main.py
mishushakov: because I made template private property, I think it's better this way because you're explicitly have to call the function so you see where the template builder is coming from

mishushakov:
outdated

PR #896, opened 2025-09-08
Replace async import() with require() for correctness

mishushakov:


ValentaTomas:
@mishushakov Is this still relevant?

mishushakov:
no it has no effect eitherway

PR #901, opened 2025-09-09
Render failed layer stack traces on build error

mishushakov:
### Changelog

- Added traces for Python and JavaScript SDK.
- Added methods for stack trace collection, stack trace context
- Updated file upload mechanism for tracing
- Changed API handling to support throwing API errors with traces

> **Note**
> For functions like `aptInstall`, `fromRegistry` which call `.run` and `.fromImage`, the stack trace has to be replaced so it throws correctly from parent.

> **Note**
> There are no build logs in the trace, we can add this at later point.

### Examples

Keep in mind that the way the traces are rendered entirely depends on environment (Bun, Node, Deno, Vitest, Pytest)

**Standard Python**

```
Traceback (most recent call last):
  File "/Users/mish/Documents/Projects/E2B/E2B/packages/python-sdk/tests/sync/template_sync/test_stacktrace.py", line 17, in <module>
    Template.build(
    ~~~~~~~~~~~~~~^
        template,
        ^^^^^^^^^
    ...<2 lines>...
        memory_mb=1024,
        ^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/mish/Documents/Projects/E2B/E2B/packages/python-sdk/e2b/template_sync/main.py", line 157, in build
    wait_for_build_finish(
    ~~~~~~~~~~~~~~~~~~~~~^
        api_client,
        ^^^^^^^^^^^
    ...<4 lines>...
        stack_traces=template._template._stack_traces,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/mish/Documents/Projects/E2B/E2B/packages/python-sdk/e2b/template_sync/build_api.py", line 199, in wait_for_build_finish
    raise BuildException(error_message).with_traceback(stack_trace)
  File "/Users/mish/Documents/Projects/E2B/E2B/packages/python-sdk/tests/sync/template_sync/test_stacktrace.py", line 12, in <module>
    .run_cmd("cat non_existing_file.txt")
     ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
e2b.template.exceptions.BuildException: failed to run command: exit status 1
```

**Pytest**

```
[gw0] darwin -- Python 3.13.3 /Users/mish/Library/Caches/pypoetry/virtualenvs/e2b-e6GdLxLZ-py3.13/bin/python

    @pytest.mark.skip_debug()
    def test_build():
        template = (
            Template()
            .from_image("ubuntu:22.04")
            .run_cmd("cat non_existing_file.txt")
            .set_workdir("/app")
            .set_start_cmd("echo 'Hello, world!'", wait_for_timeout(10_000))
        )
    
        # try:
>       Template.build(
            template,
            alias=str(uuid4()),
            cpu_count=1,
            memory_mb=1024,
        )

tests/sync/template_sync/test_stacktrace.py:18: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
e2b/template_sync/main.py:157: in build
    wait_for_build_finish(
e2b/template_sync/build_api.py:199: in wait_for_build_finish
    raise BuildException(error_message).with_traceback(stack_trace)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    @pytest.mark.skip_debug()
    def test_build():
        template = (
            Template()
            .from_image("ubuntu:22.04")
>           .run_cmd("cat non_existing_file.txt")
            .set_workdir("/app")
            .set_start_cmd("echo 'Hello, world!'", wait_for_timeout(10_000))
        )
E       e2b.template.exceptions.BuildException: failed to run command: exit status 1

tests/sync/template_sync/test_stacktrace.py:12: BuildException
========================================================== short test summary info ==========================================================
FAILED tests/sync/template_sync/test_stacktrace.py::test_build - e2b.template.exceptions.BuildException: failed to run command: exit status 1
```

**Node**

```
node:internal/modules/run_main:104
    triggerUncaughtException(
    ^

    at file:///Users/mish/Documents/Projects/E2B/E2B/packages/js-sdk/tests/template/trace.test.mts:8:4
    at ModuleJob.run (node:internal/modules/esm/module_job:274:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:644:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:98:5) {
  name: 'BuildError'
}

Node.js v23.11.0
```

**Bun**

```
1 | export class BuildError extends Error {
2 |   constructor(message: string, stackTrace?: string) {
3 |     super(message)
        ^
BuildError: failed to run command: exit status 1
      at /Users/mish/Documents/Projects/E2B/E2B/packages/js-sdk/src/template/errors.ts:3:5

Bun v1.2.4 (macOS arm64)
```

**Vitest**

```
BuildError: failed to run command: exit status 1
 ❯ tests/template/trace.test.ts:8:6
      6|   const template = Template()
      7|     .fromImage('ubuntu:22.04')
      8|     .runCmd('cat folder/test.txt')
       |      ^
      9|     .setWorkdir('/app')
     10|     .setStartCmd('echo "Hello, world!"', waitForTimeout(10_000))
```

### DIFF
--- a/.changeset/nine-suits-drop.md
+++ b/.changeset/nine-suits-drop.md
@@ -1,6 +1,6 @@
 ---
-'@e2b/python-sdk': minor
-'e2b': minor
+'@e2b/python-sdk': patch
+'e2b': patch
 ---
 
 added step stacktraces


mlejva:
@mishushakov 

> Keep in mind that the way the traces are rendered entirely depends on environment (Bun, Node, Deno, Vitest, Pytest)

How is the experience in these specific environments? Great in all of them or anything we should know about?

packages/js-sdk/src/template/buildApi.ts
```diff
@@ -214,7 +216,19 @@ export async function waitForBuildFinish(
         break
       }
       case 'error': {
-        throw new BuildError(buildStatus?.reason?.message ?? 'Unknown error')
+        let stackError: string | undefined
+        if (buildStatus.reason?.step) {
```
dobrac:
```suggestion
        if (buildStatus.reason?.step != undefined) {
```

packages/js-sdk/src/template/buildApi.ts
```diff
@@ -214,7 +216,19 @@ export async function waitForBuildFinish(
...
+          const step = getBuildStepIndex(
+            buildStatus.reason?.step,
```
dobrac:
```suggestion
            buildStatus.reason.step,
```

packages/js-sdk/src/template/utils.ts
```diff
@@ -98,6 +112,21 @@ export async function tarFileStreamUpload(
...
+  if (step === 'base') {
+    return 0
+  }
+
+  if (step == 'finalize') {
+    return stackTracesLength - 1
+  }
+
+  return Number(step)
+}
```
dobrac, line 147: this looks actually quite ok conversion function imo

packages/js-sdk/src/template/index.ts
```diff
@@ -401,7 +438,9 @@ export class TemplateBase
     if (options?.depth) {
       args.push(`--depth ${options.depth}`)
     }
+
     this.runCmd(args.join(' '))
+    this.replaceLastStackTrace(getCallerFrame())
```
dobrac: is the replace instead of append here (and also a lot of other places) correctly?

packages/python-sdk/e2b/template/utils.py
```diff
@@ -67,17 +68,23 @@ def strip_ansi_escape_codes(text: str) -> str:
     return ansi_escape.sub("", text)
 
 
+def get_caller_frame(depth: int = 3) -> Optional[FrameType]:
```
dobrac: why depth 3 is the default? please add a comment (same for JS)

packages/js-sdk/src/template/index.ts
dobrac: I see, can you please create then some internal-ish run function that will be private and won't include the stacktrace append instead of replacing it like this? It doesn't look transparent imo

mishushakov:
I've added the shim functions and will be adding the tests

mishushakov:
There's one potential blocker when it comes to stack traces for .copy. I will change the default error to throw the extracted trace instead of the default

packages/js-sdk/src/template/types.ts
```diff
@@ -93,9 +93,10 @@ export interface TemplateBuilder {
 
   makeSymlink(src: string, dest: string): TemplateBuilder
 
-  runCmd(command: string, options?: { user?: string }): TemplateBuilder
-
-  runCmd(commands: string[], options?: { user?: string }): TemplateBuilder
+  runCmd(
+    commandOrCommands: string | string[],
+    options?: { user?: string }
+  ): TemplateBuilder
```
dobrac, line 111: why this change?

packages/js-sdk/src/template/utils.ts
```diff
@@ -45,16 +50,44 @@ export async function calculateFilesHash(
...
+/**
+ * Get the caller directory
+ * @returns The caller directory
+ */
+export function getCallerDirectory(): string | undefined {
+  /**
+   * Levels explained:
+   * getCallerDirectory (this function) > TemplateBase.fileContextPath > Template > original caller (eg. template file)
+   */
+  const caller = getCallerFrame(4)
```
dobrac: explain why 4 is the number, also, move it probably to some const

packages/js-sdk/src/template/utils.ts
```diff
@@ -98,6 +131,21 @@ export async function tarFileStreamUpload(
...
 
+export function getBuildStepIndex(
+  step: string,
+  stackTracesLength: number
+): number {
+  if (step === 'base') {
+    return 0
+  }
+
+  if (step == 'finalize') {
```
dobrac:
```suggestion
  if (step === 'finalize') {
```

packages/js-sdk/src/template/index.ts
```diff
@@ -512,47 +559,55 @@ export class TemplateBase
...
-          new LogEntry(new Date(), 'info', `Uploaded '${file.src}'`)
+    const uploadPromises = instructionsWithHashes.map(
+      async (instruction, index) => {
+        if (instruction.type !== 'COPY') {
+          return
+        }
+
+        const src = instruction.args[0]
+        const forceUpload = instruction.forceUpload
+        const filesHash = instruction.filesHash!
```
dobrac: can we make graceful check here?

packages/js-sdk/src/template/index.ts
```diff
@@ -580,32 +635,36 @@ export class TemplateBase
...
-      }
-
-      steps.push(instruction)
-    }
-
-    return steps
+  private async instructionsWithHashes(): Promise<Instruction[]> {
+    return Promise.all(
+      this.instructions.map(async (instruction, index) => {
+        if (instruction.type !== 'COPY') {
```
dobrac: I would suggest moving "COPY" to const somewhere

packages/python-sdk/e2b/template/main.py
```diff
@@ -268,31 +291,74 @@ def __init__(
...
+
+    def _disable_stack_trace(self) -> "TemplateBase":
+        """Disable stack trace collection"""
+        self._stack_traces_enabled = False
+        return self
+
+    def _enable_stack_trace(self) -> "TemplateBase":
+        """Enable stack trace collection"""
+        self._stack_traces_enabled = True
+        return self
```
dobrac, line 335: where is this used? It feels a big hacky to control it this way. Especially when we would like to make the builder stateless

packages/python-sdk/e2b/template/main.py
```diff
@@ -420,6 +497,10 @@ def _calculate_hashes(
             )
 
             if instruction["type"] == "COPY":
```
dobrac: same here with the const 🙏

packages/python-sdk/e2b/template/utils.py
```diff
@@ -67,17 +69,23 @@ def strip_ansi_escape_codes(text: str) -> str:
     return ansi_escape.sub("", text)
 
 
+def get_caller_frame(depth: int = 2) -> Optional[FrameType]:
```
dobrac: explain the default 2 value please (in JS it's 4 if I'm not mistaken)

packages/python-sdk/e2b/template_sync/main.py
```diff
@@ -73,31 +73,30 @@ def build(
...
+            instructions_with_hashes = template._template._instructions_with_hashes()
 
             # Upload files
-            for file_upload in file_uploads:
+            for index, file_upload in enumerate(instructions_with_hashes):
+                if file_upload["type"] != "COPY":
+                    continue
+
+                src = file_upload["args"][0]
+                force_upload = file_upload.get("forceUpload")
```
dobrac, line 85: is this type safe access?

packages/js-sdk/tests/template/stacktrace.test.ts
```diff
@@ -0,0 +1,222 @@
...
+  const template = Template()
+    .fromBaseImage()
+    .copy(nonExistentPath, nonExistentPath)
+  await expectToThrowAndCheckTrace(async () => {
+    await buildTemplate(template)
+  }, 'copy')
+})
+
+test('traces on remove', { timeout: testTimeout }, async () => {
+  const template = Template().fromBaseImage().remove(nonExistentPath)
```
dobrac: this will always pass, you need to make them multiline or split to multiple assignments (applies to most of other tests too)

packages/python-sdk/tests/async/template_async/test_stacktrace.py
```diff
@@ -0,0 +1,173 @@
...
+
+# @pytest.mark.skip_debug()
+# async def test_traces_on_from_template():
+#     template = AsyncTemplate().from_template("this-template-does-not-exist")
+#     await _expect_to_throw_and_check_trace(lambda: build(template), "from_template")
+
+
+@pytest.mark.skip_debug()
+async def test_traces_on_from_dockerfile():
+    template = AsyncTemplate().from_dockerfile("FROM ubuntu:22.04\nRUN nonexistent")
```
dobrac: this will always pass, you need to make them multiline or split to multiple assignments (applies to most of other tests too)

packages/python-sdk/e2b/template/main.py
mishushakov: only for creating new stack trace context to disable stack trace collection of the child function temporarily, that's the only way to get it working without shim functions

dobrac: I though the with traces context will handle it? Maybe just moving this outside of the huge builder class would be enough

packages/js-sdk/src/template/types.ts
mishushakov: no need to overload here, it makes no difference

dobrac: would it be better for linters to override? (to nicely show you have 2 options instead of the ugly union type)

packages/js-sdk/tests/template/stacktrace.test.ts
```diff
@@ -111,28 +110,28 @@ test('traces on fromAWSRegistry', { timeout: testTimeout }, async () => {
...
     await buildTemplate(template)
   }, 'fromGCPRegistry')
 })
 
 test('traces on copy', { timeout: testTimeout }, async () => {
   const template = Template()
-    .fromBaseImage()
-    .copy(nonExistentPath, nonExistentPath)
+  template.fromBaseImage()
+  template.copy(nonExistentPath, nonExistentPath)
```
dobrac: this should be invalid usage (even though it unfortunately works now), because the builder needs to be stateless

packages/python-sdk/e2b/template/main.py
mishushakov: I can remove _disable_stack_trace and _enable_stack_trace, but the run in context function will require self._stack_traces_enabled state to be passed around.
I am not sure how to do it without state, the collect stack traces function explicitly checks if enabled is true

mishushakov: that's the pattern

```js
let skipB = false;

function mywrapepd(f) {
  skipB = true;
  f();
  skipB = false;
}

function test() {
  console.log("doing something");
  b();
}

function b() {
  if (skipB) return;
  console.log("real b called");
}

mywrapepd(test); // skips b
```

packages/js-sdk/tests/template/stacktrace.test.ts
```diff
@@ -0,0 +1,224 @@
+import { assert, test } from 'vitest'
+import { Template, TemplateClass, waitForTimeout } from '../../src'
+import { randomUUID } from 'node:crypto'
+import fs from 'node:fs'
+
+const __fileContent = fs.readFileSync(__filename, 'utf8') // read current file content
+const testTimeout = 180000 // 3 minutes
```
dobrac: I think this can be set somehow more globally for the file/test groups

packages/js-sdk/src/template/index.ts
```diff
@@ -512,47 +560,55 @@ export class TemplateBase
...
-          new LogEntry(new Date(), 'info', `Uploaded '${file.src}'`)
+    const uploadPromises = instructionsWithHashes.map(
+      async (instruction, index) => {
+        if (instruction.type !== InstructionType.COPY) {
+          return
+        }
+
+        const src = instruction.args[0] ?? ''
+        const forceUpload = instruction.forceUpload
+        const filesHash = instruction.filesHash ?? ''
```
dobrac: the default should be error imo, not empty

packages/js-sdk/tests/template/stacktrace.test.ts
```diff
@@ -0,0 +1,224 @@
...
+// test('traces on fromTemplate', { timeout: testTimeout }, async () => {
+//   const template = Template().fromTemplate('this-template-does-not-exist')
+//   await expectToThrowAndCheckTrace(async () => {
+//     await buildTemplate(template)
+//   }, 'fromTemplate')
+// })
+
+test('traces on fromDockerfile', { timeout: testTimeout }, async () => {
+  const template = Template()
+  template.fromDockerfile('FROM ubuntu:22.04\nRUN nonexistent')
```
dobrac: this is still not fixed

packages/python-sdk/e2b/template_sync/main.py
```diff
@@ -73,31 +73,30 @@ def build(
...
-                }
-                for instruction in instructions_with_hashes
-                if instruction["type"] == "COPY"
-            ]
+            instructions_with_hashes = template._template._instructions_with_hashes()
 
             # Upload files
-            for file_upload in file_uploads:
+            for index, file_upload in enumerate(instructions_with_hashes):
+                if file_upload["type"] != "COPY":
```
dobrac: use the const

packages/python-sdk/e2b/template_sync/main.py
dobrac: ?

packages/js-sdk/tests/template/stacktrace.test.ts
mishushakov: good catch, vitest workspaces

packages/python-sdk/e2b/template/main.py
dobrac: it's ok for now, it'll have to be refactored with the stateless builder anyway

dobrac (approved):
lg!

PR #903, opened 2025-09-12
Added format option to CLI sandbox + templates list

mishushakov:
This PR is based on https://github.com/e2b-dev/E2B/pull/902, please proceed with that first

**Changelog**

- Added `--format` option to "e2b sandbox list" and "e2b template list". Can either be "table" or "json"

Sandboxes list in JSON format

Screenshot 2025-09-12 at 11 17 45

Templates list in JSON format

Screenshot 2025-09-12 at 11 18 55

### DIFF
--- a/packages/cli/src/commands/sandbox/list.ts
+++ b/packages/cli/src/commands/sandbox/list.ts
@@ -41,11 +41,6 @@ export const listCommand = new commander.Command('list')
         metadataRaw: options.metadata,
       })
 
-      if (!sandboxes?.length) {
-        console.log('No sandboxes found')
-        return
-      }
-
       if (format === 'table') {
         renderTable(sandboxes, state)
       } else if (format === 'json') {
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/sandbox/list.ts
```diff
@@ -33,76 +43,90 @@ export const listCommand = new commander.Command('list')
 
       if (!sandboxes?.length) {
         console.log('No sandboxes found')
```
jakubno: you should return empty array for `json`

packages/cli/src/commands/template/list.ts
```diff
@@ -29,74 +31,87 @@ export const listCommand = new commander.Command('list')
 
       if (!templates?.length) {
         console.log('No templates found.')
```
jakubno: also here

PR #916, opened 2025-09-25
Correct file hashing logic in the SDK

mishushakov:
- Adds relative path, mode, uid, gid, size, mtimeMs

### DIFF
--- a/packages/js-sdk/src/template/utils.ts
+++ b/packages/js-sdk/src/template/utils.ts
@@ -51,6 +51,14 @@ export async function calculateFilesHash(
 
     const content = fs.readFileSync(file.fullpath())
     hash.update(new Uint8Array(content))
+
+    // Add stat information to hash calculation
+    const stats = fs.statSync(file.fullpath())
+    hash.update(stats.mode.toString())
+    hash.update(stats.uid.toString())
+    hash.update(stats.gid.toString())
+    hash.update(stats.size.toString())
+    hash.update(stats.mtimeMs.toString())
   }
 
   return hash.digest('hex')
--- a/packages/js-sdk/src/template/utils.ts
+++ b/packages/js-sdk/src/template/utils.ts
@@ -52,6 +52,10 @@ export async function calculateFilesHash(
     const content = fs.readFileSync(file.fullpath())
     hash.update(new Uint8Array(content))
 
+    // Add relative path to hash calculation
+    const relativePath = path.relative(contextPath, file.fullpath())
+    hash.update(relativePath)
+
     // Add stat information to hash calculation
     const stats = fs.statSync(file.fullpath())
     hash.update(stats.mode.toString())
--- a/packages/python-sdk/e2b/template/utils.py
+++ b/packages/python-sdk/e2b/template/utils.py
@@ -58,6 +58,14 @@ def calculate_files_hash(
             continue
         with open(file, "rb") as f:
             hash_obj.update(f.read())
+        
+        # Add stat information to hash calculation
+        stat_info = os.stat(file)
+        hash_obj.update(str(stat_info.st_mode).encode())
+        hash_obj.update(str(stat_info.st_uid).encode())
+        hash_obj.update(str(stat_info.st_gid).encode())
+        hash_obj.update(str(stat_info.st_size).encode())
+        hash_obj.update(str(stat_info.st_mtime).encode())
 
     return hash_obj.hexdigest()
 
--- a/packages/python-sdk/e2b/template/utils.py
+++ b/packages/python-sdk/e2b/template/utils.py
@@ -59,6 +59,10 @@ def calculate_files_hash(
         with open(file, "rb") as f:
             hash_obj.update(f.read())
         
+        # Add relative path to hash calculation
+        relative_path = os.path.relpath(file, context_path)
+        hash_obj.update(relative_path.encode())
+        
         # Add stat information to hash calculation
         stat_info = os.stat(file)
         hash_obj.update(str(stat_info.st_mode).encode())


packages/js-sdk/src/template/utils.ts
```diff
@@ -45,12 +45,36 @@ export async function calculateFilesHash(
...
-    if (!file.isFile()) {
+    if (file.isDirectory()) {
       continue
     }
 
-    const content = fs.readFileSync(file.fullpath())
-    hash.update(new Uint8Array(content))
+    // Add relative path to hash calculation
+    const relativePath = path.relative(contextPath, file.fullpath())
+    hash.update(relativePath)
```
dobrac, line 60: this should be calculated for directories too

packages/js-sdk/src/template/utils.ts
```diff
@@ -105,34 +129,48 @@ export function padOctal(mode: number): string {
...
   const { globSync } = await dynamicGlob()
   const { create } = await dynamicTar()
-  const files = globSync(fileName, { cwd: fileContextPath, nodir: false })
+  const files = globSync(fileName, { cwd: fileContextPath })
 
   return create(
     {
       gzip: true,
       cwd: fileContextPath,
+      follow: resolveSymlinks,
```
dobrac, line 156: this is missing for hashes calculation

packages/python-sdk/e2b/template/utils.py
```diff
@@ -53,11 +53,33 @@ def calculate_files_hash(
...
-        if not os.path.isfile(file):
+        if os.path.isdir(file):
             # skip folders
             continue
-        with open(file, "rb") as f:
-            hash_obj.update(f.read())
+
+        # Add relative path to hash calculation
+        relative_path = os.path.relpath(file, context_path)
+        hash_obj.update(relative_path.encode())
```
dobrac, line 67: should be also for directories

packages/python-sdk/e2b/template_async/build_api.py
```diff
@@ -85,11 +85,14 @@ async def upload_file(
...
     url: str,
     stack_trace: Optional[TracebackType] = None,
+    resolve_symlinks: bool = False,
 ):
     tar_buffer = io.BytesIO()
 
     try:
-        with tarfile.open(fileobj=tar_buffer, mode="w:gz") as tar:
+        with tarfile.open(
+            fileobj=tar_buffer, mode="w:gz", dereference=resolve_symlinks
```
dobrac: is this correctly done also in the hash computation?

packages/python-sdk/e2b/template/utils.py
```diff
@@ -53,11 +53,33 @@ def calculate_files_hash(
...
+
+        # Add stat information to hash calculation
+        if os.path.islink(file):
```
dobrac, line 70: you might want different behavior with resolve_symlinks true vs false. Sometimes you might want to track the file changes the symlink points to, not the symlink itself (maybe we should track both at that point?)

packages/python-sdk/e2b/template_sync/main.py
```diff
@@ -103,6 +104,8 @@ def build(
                         src,
                         template._template._file_context_path,
                         file_info.url,
+                        resolve_symlinks,
+                        stack_trace,
```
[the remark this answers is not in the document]
dobrac: this comment is correct, it's incorrectly passed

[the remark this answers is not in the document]
mishushakov: @dobrac are you sure? I think that thing is hallucinating

packages/js-sdk/src/template/index.ts
```diff
@@ -269,6 +282,7 @@ export class TemplateBase
         args,
         force: item.forceUpload ?? this.forceNextLayer,
         forceUpload: item.forceUpload,
+        resolveSymlinks: item.resolveSymlinks ?? false,
```
dobrac: can you make the default a const? (same for other usages)

packages/js-sdk/src/template/utils.ts
```diff
@@ -105,34 +126,48 @@ export function padOctal(mode: number): string {
   return mode.toString(8).padStart(4, '0')
 }
 
-export async function tarFileStream(fileName: string, fileContextPath: string) {
+export async function tarFileStream(
+  fileName: string,
+  fileContextPath: string,
+  resolveSymlinks: boolean = false
```
dobrac: this can also reuse the const default

packages/python-sdk/e2b/template_async/build_api.py
```diff
@@ -84,12 +84,15 @@ async def upload_file(
     file_name: str,
     context_path: str,
     url: str,
+    resolve_symlinks: bool = False,
```
dobrac: move the default to const

mishushakov:
@dobrac I have changed signature to either boolean or undefined

packages/python-sdk/e2b/template_async/build_api.py
```diff
@@ -84,7 +85,7 @@ async def upload_file(
     file_name: str,
     context_path: str,
     url: str,
-    resolve_symlinks: Optional[bool] = None,
+    resolve_symlinks: Optional[bool] = RESOLVE_SYMLINKS,
```
dobrac: does it need to be optional?

packages/python-sdk/e2b/template_sync/build_api.py
```diff
@@ -84,7 +85,7 @@ def upload_file(
     file_name: str,
     context_path: str,
     url: str,
-    resolve_symlinks: Optional[bool] = None,
+    resolve_symlinks: Optional[bool] = RESOLVE_SYMLINKS,
```
dobrac: does it need to be optional?

packages/python-sdk/e2b/template/main.py
```diff
@@ -65,6 +67,7 @@ def copy(
                 args=args,
                 force=force_upload or self._template._force_next_layer,
                 forceUpload=force_upload,
+                resolveSymlinks=resolve_symlinks,
```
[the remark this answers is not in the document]
dobrac: probably ok for now

packages/js-sdk/src/template/utils.ts
```diff
@@ -44,13 +45,47 @@ export async function calculateFilesHash(
...
-    const content = fs.readFileSync(file.fullpath())
-    hash.update(new Uint8Array(content))
+    const stats = fs.statSync(file.fullpath())
+
+    hashStats(stats)
+
+    if (stats.isFile()) {
+      const content = fs.readFileSync(file.fullpath())
+      hash.update(new Uint8Array(content))
+    }
```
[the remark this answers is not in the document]
dobrac: thats ok

packages/python-sdk/e2b/template/main.py
```diff
@@ -538,8 +544,28 @@ def _instructions_with_hashes(
...
+
+        for index, instruction in enumerate(steps):
+            step = {
+                # Serialize enum to string value
+                "type": instruction["type"].value,
+                "args": instruction["args"],
+                "force": instruction["force"],
+                "filesHash": instruction.get("filesHash"),
+                "forceUpload": instruction.get("forceUpload"),
+            }
```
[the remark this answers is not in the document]
dobrac: it's used only locally, a bit misused type

packages/python-sdk/e2b/template/main.py
```diff
@@ -538,8 +544,28 @@ def _instructions_with_hashes(
...
+
+            if step["filesHash"] is None:
+                del step["filesHash"]
+
+            if step.get("forceUpload") is None:
+                del step["forceUpload"]
+
+            _steps.append(step)
```
mishushakov, line 563: I don't like the approach of deleting, please do it like the code below (if instruction[forceUpload] is not None, then step[forceUpload] = instruction[forceUpload])

packages/python-sdk/e2b/template_async/main.py
```diff
@@ -15,6 +15,7 @@
     upload_file,
     wait_for_build_finish,
 )
+from ..template.consts import RESOLVE_SYMLINKS
```
mishushakov: from e2b.template.consts (we follow this pattern)

packages/python-sdk/e2b/template/utils.py
```diff
@@ -52,12 +54,38 @@ def calculate_files_hash(
...
+                content = os.readlink(file)
+                hash_obj.update(content.encode())
+                continue
+
+        stats = os.stat(file)
+        hash_stats(stats)
+
+        if stat.S_ISREG(stats.st_mode):
+            with open(file, "rb") as f:
+                hash_obj.update(f.read())
```
mishushakov, line 88: what does this condition check?

dobrac: this checks if the stats represents a file so its contents can be read (e.g. it's not a folder)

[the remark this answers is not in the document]
dobrac: "fails to follow broken symlinks" :D

mishushakov: you can just do os.path.isfile(file), it should follow symlinks

dobrac: it's unnecessary call, also, in this place symlinks are already followed

dobrac (approved):

PR #919, opened 2025-09-26
Set stdin to `/dev/null` by default

jakubno:
Disable stdin (setting it to `/dev/null` for `command.run()`, but enable to setting it to pipe, which you can send the input via `command.sendStdin()`
This fixes issues with tools checking for stdin and then hanging indefinitely

---

> [!NOTE]
> Disables stdin by default and introduces a new `StartRequest.stdin` flag to opt-in to stdin piping, updating JS/Python SDKs, version gating, and tests.
> 
> - **Spec/Protobuf**:
>   - Add `optional bool stdin` to `process.StartRequest` in `spec/envd/process/process.proto` and regenerate JS/Python protos.
> - **JS SDK**:
>   - Add `ENVD_COMMANDS_STDIN = '0.3.0'` and version check; error if `stdin === false` on older envd.
>   - Extend `CommandStartOpts` with `stdin` (default `false`); pass `stdin` to `rpc.start`.
> - **Python SDK**:
>   - Add `ENVD_COMMANDS_STDIN = Version("0.3.0")` and version check mirroring JS.
>   - Extend async/sync `commands.run()` with `stdin` (default `False`); pass to `StartRequest`.
>   - Update generated `process_pb2`/`.pyi` with `StartRequest.stdin`.
> - **Tests**:
>   - Update send-stdin tests to run commands with `stdin: true`/`stdin=True`.
> - **Changeset**:
>   - Minor bump for `@e2b/python-sdk` and `e2b`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 458a15187e6eadccd43540e6b8972c4d45e1bb78. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/python-sdk/e2b/sandbox_async/commands/command.py
```diff
@@ -201,16 +206,28 @@ async def run(
         cwd: Optional[str] = None,
         on_stdout: Optional[OutputHandler[Stdout]] = None,
         on_stderr: Optional[OutputHandler[Stderr]] = None,
+        stdin: Optional[bool] = None,
```
jakubno, line 209: Not 100% sure about this signiture? Should this be `False`? But that we can't check if the usage is incorrect

ValentaTomas:
> > @jakubno The API throws/fails if we pass an unknown argument here?
> 
> What's unknown input?

Unknown as in a field that was not expected.

jakubno:
> > > @jakubno The API throws/fails if we pass an unknown argument here?
> > 
> > 
> > What's unknown input?
> 
> Unknown as in a field that was not expected.

No, envd just ignores it

packages/js-sdk/src/sandbox/commands/index.ts
```diff
@@ -364,6 +371,15 @@ export class Commands {
         }, requestTimeoutMs)
       : undefined
 
+    if (
+      opts?.stdin === false &&
+      compareVersions(this.envdVersion, ENVD_COMMANDS_STDIN) < 0
+    ) {
+      throw new Error(
```
dobrac: we raise sandboxexception in Python (not 100% if we don't wrap it somewhere in JS along the path, not if we don't want to change it in the Python implementation)

dobrac (commented):
one nit

PR #920, opened 2025-09-26
Improve envd version handling

jakubno:
The envd is always present, so we can simplify the handling + pass it to filesystem module as a preparation for stdin


ValentaTomas (approved):
Lgtm, but just to check—why was the version optional before?

jakubno:
> Lgtm, but just to check—why was the version optional before?

I think this wasn't there before / it was optional

PR #926, opened 2025-10-02
feat: add build v2 command that uses template SDK

dobrac:
Add template `build-v2` command that uses template SDK

---

> [!NOTE]
> Introduce `template build-v2` command that builds templates directly from a Dockerfile via the Template SDK and wire it into the CLI.
> 
> - **CLI - Template**:
>   - **New Command**: `template build-v2` (`packages/cli/src/commands/template/build-v2.ts`)
>     - Builds from Dockerfile using SDK: `Template(...).fromDockerfile(...)` and `Template.build(...)`.
>     - Options: `--dockerfile`, `--cmd`, `--ready-cmd`, `--cpu-count`, `--memory-mb`, `--no-cache`, path via `--path`; alias `bd-v2`.
>     - Validations: template name format, even `memory-mb`.
>     - Auth/config: uses access token, API key, domain; streams build logs; prints success examples.
>   - **Wiring**: Add `buildV2Command` to `packages/cli/src/commands/template/index.ts`.
> - **Release**: Changeset adds patch for `@e2b/cli`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 33578b56dfbcd76c3bb2ecf7fdf6f2cb80c3d0dd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/cli/src/commands/template/build-v2.ts
+++ b/packages/cli/src/commands/template/build-v2.ts
@@ -21,7 +21,7 @@ import { getDockerfile } from './build'
 
 export const buildV2Command = new commander.Command('build-v2')
   .description(
-    'build sandbox template as E2B template. This command reads a Dockerfile and builds it directly.'
+    'build Dockerfile as a Sandbox template. This command reads a Dockerfile and builds it directly.'
   )
   .argument(
     '<template-name>',


packages/cli/src/commands/template/build-v2.ts
```diff
@@ -0,0 +1,211 @@
...
+import { getRoot } from 'src/utils/filesystem'
+import {
+  asFormattedSandboxTemplate,
+  asLocal,
+  asLocalRelative,
+  asPrimary,
+  asPython,
+  asTypescript,
+  withDelimiter,
+} from 'src/utils/format'
```
mishushakov: use relative path please

packages/cli/src/commands/template/build-v2.ts
```diff
@@ -0,0 +1,211 @@
...
+import { getDockerfile } from './build'
+
+export const buildV2Command = new commander.Command('build-v2')
+  .description(
+    'build sandbox template using the E2B Template SDK. This command reads a Dockerfile and builds it directly.'
```
mishushakov: maybe "as E2B Template", there's no such thing as E2B Template SDK

packages/cli/src/commands/template/build-v2.ts
```diff
@@ -0,0 +1,211 @@
...
+    '<template-name>',
+    'template name to create or rebuild. The template name must be lowercase and contain only letters, numbers, dashes and underscores.'
+  )
+  .addOption(pathOption)
+  .option(
+    '-d, --dockerfile <file>',
+    `specify path to Dockerfile. By default E2B tries to find ${asLocal(
+      defaultDockerfileName
+    )} or ${asLocal(fallbackDockerfileName)} in root directory.`
+  )
```
mishushakov, line 36: this should not be option but positional argument

packages/cli/src/commands/template/build-v2.ts
```diff
@@ -0,0 +1,211 @@
...
+    'specify command that will be executed when the sandbox is started.'
+  )
+  .option(
+    '--ready-cmd <ready-command>',
+    'specify command that will need to exit 0 for the template to be ready.'
+  )
+  .option(
+    '--cpu-count <cpu-count>',
+    'specify the number of CPUs that will be used to run the sandbox. The default value is 2.',
+    parseInt
```
mishushakov, line 48: make sure it fails gracefully

packages/cli/src/commands/template/build-v2.ts
```diff
@@ -0,0 +1,211 @@
...
+  )
+  .option(
+    '--memory-mb <memory-mb>',
+    'specify the amount of memory in megabytes that will be used to run the sandbox. Must be an even number. The default value is 512.',
+    parseInt
```
mishushakov, line 53: same as above

packages/cli/src/commands/template/build-v2.ts
```diff
@@ -0,0 +1,211 @@
...
+        // Build the template using SDK
+        try {
+          await Template.build(finalTemplate, {
+            alias: templateName,
+            cpuCount: cpuCount,
+            memoryMB: memoryMB,
+            skipCache: opts.noCache,
+            apiKey: apiKey,
+            domain: domain,
+            onBuildLogs: (logEntry) => console.log(logEntry.toString()),
```
mishushakov, line 147: we should use default logger here

packages/cli/src/commands/template/build-v2.ts
```diff
@@ -0,0 +1,211 @@
...
+          })
+        } catch (error) {
+          console.error('\n❌ Template build failed.')
+          if (error instanceof Error) {
```
mishushakov, line 151: why this check?

mishushakov (changes requested):
requesting changes

packages/cli/src/commands/template/build-v2.ts
dobrac: it's not required, why positional then?

packages/cli/src/commands/template/build-v2.ts
dobrac: it's the same handling as everywhere else we use it (it's copied)

packages/cli/src/commands/template/build-v2.ts
dobrac: it doesn't exist yet 😢

packages/cli/src/commands/template/build-v2.ts
dobrac: just to typecheck for the error.message

packages/cli/src/commands/template/build-v2.ts
dobrac: this is probably not an issue, we have it like this in many places. If it's an issue, we should enforce it package wide

packages/cli/src/commands/template/build-v2.ts
mishushakov: did you try `catch (error: Error) {}`? or was it still complaining?

packages/cli/src/commands/template/build-v2.ts
mishushakov: okay but please note to change after the merge

packages/cli/src/commands/template/build-v2.ts
```diff
@@ -0,0 +1,211 @@
...
+  asPrimary,
+  asPython,
+  asTypescript,
+  withDelimiter,
+} from '../../utils/format'
+import { getDockerfile } from './build'
+
+export const buildV2Command = new commander.Command('build-v2')
+  .description(
+    'build sandbox template as E2B template. This command reads a Dockerfile and builds it directly.'
```
mishushakov: build Dockerfile as a Sandbox template

PR #935, opened 2025-10-08
Update node version in SDK and CLI

ValentaTomas:


ValentaTomas:
Includes type fix commit from https://github.com/e2b-dev/E2B/pull/781

ValentaTomas:
Includes type fix commit from https://github.com/e2b-dev/E2B/pull/781

PR #944, opened 2025-10-13
feat: keep user and workdir from the template in the sandbox

dobrac:
There is a DX issue with the difference of the default user and working directory between templates and sandboxes (especially visible in the Build System 2.0). The default user and workdir for template is root (/root), but for the sandbox is "user" (/home/user).  This creates an unfortunate hidden difference between the build and sandbox run.

**Current state**

```ts
const template = Template()
  .fromBaseImage()
  .runCmd("whoami") // root
  .runCmd("pwd") // /root

const sbx = await Sandbox.create()
sbx.commands.run("whoami") // user
sbx.commands.run("pwd") // /home/user
```

To solve this, we could implement the same logic as Docker uses. The user and workdir would persist from the template to the sandbox. This would mean, that the default user used in the sandbox would be inherited from the last set user in the template. If you don't adjust the user in the template, it will be "root" also in the sandbox. If you adjust it, it can be anything (also, it doesn't have to be "user" at that point).

_Example with root_

```ts
const template = Template()
  .fromBaseImage()
  .runCmd("whoami") // root
  .runCmd("pwd") // /root

const sbx = await Sandbox.create()
sbx.commands.run("whoami") // root
sbx.commands.run("pwd") // //root
```

_Example with "user"_

```ts
const template = Template({
  fileContextPath: "myfolder",
})
  .fromBaseImage()
  .setUser("root")
  .runCmd("whoami") // root
  .runCmd("pwd") // /root
  .setUser("user")
  .runCmd("whoami") // user
  .runCmd("pwd") // /home/user

const sbx = await Sandbox.create()
sbx.commands.run("whoami") // user
sbx.commands.run("pwd") // /home/user
```

Related infrastructure changes: https://github.com/e2b-dev/infra/pull/1349

---

> [!NOTE]
> Enable inheriting default user/workdir from templates by allowing omitted `username`, updating SDK auth/signature logic and envd APIs/spec accordingly.
> 
> - **API/Spec (envd)**:
>   - Make `username` optional in `/files` (download/upload) and expand `Metrics` fields.
>   - Add `/init` fields: `defaultUser`, `defaultWorkdir`, `hyperloopIP`, `timestamp`.
> - **JS SDK**:
>   - Generalize `Username` to `string`; introduce version-gated default via `ENVD_DEFAULT_USER`.
>   - Update `authenticationHeader` to accept `envdVersion`, return empty header when user omitted; propagate through `commands`, `pty`, `filesystem`.
>   - File URL/signature: stop always attaching `username`; allow undefined user (with version fallback); adjust signing to handle empty user.
> - **Python SDK**:
>   - Mirror JS changes: `Username = str`, version-aware defaults, optional user in file ops/URLs, updated `authentication_header` and signature handling; thread `envd_version` into commands/pty/filesystem.
> - **Misc**:
>   - Regenerate OpenAPI client schemas; add `ENVD_DEFAULT_USER`; minor formatting/ignore updates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5c1e0be05c91774c661315fefe1ecd621c2cf0e0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -247,11 +250,19 @@ export class Filesystem {
...
+    ) {
+      user = defaultUsername
+    }
+
     const res = await this.envdApi.api.GET('/files', {
       params: {
         query: {
           path,
-          username: opts?.user || defaultUsername,
+          username: user,
```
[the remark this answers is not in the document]
dobrac: @cursor verify if it's true and it's actually an issue or not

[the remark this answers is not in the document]
dobrac: so resolved then 👍

packages/js-sdk/src/connectionConfig.ts
```diff
@@ -109,6 +109,6 @@ export class ConnectionConfig {
 /**
  * User used for the operation in the sandbox.
  */
-export type Username = 'root' | 'user'
 
 export const defaultUsername: Username = 'user'
+export type Username = 'root' | 'user' | string
```
mishushakov: this will always be string

packages/js-sdk/src/sandbox/commands/pty.ts
```diff
@@ -109,7 +114,7 @@ export class Pty {
       },
       {
         headers: {
-          ...authenticationHeader(opts?.user),
+          ...authenticationHeader(this.envdVersion, opts?.user),
```
mishushakov, line 117: I think it maybe confusing there's user and opts?.user

maybe user should be sandboxUser?

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -613,7 +622,7 @@ export class Sandbox extends SandboxApi {
       const sig = await getSignature({
         path: filePath,
         operation: 'write',
-        user: username,
+        user: username ?? '',
```
mishushakov: why empty string?

dobrac: the user is required for the signature, but if no username is specified, no user can be filled there. Empty is just a filler for it. Same change is already on the infra

packages/js-sdk/src/sandbox/commands/pty.ts
dobrac: what do you mean here?

packages/js-sdk/src/sandbox/index.ts
dobrac: this is how it's constructed then:

```
 if (signatureExpiration === null) {
    signatureRaw = `${path}:${operation}:${user}:${envdAccessToken}`
  } else {
    signatureRaw = `${path}:${operation}:${user}:${envdAccessToken}:${signatureExpiration.toString()}`
  }
```

packages/js-sdk/src/sandbox/commands/pty.ts
mishushakov: nevermind

packages/js-sdk/src/sandbox/index.ts
mishushakov: maybe this should be handled in the getSignature function not here

dobrac: ok, I'll move it then

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -716,10 +732,12 @@ export class Sandbox extends SandboxApi {
...
   }
 
-  private fileUrl(path?: string, username?: string) {
+  private fileUrl(path: string | undefined, username: string | undefined) {
     const url = new URL('/files', this.envdApiUrl)
 
-    url.searchParams.set('username', username ?? defaultUsername)
+    if (username) {
+      url.searchParams.set('username', username)
+    }
```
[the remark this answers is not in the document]
dobrac: this is ok

packages/js-sdk/src/sandbox/commands/index.ts
```diff
@@ -49,7 +49,7 @@ export interface CommandStartOpts extends CommandRequestOpts {
   /**
    * User to run the command as.
    *
-   * @default `user`
+   * @default `last used user in the template`
```
mishushakov: that might be bit unclear, especially for people not using custom templates

dobrac: "default user for the template" or "template default user"

how about this?

mishushakov: default Sandbox user

mishushakov: (as specified in the template)

dobrac: updated to: `default Sandbox user (as specified in the template)`

PR #950, opened 2025-10-14
Add auth to MCP

Code42Cate:
Changed to use commands instead of API and added auth. Will probably change  how the "i am ready" `onStdout: () => resolve(),` works in the next PR.

Needs to be sudo because of Docker.

---

> [!NOTE]
> Introduce MCP auth by launching mcp-gateway with an access token and exposing a token getter; replace HTTP config with command execution and bump default MCP template to v0-1.
> 
> - **SDKs**:
>   - **MCP auth**:
>     - Launch `mcp-gateway` via `commands.run` with `GATEWAY_ACCESS_TOKEN` set to a generated token (JS/TS and Python, async/sync).
>     - Expose token retrieval: `betaGetMcpToken()` (JS) and `beta_get_mcp_token()` (Python) reading `/etc/mcp-gateway/.token` if needed.
>   - **Configuration flow**:
>     - Replace HTTP `POST /config` setup with command-based startup; remove retry logic (JS and Python).
>   - **Defaults**:
>     - Update `defaultMcpTemplate`/`default_mcp_template` to `mcp-gateway-v0-1` (JS and Python).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4456a1822d87fa32110f6659c3bf9135bff71045. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/python-sdk/e2b/sandbox_sync/main.py
```diff
@@ -551,42 +553,16 @@ def beta_create(
...
-                transport=sandbox._transport,
-                headers=sandbox.connection_config.sandbox_headers,
+            token = str(uuid.uuid4())
+            sandbox._set_mcp_token(token)
+
+            sandbox.commands.run(
+                f"sudo -E mcp-gateway --config '{json.dumps(mcp)}'",
+                background=True,
+                envs={"TOKEN": token},
+                timeout=0,
```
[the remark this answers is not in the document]
Code42Cate: probably ok if users want to inject commands into their own sandbox:D

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -333,42 +333,18 @@ export class Sandbox extends SandboxApi {
...
-            mcpConfigured = true
-
-            break
-          }
-        } catch (e) {
-          config.logger?.warn?.(`Failed to configure MCP server: ${e}`)
+      sandbox.mcpToken = crypto.randomUUID()
+
+      await sandbox.commands.run(
+        `sudo -E mcp-gateway --config '${JSON.stringify(sandboxOpts?.mcp)}'`,
```
mishushakov: btw. instead of sudo you can just pass { user: 'root' }

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -333,42 +333,18 @@ export class Sandbox extends SandboxApi {
...
+        {
+          envs: {
+            TOKEN: sandbox.mcpToken ?? '',
```
mishushakov: maybe better naming here GATEWAY_ACCESS_TOKEN

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -333,42 +333,18 @@ export class Sandbox extends SandboxApi {
...
-          })
-
-          if (res.ok) {
-            mcpConfigured = true
-
-            break
-          }
-        } catch (e) {
-          config.logger?.warn?.(`Failed to configure MCP server: ${e}`)
+      sandbox.mcpToken = crypto.randomUUID()
```
mishushakov, line 336: I am thinking whether you can't just re-use envd access token?

mishushakov: also if the client disconnects, how do I retrieve back the token after Sandbox.connect()

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -582,6 +558,17 @@ export class Sandbox extends SandboxApi {
...
 
+  /**
+   * @beta This feature is in beta and may change in the future.
+   *
+   * Get the MCP token for the sandbox.
+   *
+   * @returns MCP token for the sandbox, or undefined if MCP is not enabled.
+   */
+  betaGetMcpToken(): string | undefined {
+    return this.mcpToken
```
mishushakov, line 577: see above, you won't be able to retrieve this again after losing client

packages/js-sdk/src/sandbox/index.ts
Code42Cate: root and sudo are not the same, this was (for now) deliberate. Might change later

packages/js-sdk/src/sandbox/index.ts
Code42Cate: I think it makes sense that its 2 separate tokens to have some amount of least privilege. If you need it longer, save it yourself?

packages/js-sdk/src/sandbox/index.ts
Code42Cate: I think thats fine

packages/js-sdk/src/sandbox/index.ts
Code42Cate: good point, will fix

packages/js-sdk/src/sandbox/index.ts
mishushakov: we can't ask user such things

mishushakov: I think you'll have to return it from the sandbox (see getInfo endpoint), no other way around

Code42Cate: ok fair enough, this will require API changes first. Can we merge this and I follow up with another PR when I got the API deployed with the needed changes?

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -582,6 +559,23 @@ export class Sandbox extends SandboxApi {
...
+   * @returns MCP token for the sandbox, or undefined if MCP is not enabled.
+   */
+  async betaGetMcpToken(): Promise<string | undefined> {
+    if (!this.mcpToken) {
+      this.mcpToken = await this.files.read('/etc/mcp-gateway/.token', {
+        user: 'root',
+      })
+    }
+
+    return this.mcpToken
```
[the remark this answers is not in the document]
ValentaTomas: @Code42Cate This might be valid.

packages/python-sdk/e2b/sandbox_async/main.py
```diff
@@ -634,6 +609,20 @@ async def beta_pause(
...
+        Get the MCP token for the sandbox.
+
+        :return: MCP token for the sandbox, or None if MCP is not enabled.
+        """
+        if not self._mcp_token:
+            self._mcp_token = await self.files.read(
+                "/etc/mcp-gateway/.token", user="root"
+            )
+        return self._mcp_token
+
```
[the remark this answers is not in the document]
Code42Cate: on sbx create we write it into the property first, so if you get the token right after there should not be a race condition

PR #994, opened 2025-10-28
Expand folder with double star pattern for recursive hash computation

mishushakov:
Currently, when you provide a simple folder pattern like `folder/` to copy() it will not check folders contents for changes (only the folder metadata itself) which leads to stale caches.

This fix appends ** pattern when no pattern is specified for recursive hash computation.

---

> [!NOTE]
> Ensure folder COPY patterns hash contents recursively to fix cache invalidation, adding a shared file-gathering helper, tests, and a Python glob dependency.
> 
> - **Template hashing (JS + Python SDKs)**:
>   - Add `getAllFilesForFilesHash` (`packages/js-sdk/src/template/utils.ts`) and `get_all_files_for_files_hash` (`packages/python-sdk/e2b/template/utils.py`) to collect directories and nested files using glob patterns and ignores.
>   - Update `calculateFilesHash`/`calculate_files_hash` to use the new helpers, hashing recursive paths, metadata, and file contents; handle symlinks consistently.
> - **Tests**:
>   - Add comprehensive tests for recursive matching, ignore patterns, sorting, empty dirs, symlinks in `packages/js-sdk/tests/.../getAllFilesForFilesHash.test.ts` and Python async/sync tests under `packages/python-sdk/tests/.../test_get_all_files_for_files_hash.py`.
> - **Dependencies**:
>   - Add `wcmatch` in `packages/python-sdk/pyproject.toml` for advanced globbing.
> - **Release**:
>   - Patch bump via `.changeset/green-mice-watch.md`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d495427d86ad4e4d5a280567baa63df40a3271fb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/python-sdk/e2b/template/utils.py
+++ b/packages/python-sdk/e2b/template/utils.py
@@ -56,8 +56,8 @@ def get_all_files_for_files_hash(
     )
 
     for file in files_glob:
-        # Join it with context_path to get the absolute path
-        file_path = os.path.join(context_path, file)
+        # Join it with abs_context_path to get the absolute path
+        file_path = os.path.join(abs_context_path, file)
 
         if os.path.isdir(file_path):
             # If it's a directory, add the directory and all entries recursively
@@ -69,7 +69,7 @@ def get_all_files_for_files_hash(
                 exclude=ignore_patterns,
             )
             for dir_file in dir_files:
-                dir_file_path = os.path.join(context_path, dir_file)
+                dir_file_path = os.path.join(abs_context_path, dir_file)
                 files.add(dir_file_path)
         else:
             files.add(file_path)


packages/js-sdk/src/template/utils.ts
```diff
@@ -45,12 +45,20 @@ export async function calculateFilesHash(
...
-  const srcPath = path.join(contextPath, src)
+  let srcPath = path.join(contextPath, src)
   const hash = crypto.createHash('sha256')
   const content = `COPY ${src} ${dest}`
 
   hash.update(content)
 
+  // check if the srcPath is a directory appending ** if no pattern is used
+  if (
+    !src.includes('*') &&
```
dobrac: this will fail for: myfolder/**/anotherfolder

dobrac (changes requested):

packages/js-sdk/src/template/utils.ts
```diff
@@ -23,6 +24,53 @@ export function readDockerignore(contextPath: string): string[] {
...
+  const files = new Set<Path>()
+
+  const globFiles = await glob(srcPath, {
+    ignore: ignorePatterns,
+    withFileTypes: true,
+  })
+
+  for (const file of globFiles) {
+    if (file.isDirectory()) {
+      const dirFiles = await glob(path.join(srcPath, file.fullpath(), '**/*'), {
```
dobrac: is the srcPath required? Isn't there an absolute/fullpath on the file itself?

packages/js-sdk/src/template/utils.ts
```diff
@@ -23,6 +24,53 @@ export function readDockerignore(contextPath: string): string[] {
...
+        nodir: true,
```
dobrac: I guess we want also dirs here to do proper metadata hashes

packages/js-sdk/src/template/utils.ts
```diff
@@ -23,6 +24,53 @@ export function readDockerignore(contextPath: string): string[] {
...
+        ignore: ignorePatterns,
+        withFileTypes: true,
+      })
+      dirFiles.forEach((f) => files.add(f))
+      continue
+    }
+
+    files.add(file)
```
dobrac: do we not want to add also the directories?

packages/js-sdk/src/template/utils.ts
```diff
@@ -97,11 +134,7 @@ export async function calculateFilesHash(
       }
     }
 
-    const stats = fs.statSync(file.fullpath())
-
-    hashStats(stats)
-
-    if (stats.isFile()) {
+    if (stats?.isFile()) {
```
dobrac: why the move of stats to the top?

packages/js-sdk/src/template/utils.ts
```diff
@@ -23,6 +24,53 @@ export function readDockerignore(contextPath: string): string[] {
...
+/**
+ * Hash the stats of a file or directory.
+ *
+ * @param hash Hash object to update
+ * @param stats File or directory stats
+ */
+function hashStats(hash: crypto.Hash, stats: fs.Stats | undefined): void {
+  if (!stats) {
+    return
+  }
```
dobrac: it's better to handle these cases upstream

packages/js-sdk/src/template/utils.ts
mishushakov: because otherwise we'd have to do it twice - for files and for the symbolic link

dobrac: it wasn't called twice before, there was "continue"

dobrac: I agree that it's not the best flow, but I don't think this one is any better (especially if you move the stats undefined check up the stack, which should be imo done)

dobrac: Oh I see, the `const stats = fs.statSync(file.fullpath(), { throwIfNoEntry: false })` call itself was called twice, but not the hashStats function

packages/js-sdk/src/template/utils.ts
```diff
@@ -44,51 +81,52 @@ export async function calculateFilesHash(
...
+      error.stack = stackTrace
+    }
+    throw error
+  }
+
+  // Process files recursively
   for (const file of files) {
-    // Add a relative path to hash calculation
+    const stats = fs.statSync(file.fullpath(), { throwIfNoEntry: false })
+    hashStats(hash, stats)
```
dobrac: this will incorrectly add to the hash the symlink info when resolveSymlinks is enabled

packages/js-sdk/src/template/utils.ts
dobrac: this is not resolved, I've meant the undefined stats variable value and early return, rather than the function placement (that was completely ok)

packages/js-sdk/src/template/utils.ts
```diff
@@ -23,6 +24,47 @@ export function readDockerignore(contextPath: string): string[] {
...
+  contextPath: string,
+  ignorePatterns: string[]
+) {
+  const { glob } = await dynamicGlob()
+  const files = new Set<Path>()
+
+  const globFiles = await glob(src, {
+    ignore: ignorePatterns,
+    withFileTypes: true,
+    cwd: contextPath,
```
mishushakov, line 47: this is required so that the ignore pattern is relative to the file path

packages/python-sdk/e2b/template/utils.py
```diff
@@ -34,6 +34,45 @@ def read_dockerignore(context_path: str) -> List[str]:
...
+    :param context_path: Base directory for resolving relative paths
+    :param ignore_patterns: Ignore patterns
+    :return: Array of files
+    """
+    files = set()
+
+    spec = PathSpec.from_lines("gitwildmatch", ignore_patterns)
+
+    def matches_ignore(path: str) -> bool:
+        return spec.match_file(os.path.relpath(path, context_path))
```
mishushakov: here we are converting file full path into relative path for ignore check

packages/python-sdk/e2b/template/utils.py
```diff
@@ -34,6 +34,45 @@ def read_dockerignore(context_path: str) -> List[str]:
...
+            dir_files = glob(os.path.join(file, "**/*"), recursive=True)
+            for dir_file in dir_files:
+                if ignore_patterns and matches_ignore(dir_file):
+                    continue
+                files.add(dir_file)
+        else:
+            files.add(file)
+
+    return sorted(list(files))
+
```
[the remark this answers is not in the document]
mishushakov: I think it's okay because the glob lib in python doesn't have built-in mechanism for exclusion

packages/python-sdk/e2b/template/utils.py
```diff
@@ -85,29 +121,32 @@ def hash_stats(stat_info: os.stat_result) -> None:
...
     for file in files:
-        # Add a relative path to hash calculation
-        relative_path = os.path.relpath(file, context_path)
-        hash_obj.update(relative_path.encode())
+        # file is a relative path from get_all_files_for_files_hash
+        # Join it with context_path to get the absolute path
+        file_path = (
+            os.path.join(context_path, file) if not os.path.isabs(file) else file
+        )
+        hash_obj.update(file.encode())
```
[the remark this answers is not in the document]
mishushakov: oversight

packages/python-sdk/e2b/template/utils.py
```diff
@@ -85,9 +120,8 @@ def hash_stats(stat_info: os.stat_result) -> None:
         hash_obj.update(str(stat_info.st_mtime).encode())
 
     for file in files:
-        # Add a relative path to hash calculation
-        relative_path = os.path.relpath(file, context_path)
-        hash_obj.update(relative_path.encode())
+        # Hash the relative path (normalized by glob, matching JavaScript path.relative behavior)
+        hash_obj.update(file.encode())
```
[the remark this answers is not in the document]
mishushakov: oh fuk no

.changeset/green-mice-watch.md
```diff
@@ -0,0 +1,6 @@
+---
+'@e2b/python-sdk': patch
+'e2b': patch
+---
+
+add doublestar pattern when not specified to folder when calling .copy()
```
dobrac:
```suggestion
fix cache invalidation when calling .copy() on updated folder during the template build
```

packages/js-sdk/src/template/utils.ts
dobrac: add this comment to the code please

packages/js-sdk/src/template/utils.ts
```diff
@@ -23,6 +24,50 @@ export function readDockerignore(contextPath: string): string[] {
...
+      files.set(file.fullpath(), file)
+      const dirFiles = await glob(
+        path.join(path.relative(contextPath, file.fullpath()), '**/*'),
+        {
+          ignore: ignorePatterns,
+          withFileTypes: true,
+          cwd: contextPath,
+        }
+      )
+      dirFiles.forEach((f) => files.set(f.fullpath(), f))
```
dobrac, line 62: this is incorrect for the case where you have e.g. README.md in "/src/myfolder/README.md" and also in "/src/myfolder/another/README.md" and do copy("src", ".")

dobrac (changes requested):

packages/python-sdk/e2b/template/utils.py
```diff
@@ -34,6 +33,50 @@ def read_dockerignore(context_path: str) -> List[str]:
...
+
+    for file in files_glob:
+        # Join it with context_path to get the absolute path
+        file_path = os.path.join(context_path, file)
+
+        if os.path.isdir(file_path):
+            # If it's a directory, add the directory and all entries recursively
+            files.add(file_path)
+            dir_files = glob.glob(
+                os.path.join(file, "**/*"),
```
dobrac, line 66: is this correct file full path?

mishushakov: this is absolute path to context dir + file, same as JS version

packages/js-sdk/src/template/utils.ts
mishushakov: we're adding full absolute path, not just README.md to the map

packages/js-sdk/src/template/utils.ts
dobrac, line 47:
```suggestion
    // this is required so that the ignore pattern is relative to the file path
    cwd: contextPath,
```

PR #999, opened 2025-10-29
Add missing signal for request timeout

ValentaTomas:


dobrac (approved):
note: tests for this would be great

PR #1021, opened 2025-11-13
Some general pytest cleanup

djeebus:
- Individual tests must complete in less than 5 minutes
- Add a `make test` option that runs tests
- Upgrade poetry to 2.1.1 (the lock file was generated by this version, so this just matches what we already expect)

### DIFF
--- a/packages/python-sdk/pytest.ini
+++ b/packages/python-sdk/pytest.ini
@@ -4,4 +4,5 @@ markers =
     skip_debug: skip test if E2B_DEBUG is set.
 
 asyncio_mode=auto
-addopts = "--import-mode=importlib" "--numprocesses=4"
+addopts = "--import-mode=importlib"
+timeout = 300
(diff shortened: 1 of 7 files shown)


packages/python-sdk/pytest.ini
```diff
@@ -4,4 +4,5 @@ markers =
     skip_debug: skip test if E2B_DEBUG is set.
 
 asyncio_mode=auto
-addopts = "--import-mode=importlib" "--numprocesses=4"
```
ValentaTomas, line 7: What was the reason for changing the numprocesses?

(In the past it might have been to reduce load?)

djeebus: This configures options that are *always* added when running `pytest`. This option isn't super great locally, but is super useful when running CI, so I moved it into the Makefile. End result - running `pytest` won't parallelize by default, but CI will still use 4 processes

ValentaTomas (approved):

PR #1067, opened 2026-01-08
Throw when copying from paths outside the context dir

mishushakov:
> [!NOTE]
> **Medium Risk**
> Introduces stricter validation and path normalization for `copy` inputs, which can break existing templates that relied on absolute paths or `..` traversal; otherwise changes are localized and well-covered by tests.
> 
> **Overview**
> Prevents path traversal in template `copy` operations by **throwing when the source path is outside the template context directory** (absolute paths, `..` escapes), while preserving stack traces for better error attribution.
> 
> Adds cross-platform path normalization (`normalizePath`/`normalize_path`) and a shared safety check (`isSafeRelative`/`is_safe_relative`) in both the JS and Python SDKs, and updates CLI fixtures and Dockerfile parsing/tests to reflect normalized destinations (e.g., removing trailing slashes) plus new unit coverage for the helpers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bd6030477454690b3d899eb93b29afe311727d37. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/python-sdk/e2b/template/utils.py
```diff
@@ -119,6 +119,17 @@ def calculate_files_hash(
 
     :raises ValueError: If no files match the source pattern
     """
+    # check that src is not an absolute path or a path outside of the context directory
+    norm_path = os.path.normpath(src)
+    if (
+        os.path.isabs(norm_path)
+        or norm_path == ".."
+        or norm_path.startswith("../")
+        or norm_path.startswith("..\\")
```
jakubno: wouldn't it be enough to check if it starts with `..`? is there some edge case?

jakubno (commented):
why can't we throw directly in `.copy()` command? Can't you resolve the path directly there? You want to work with the relative path ideally everywhere anyway?

packages/python-sdk/e2b/template/utils.py
mishushakov: @jakubno to handle files like ..hello.txt. don't throw in .copy because we don't want any logic/side-effects happening in the functions (they're only for the template definition), so should throw only when you call .build

packages/js-sdk/src/template/utils.ts
```diff
@@ -369,3 +369,17 @@ export function readGCPServiceAccountJSON(
   }
   return JSON.stringify(pathOrContent)
 }
+
+export function isPathOutsideContext(src: string): boolean {
```
jakubno: shouldn't we pass the context dir? I think it can be simplified to something like this
```
export function isPathOutsideContext(src: string, contextDir: string): boolean {
  const resolvedPath = path.resolve(contextDir, src);
  const resolvedContext = path.resolve(contextDir);

  return (
    resolvedPath !== resolvedContext &&
    !resolvedPath.startsWith(resolvedContext + path.sep)
  );
}
```

packages/js-sdk/src/template/index.ts
```diff
@@ -375,6 +376,17 @@ export class TemplateBase
     const srcs = Array.isArray(src) ? src : [src]
 
     for (const src of srcs) {
+      if (isPathOutsideContext(src.toString())) {
```
jakubno: wouldn't it be better to work with resolved path after this or is there a reason why we need to keep it as the user passed it?

mishushakov: @jakubno it will be normalized by glob, but when I was passing a normalized path to it I saw some CLI tests failing because the normalization would change the separator from / to \ on windows:
https://github.com/e2b-dev/E2B/actions/runs/20931103739/job/60141939087

packages/js-sdk/src/template/utils.ts
mishushakov: @jakubno ~I am worried about the symlinks. we need to support symlinks that point to a location not on a local machine but inside the sandbox~

nevermind, i see that path.resolve does not resolve them

mishushakov: the problem here is that it will still allow ../ and /abs/path as long as the final path is within the context directory - we want to only allow relative paths

packages/python-sdk/e2b/template/utils.py
```diff
@@ -318,3 +318,19 @@ def read_gcp_service_account_json(
...
+    :param src: Path to check
+    :return: True if the path is outside of the context directory, False otherwise
+    """
+    norm_path = os.path.normpath(src)
+    return (
+        os.path.isabs(norm_path)
+        or norm_path == ".."
+        or norm_path.startswith("../")
+        or norm_path.startswith("..\\")
+    )
```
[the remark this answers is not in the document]
mishushakov: I thought it was different in Python

packages/js-sdk/src/template/index.ts
jakubno: that's a CI issue isn't it? instead of string equal it should be path equal right?

mishushakov:
Cursor review

packages/js-sdk/src/template/index.ts
```diff
@@ -375,9 +377,23 @@ export class TemplateBase
     const srcs = Array.isArray(src) ? src : [src]
 
     for (const src of srcs) {
+      const normalizedSrc = normalizePath(src.toString())
+      const normalizedDest = normalizePath(dest.toString())
```
jakubno, line 528: You can move this outside of the for loop

packages/js-sdk/tests/template/utils/isSafeRelative.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  describe('parent directory traversal', () => {
+    test('should return false for parent directory only', () => {
+      expect(isSafeRelative('..')).toBe(false)
+    })
+
+    test('should return true for paths starting with ../', () => {
+      expect(isSafeRelative('../file.txt')).toBe(false)
+    })
+
+    test('should return true for paths starting with ..\\', () => {
```
jakubno, line 20:
```suggestion
    test('should return false for paths starting with ..\\', () => {
```

packages/js-sdk/tests/template/utils/isSafeRelative.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      expect(isSafeRelative('..\\file.txt')).toBe(false)
+    })
+
+    test('should return true for normalized paths that escape context', () => {
```
jakubno, line 24:
```suggestion
    test('should return false for normalized paths that escape context', () => {
```

packages/js-sdk/tests/template/utils/isSafeRelative.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      expect(isSafeRelative('/absolute/path')).toBe(false)
+    })
+  })
+
+  describe('parent directory traversal', () => {
+    test('should return false for parent directory only', () => {
+      expect(isSafeRelative('..')).toBe(false)
+    })
+
+    test('should return true for paths starting with ../', () => {
```
jakubno, line 16:
```suggestion
    test('should return false for paths starting with ../', () => {
```

packages/js-sdk/tests/template/utils/isSafeRelative.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      expect(isSafeRelative('folder/file.txt')).toBe(true)
+    })
+
+    test('should return true for current directory references', () => {
+      expect(isSafeRelative('.')).toBe(true)
+      expect(isSafeRelative('./file.txt')).toBe(true)
+      expect(isSafeRelative('./folder/file.txt')).toBe(true)
+    })
+
+    test('should return false for glob patterns', () => {
```
jakubno, line 41:
```suggestion
    test('should return true for glob patterns', () => {
```

packages/js-sdk/tests/template/utils/isSafeRelative.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      expect(isSafeRelative('*.txt')).toBe(true)
+      expect(isSafeRelative('**/*.ts')).toBe(true)
+      expect(isSafeRelative('src/**/*')).toBe(true)
+    })
+
+    test('should return false for hidden files and directories', () => {
```
jakubno, line 47:
```suggestion
    test('should return true for hidden files and directories', () => {
```

packages/python-sdk/e2b/template/main.py
```diff
@@ -65,9 +67,27 @@ def copy(
         srcs = [src] if isinstance(src, (str, Path)) else src
 
         for src_item in srcs:
+            normalized_src_item = normalize_path(str(src_item))
+            normalized_dest = normalize_path(str(dest))
```
jakubno, line 71: This can be done outside of the loop

packages/python-sdk/e2b/template/main.py
```diff
@@ -65,9 +67,27 @@ def copy(
...
+
+            if not is_safe_relative(str(src_item)):
```
jakubno, line 73: shouldn't this use the normalized path?

packages/js-sdk/src/template/utils.ts
```diff
@@ -25,12 +25,50 @@ export function readDockerignore(contextPath: string): string[] {
...
+  let workingPath = path
+  if (/^[a-zA-Z]:/.test(path)) {
+    workingPath = path.substring(2)
+  }
+
+  // Check if path is absolute
+  const isAbsolute = workingPath.startsWith('/') || workingPath.startsWith('\\')
+
+  // Normalize all separators to forward slashes and split
+  const normalizedPath = workingPath.replace(/[\\/]+/g, '/')
```
jakubno, line 86:
```suggestion
  // Normalize all separators to forward slashes and split
  const normalizedPath = workingPath.replace(/[\\/]+/g, '/')

  // Check if path is absolute
  const isAbsolute = workingPath.startsWith('/')
```

packages/js-sdk/src/template/utils.ts
```diff
@@ -25,12 +25,50 @@ export function readDockerignore(contextPath: string): string[] {
...
+  const parts = normalizedPath.split('/').filter((part) => part && part !== '.')
+
+  const normalized: string[] = []
+
+  for (const part of parts) {
+    if (part === '..') {
+      // Go up one directory if possible (but not past root for absolute paths)
+      if (normalized.length > 0 && normalized[normalized.length - 1] !== '..') {
```
jakubno, line 94: what if last part was `'.'`?

packages/js-sdk/tests/template/utils/normalizePath.test.ts
```diff
@@ -0,0 +1,78 @@
+import { expect, test, describe } from 'vitest'
+import { normalizePath } from '../../../src/template/utils'
+
+describe('normalizePath', () => {
```
jakubno, line 4: add a test for `a/./../b/`

packages/python-sdk/e2b/template/utils.py
```diff
@@ -37,12 +37,41 @@ def read_dockerignore(context_path: str) -> List[str]:
...
+    Normalize paths in a platform-independent, POSIX-style manner.
+    Mirrors the JS normalizePath behavior.
     """
-    return path.replace(os.sep, "/")
+    if not path or path == ".":
+        return "."
+
+    # Remove drive letter if present (e.g. "C:")
+    working_path = path
+    if re.match(r"^[a-zA-Z]:", path):
```
jakubno, line 69: `:` can be in the name, so `a:b/c.txt` is valid path in UNIX, we should do this check only if we're on windows, same in JS

packages/python-sdk/e2b/template/utils.py
```diff
@@ -37,12 +37,41 @@ def read_dockerignore(context_path: str) -> List[str]:
...
+        working_path = path[2:]
+
+    # Determine if absolute
+    is_absolute = working_path.startswith("/") or working_path.startswith("\\")
+
+    # Normalize separators to '/'
+    normalized_path = re.sub(r"[\\/]+", "/", working_path)
```
jakubno, line 76: reverse the order to simplify

packages/python-sdk/e2b/template/utils.py
```diff
@@ -37,12 +37,41 @@ def read_dockerignore(context_path: str) -> List[str]:
...
+
+    # Split and process components
+    parts = [p for p in normalized_path.split("/") if p and p != "."]
+
+    normalized = []
+
+    for part in parts:
+        if part == "..":
+            if normalized and normalized[-1] != "..":
```
jakubno, line 85: same issue as in JS with `./..`

packages/python-sdk/e2b/template/utils.py
```diff
@@ -318,3 +347,21 @@ def read_gcp_service_account_json(
...
+    :param src: The path to check
+
+    :return: True if the path is a safe relative path, False otherwise
+    """
+    norm_path = normalize_path(src)
+    return not (
+        os.path.isabs(norm_path)
+        or norm_path == ".."
+        or norm_path.startswith("../")
+        or norm_path.startswith("..\\")
```
jakubno, line 387: normpath doesn't contain `\` anymore right?

packages/python-sdk/tests/async/template_async/methods/test_from_dockerfile.py
```diff
@@ -69,8 +69,8 @@ async def test_from_dockerfile_with_custom_user_and_workdir():
 @pytest.mark.skip_debug()
 async def test_from_dockerfile_with_copy_chown():
     dockerfile = """FROM node:24
-COPY --chown=myuser:mygroup app.js /app/
-COPY --chown=anotheruser config.json /config/"""
+COPY --chown=myuser:mygroup app.js /app
```
jakubno, line 72: why did you remove the extra slashes?

packages/python-sdk/tests/shared/template/utils/test_normalize_path.py
```diff
@@ -0,0 +1,61 @@
+from e2b.template.utils import normalize_path
+
+
+class TestBasicPathNormalization:
```
jakubno, line 4: add a test with `./..`

Issue #1055, opened 2025-12-22
CLI resets the timeout to 5 minutes after connecting to a running Sandbox


mishushakov:
**Describe the bug**

CLI overrides the Sandbox timeout to 5 minutes after calling `e2b sbx connect <sandbox id>`.

**To Reproduce**
1. I launch a sandbox via the API with 30 minutes timeout - works great
2. I don't do anything else on the sandbox via the API
3. I log in to the sandbox using the E2B CLI
4. The sandbox is killed exactly 5 minutes after I logged in via the CLI.

**Expected behavior**
The timeout should be kept as is.

**Terminal commands & output**

Before connecting via CLI (30 min timeout):

```
icavd7j6rub8vi6fczr7x  rki5dems9wqfm4r03t7g  base   12/22/2025, 2:06:31 PM  12/22/2025, 2:36:31 PM  Running  2      512      0.4.3         {}
```

After connecting via CLI (5 min timeout):

```
icavd7j6rub8vi6fczr7x  rki5dems9wqfm4r03t7g  base   12/22/2025, 2:06:31 PM  12/22/2025, 2:11:45 PM  Running  2      512      0.4.3         {}
```

**Additional context**
CLI version: 2.3.1

bxxf:
Hey, I decided to have a look into this and I believe this issue has been **fixed in CLI version 2.4.0+** (you were using 2.3.1). There were changes made to the API endpoints and SDK behavior to prevent reducing the timeout for running sandboxes. Pls check my findings below - I'm referencing old commits there which i found to be relevant.

---

### What Was Happening (CLI 2.3.1 and earlier)

The old SDK implementation used two separate API calls when connecting to a sandbox:

1. `POST /sandboxes/{id}/resume` - to resume the sandbox
2. `POST /sandboxes/{id}/timeout` - to set the timeout

The `/timeout` endpoint internally called `KeepAliveFor()` with `allowShorter=true`, which **allowed** reducing the timeout. Since the CLI didn't specify a custom timeout, the SDK defaulted to 5 minutes (`DEFAULT_SANDBOX_TIMEOUT_MS = 300000`), which overwrote your 30-minute timeout.

**Code reference (old behavior):**

[...]

PR #1075, opened 2026-01-10
Meta feature: add devcontainer support for standardization (Fedora Linux fails when running tests)

digitr4ce:
# Issue
Running tests in `js-sdk` in a Fedora Linux distribution fails, because Playwright does not have official support for Linux distributions outside of Debian / Ubuntu based distros.

# Details
Ran into the issue while running the tests for a change I was trying to do regarding bug #1055. I was trying to fix it to get a feel for the repository and code structure.

# Fix
Add a `.devcontainer` folder with a configuration file for VSCode and a generic Dockerfile. This allows for development in an isolated container based on Ubuntu, no matter the Linux distribution.

This should allow Linux users to more freely contribute to the project.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,36 @@
+FROM ubuntu:22.04
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install -y \
+    curl \
+    git \
+    wget \
+    vim \
+    ssh \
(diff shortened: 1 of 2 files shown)


mishushakov:
Thanks for trying to contribute!
We are not proceeding with this PR as our team does not actively use DevContainers (we are using E2B Sandboxes).

Closing this. Hope for your understanding 🙏

PR #1076, opened 2026-01-11
fix: convert float timeout to int before string conversion

### LINKED ISSUE #1063: Bug: Inconsistent typing for timeout causes the backend to fail
**Describe the bug**

Inconsistent typing for `timeout` parameter in `commands.run()`. The public API declares `timeout: Optional[float] = 60` but the internal implementation in `e2b_connect` expects an `int`. When a float is passed (e.g., `60.0`), it gets multiplied by 1000 and converted to string as `"60000.0"`, which causes the backend to fail parsing with `strconv.ParseInt`.

**To Reproduce**

1. Go to the [function definition](https://github.com/e2b-dev/E2B/blob/ca556d8b426e9ffe6b130bf9732728471a8e01d0/packages/python-sdk/e2b/sandbox_sync/commands/command.py#L146) - note `timeout: Optional[float] = 60`

2. Compare with [`_create_stream_timeout`](https://github.com/e2b-dev/E2B/blob/ca556d8b426e9ffe6b130bf9732728471a8e01d0/packages/python-sdk/e2b_connect/client.py#L308) in `e2b_connect/client.py`:
   ```python
   def _create_stream_timeout(self, timeout: Optional[int]):
       if timeout:
           return {"connect-timeout-ms": str(timeout * 1000)}
       return {}
   ```

3. Call `sandbox.commands.run("echo hello", timeout=60.0)`

4. Observe error: `InvalidArgumentException: parse timeout: strconv.ParseInt: parsing "60000.0": invalid syntax`

**Expected behavior**

Either:
- The public API type should be `int` to match internal expectations, OR
- The internal `_create_stream_timeout` should cast to int: `str(int(timeout * 1000))`

**Terminal commands & output**

```
[DEBUG] EXCEPTION in execute_code_with_pool: InvalidArgumentException: parse timeout: strconv.ParseInt: parsing "60000.0": invalid syntax
output: Execution failed: parse timeout: strconv.ParseInt: parsing "60000.0": invalid syntax
```

**Desktop (please complete the following information):**

- OS: Linux 4.18.0-553.el8_10.x86_64
- Python: 3.12/3.13
- e2b version: latest (installed via pip)

**Additional context**

The issue is in the type mismatch between:
1. `e2b/sandbox_sync/commands/command.py` line 146: `timeout: Optional[float] = 60`
2. `e2b_connect/client.py` line 308: `def _create_stream_timeout(self, timeout: Optional[int])`

When `timeout=60.0` (float) is passed:
- `60.0 * 1000 = 60000.0` (still a float)
- `str(60000.0) = "60000.0"` (includes decimal point)
- Go's `strconv.ParseInt("60000.0")` fails because it can't parse floats

**Workaround**

Cast timeout to int when calling:
```python
sandbox.commands.run(cmd, timeout=int(timeout))
```

theonlypal:
## Summary

The public API accepts `Optional[float]` for timeout parameters (e.g., `60.0`), but `_create_stream_timeout` was typed as `Optional[int]` and passed the value directly to `str()`, producing `'60000.0'` instead of `'60000'`.

The Go backend's `strconv.ParseInt` fails on float strings, causing 'invalid syntax' errors for valid timeout values.

## Changes

- Updates type hint to `Optional[float]` for consistency with public API
- Wraps `timeout * 1000` in `int()` before `str()` conversion

## Test

```python
# Before: str(60.0 * 1000) -> '60000.0' (fails ParseInt)
# After:  str(int(60.0 * 1000)) -> '60000' (works)
```

Fixes #1063

### DIFF
--- a/.changeset/fix-timeout-float-conversion.md
+++ b/.changeset/fix-timeout-float-conversion.md
@@ -1,0 +1,5 @@
+---
+'e2b': patch
+---
+
+fix float timeout to int conversion for Go backend compatibility
--- a/packages/python-sdk/e2b_connect/client.py
+++ b/packages/python-sdk/e2b_connect/client.py
@@ -305,9 +305,9 @@ class Client:
         res = self.pool.request(**req_data)
         return self._process_unary_response(res)
 
-    def _create_stream_timeout(self, timeout: Optional[int]):
+    def _create_stream_timeout(self, timeout: Optional[float]):
         if timeout:
-            return {"connect-timeout-ms": str(timeout * 1000)}
+            return {"connect-timeout-ms": str(int(timeout * 1000))}
         return {}
 
     def _prepare_server_stream_request(


jakubno:
@theonlypal, thanks for the contribution!

PR #1082, opened 2026-01-14
Update bug_report form template

beran-t:
New GitHub issue form with more specific fields for customer reports

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -7,15 +7,13 @@ body:
     id: sandbox_id
     attributes:
       label: Sandbox ID or Build ID
-    validations:
-      required: true
 
-  - type: input
-    id: sdk_version
+  - type: textarea
+    id: environment
     attributes:
-      label: SDK version
-      description: Can be found in `pip list` or `npm list`
-      placeholder: e.g. e2b==1.2.3
+      label: Environment
+      description: Environment details (used libraries version, OS, etc..)
+      placeholder: e2b==1.2.3, macOS 14
     validations:
       required: true
 
@@ -36,15 +34,6 @@ body:
         - One-time occurrence
         - Happens occasionally
         - Happens every time
-    validations:
-      required: true
-
-  - type: input
-    id: os
-    attributes:
-      label: Operating system
-      description: OS used to run your code (not the sandbox OS)
-      placeholder: e.g. macOS 14, Ubuntu 22.04, Windows 11
     validations:
       required: true
 


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,81 @@
...
+      label: When did this start happening?
+      placeholder: e.g. Today / Yesterday / After upgrading SDK
+
+  - type: dropdown
+    id: frequency
+    attributes:
+      label: Frequency
+      options:
+        - One-time occurrence
+        - Happens sometimes
```
mishushakov: let's use "occasionally"

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,81 @@
...
+        4. Actual behavior  
+        5. Additional context (logs, screenshots, etc.)
+
+  - type: textarea
+    id: issue_details
+    attributes:
+      label: Issue details
+      description: Describe the issue using the guidance above
+    validations:
+      required: true
```
mishushakov: let's remove this one

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,81 @@
...
+        Please include the following in the field below:
+
+        1. Bug description  
+        2. Steps to reproduce  
+          - Go to `...`  
+          - Run `...`  
+          - See error  
+        3. Expected behavior  
+        4. Actual behavior  
+        5. Additional context (logs, screenshots, etc.)
```
mishushakov: let's do instead markdown fields:

- Expected behavior (required)
- Actual behavior (required)
- Issue reproduction (required)
- Additional context (logs, screenshots, etc.)

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,81 @@
...
+      description: When did the error occur? (include timezone please)
+      placeholder: e.g. 2026-01-12 14:32 UTC
+    validations:
+      required: true
+
+  - type: input
+    id: start_time
+    attributes:
+      label: When did this start happening?
+      placeholder: e.g. Today / Yesterday / After upgrading SDK
```
mishushakov: would probably remove this

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,81 @@
...
+labels: ["bug"]
+body:
+  - type: input
+    id: sandbox_id
+    attributes:
+      label: Sandbox ID or Build ID
+      description: Output of `sbx.sandbox_id` or the build ID printed in the console during the build
+      placeholder: e.g. sbx_123456
+    validations:
+      required: true
```
mishushakov: should not be required

.github/ISSUE_TEMPLATE/bug_report.yaml
beran-t: So you want to break it down into 4 text areas rather than having it one text-area? That's what I had originaly, but it seemed too much to me. 
Do you have something like this in mind? 
https://github.com/RatArt/testtheform/issues

PR #1087, opened 2026-01-19
Add headers to EnvdApiClient

beran-t:
Adding proper headers for EnvdApiClient.

### DIFF
--- a/.changeset/curly-kiwis-doubt.md
+++ b/.changeset/curly-kiwis-doubt.md
@@ -1,0 +1,5 @@
+---
+'e2b': patch
+---
+
+Add user-agent header to EnvdApiClient


packages/js-sdk/src/sandbox/index.ts
```diff
@@ -177,9 +177,12 @@ export class Sandbox extends SandboxApi {
         apiUrl: this.envdApiUrl,
         logger: opts?.logger,
         accessToken: this.envdAccessToken,
-        headers: this.envdAccessToken
-          ? { 'X-Access-Token': this.envdAccessToken }
-          : {},
+        headers: {
+          ...this.connectionConfig.headers,
```
ValentaTomas: I'm not 100% sure we want to add all connection headers as some might be specific to the connect rpc.

Let's just recheck that.

PR #1090, opened 2026-01-22
Add exec command to CLI

ValentaTomas:
> [!NOTE]
> Adds the ability to execute commands inside a running sandbox from the CLI.
> 
> - New `sandbox exec` command (`exec.ts`): runs a command in a specified sandbox, supports `--background`, `--cwd`, `--user`, and repeatable `--env` options; streams stdout/stderr; returns remote exit code; explicitly disallows stdin piping due to protocol limitations
> - Signal handling utility (`utils/signal.ts`): installs/removes handlers to kill the remote process on termination signals
> - Registers `exec` in `sandbox/index.ts`
> - Updates `configOption` help in `options.ts` to recommend the new build system
> - Adds changeset for `@e2b/cli` minor release
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dcbc64211934c1f92ff033ff77ff6ad50497e8a7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/cli/src/commands/sandbox/exec.ts
+++ b/packages/cli/src/commands/sandbox/exec.ts
@@ -101,20 +101,20 @@ async function runCommand(
     user: opts.user,
     envs: opts.env,
     timeoutMs: NO_COMMAND_TIMEOUT,
-    onStdout: (data) => {
+    onStdout: async (data) => {
       try {
         process.stdout.write(data)
       } catch (err: any) {
         console.error(err)
-        handle.kill()
+        await handle.kill()
       }
     },
-    onStderr: (data) => {
+    onStderr: async (data) => {
       try {
         process.stderr.write(data)
       } catch (err: any) {
         console.error(err)
-        handle.kill()
+        await handle.kill()
       }
     },
   })


packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,230 @@
...
+    startForwarding() {
+      process.stdin.removeAllListeners('data')
+      process.stdin.on('data', async (data) => {
+        try {
+          await sendStdin(data.toString())
+        } catch {
+          // Command may have exited
+        }
+      })
+    },
```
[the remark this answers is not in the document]
ValentaTomas: removed stdin for now

packages/cli/src/terminal.ts
```diff
@@ -62,7 +62,7 @@ export async function spawnConnectedTerminal(sandbox: e2b.Sandbox) {
   }
 }
 
-class BatchedQueue<T> {
+export class BatchedQueue<T> {
```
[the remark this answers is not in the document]
ValentaTomas: not used now

packages/cli/src/terminal.ts
```diff
@@ -95,6 +95,14 @@ class BatchedQueue<T> {
     await this.flush()
   }
 
+  cancel() {
+    if (this.intervalId) {
+      clearInterval(this.intervalId)
+      this.intervalId = undefined
+    }
+    this.queue = []
+  }
```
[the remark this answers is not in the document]
ValentaTomas: not used now

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,230 @@
...
+  if (err instanceof AuthenticationError) {
+    console.error('e2b: authentication failed - check E2B_API_KEY')
+    process.exit(1)
+  }
+
+  // Invalid argument - translate SDK params to CLI flags
+  if (err instanceof InvalidArgumentError) {
+    let message = err.message
+    message = message.replace(/\bcwd\b/gi, '--cwd')
+    message = message.replace(/\buser\b/gi, '--user')
```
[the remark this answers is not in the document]
ValentaTomas: Minimized changes

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,189 @@
...
+    },
+    {} as Record<string, string>
+  )
+  .option(
+    '-t, --timeout <ms>',
+    'timeout in milliseconds (default 0 = no timeout)',
+    parseInt,
+    0
+  )
+  .alias('ex')
```
mishushakov, line 46: probably don't need this

mishushakov: shortcut for execute is exec anyways

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,189 @@
...
+  .action(
+    async (sandboxID: string, commandParts: string[], opts: ExecOptions) => {
+      // Warn if stdin is being piped (not supported)
+      if (!process.stdin.isTTY) {
+        console.error(
+          'e2b: warning: stdin piping is not supported, input will be ignored'
+        )
```
mishushakov: make sure we exit here with 1

packages/cli/src/commands/sandbox/exec.ts
ValentaTomas: Keeping for symmetry :D

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,189 @@
...
+        if (opts.background) {
+          const handle = await sandbox.commands.run(command, {
+            background: true,
+            cwd: opts.cwd,
+            user: opts.user,
+            envs: opts.env,
+            timeoutMs: opts.timeout,
+          })
+
+          console.error(handle.pid)
```
mishushakov, line 75: please through stdout not stderr

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,189 @@
...
+        await runCommand(sandbox, command, opts)
+
+        process.exit(0)
+      } catch (err: any) {
+        handleExecError(err, sandboxID)
+      }
+    }
+  )
+
+interface ExecOptions {
```
mishushakov: move this all the way to the top

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  try {
+    const result = await handle.wait()
+
+    if (result.exitCode !== 0) {
+      process.exit(result.exitCode)
+    }
+  } catch (err) {
+    if (err instanceof CommandExitError) {
+      process.exit(err.exitCode)
```
mishushakov: I am thinking that maybe you should explicitly add return when you do process.exit because it might be confusing

packages/cli/src/commands/sandbox/exec.ts
ValentaTomas: Exit yes, but maybe this should not be 0?

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,189 @@
...
+  if (err instanceof AuthenticationError) {
+    console.error('e2b: authentication failed - check E2B_API_KEY')
+    process.exit(1)
+  }
+
+  // Invalid argument - translate SDK params to CLI flags
+  // Only replace technical terms unlikely to appear in normal text
+  if (err instanceof InvalidArgumentError) {
+    let message = err.message
+    message = message.replace(/\btimeout\b/g, '--timeout')
```
mishushakov: we probably don't need this

packages/cli/src/commands/sandbox/exec.ts
ValentaTomas: I'm not sure here. As for example if I use `&` for background process, the process ID info is printed to stderr. This is effectively not output, but adjanced info.

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    console.error(`e2b: ${message}`)
+    process.exit(1)
+  }
+
+  // Generic error - translate common SDK params to CLI flags
+  const errMessage = err instanceof Error ? err.message : String(err)
+  let message = errMessage
+  message = message.replace(/\btimeout\b/gi, '--timeout')
```
mishushakov: again, unsure about this

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,189 @@
...
+    await handle.kill().catch(() => {})
+  })
+
+  try {
+    const result = await handle.wait()
+
+    if (result.exitCode !== 0) {
+      process.exit(result.exitCode)
+    }
+  } catch (err) {
```
mishushakov, line 131: here you don't run handleExecError? can we standardize a bit?

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,189 @@
...
+    if (err instanceof CommandExitError) {
+      process.exit(err.exitCode)
+    }
+
+    throw err
```
mishushakov, line 141: here you throw but in the other places like handleExecError you exit?

packages/cli/src/commands/sandbox/exec.ts
ValentaTomas: Anything after gets linted out as inaccessible, so probably won't be adding that as that clarifies.

packages/cli/src/commands/sandbox/exec.ts
ValentaTomas: Hmm, returning `2` as "Misuse of Shell Built-in" could work well here.

packages/cli/src/commands/sandbox/exec.ts
ValentaTomas: refactored

packages/cli/src/commands/sandbox/exec.ts
ValentaTomas: ok

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,219 @@
...
+    // If we're exiting due to a signal, use the signal exit code
+    if (signalExit !== null) {
+      process.exit(signalExit)
+    }
+
+    handleExecError(err, sandbox.sandboxId)
+  }
+}
+
+function handleExecError(err: unknown, sandboxID: string): never {
```
mishushakov: tbh I am thinking if we should't just throw err and be done with it

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,219 @@
...
+    // Mark that we're exiting due to signal
+    signalExit = getSignalExitCode(signal)
+
+    // Kill the remote process and wait for it.
+    // The exec handler should also remove the signal handler and process exit.
+    try {
+      await handle.kill()
+    } catch {
+      // Ignore kill errors (sandbox may already be gone) or there might be network issues.
+      removeSignalHandlers()
```
mishushakov: this doesn't do anything?

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,219 @@
...
+    if (handle.error) {
+      console.error(handle.error)
+    }
+
+    // If we're exiting due to a signal, use the signal exit code
+    if (signalExit !== null) {
+      process.exit(signalExit)
+    }
+
+    handleExecError(err, sandbox.sandboxId)
```
mishushakov: I'd put this here instead:

```ts
  if (err instanceof CommandExitError) {
    process.exit(err.exitCode)
  }
```

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,219 @@
...
+
+import * as e2b from 'e2b'
+import {
+  CommandExitError,
+  SandboxError,
+  TimeoutError,
+  NotFoundError,
+  AuthenticationError,
+  InvalidArgumentError,
+} from 'e2b'
```
mishushakov: duplicate import

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,129 @@
...
+    user: opts.user,
+    envs: opts.env,
+    timeoutMs: NO_COMMAND_TIMEOUT,
+    onStdout: (data) => {
+      process.stdout.write(data)
+    },
+    onStderr: (data) => {
+      process.stderr.write(data)
+    },
+  })
```
mishushakov, line 120: what if this throws?

ValentaTomas: Will handle it for completeness.

ValentaTomas: I think there can be some cases when you close the output pipe you are piping the output to where it potentially can happen.

packages/cli/src/commands/sandbox/exec.ts
```diff
@@ -0,0 +1,145 @@
...
+    cwd: opts.cwd,
+    user: opts.user,
+    envs: opts.env,
+    timeoutMs: NO_COMMAND_TIMEOUT,
+    onStdout: (data) => {
+      try {
+        process.stdout.write(data)
+      } catch (err: any) {
+        console.error(err)
+        handle.kill()
```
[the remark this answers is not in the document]
ValentaTomas: Added just for cleanup, but this would not affect anything.

PR #1098, opened 2026-01-26
Avoid reading files on upload

mishushakov:
> [!NOTE]
> Optimizes file uploads in both JS and Python SDKs to avoid unnecessary reads and add broader input support.
> 
> - JS SDK: `Filesystem.write`/`writeFiles` now build a single `FormData` using `toBlob` (new util) to pass `string`/`ArrayBuffer`/`Blob`/`ReadableStream` without pre-reading; updated tests add `ReadableStream` coverage
> - Python SDK (sync/async): `write_files` accepts `str`/`bytes` directly, reads `TextIOBase`, and passes `IOBase` (binary) streams through without reading; new tests cover `BytesIO` and `StringIO`
> - Changeset: patch bumps for `@e2b/python-sdk` and `e2b`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 45516e31fa8ba8b678824a6c1adc217287a8effe. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/js-sdk/tests/sandbox/files/write.test.ts
+++ b/packages/js-sdk/tests/sandbox/files/write.test.ts
@@ -228,16 +228,25 @@ sandboxTest('writeFiles with different data types', async ({ sandbox }) => {
   const textData = 'Text string data'
   const arrayBufferData = new TextEncoder().encode('ArrayBuffer data').buffer
   const blobData = new Blob(['Blob data'], { type: 'text/plain' })
+  const streamContent = 'ReadableStream data'
+  const encoder = new TextEncoder()
+  const streamData = new ReadableStream({
+    start(controller) {
+      controller.enqueue(encoder.encode(streamContent))
+      controller.close()
+    },
+  })
 
   const files: WriteEntry[] = [
     { path: 'writefiles_text.txt', data: textData },
     { path: 'writefiles_arraybuffer.txt', data: arrayBufferData },
     { path: 'writefiles_blob.txt', data: blobData },
+    { path: 'writefiles_stream.txt', data: streamData },
   ]
 
   const infos = await sandbox.files.writeFiles(files)
 
-  assert.equal(infos.length, 3)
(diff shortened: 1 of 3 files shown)


packages/python-sdk/e2b/sandbox_sync/filesystem/filesystem.py
```diff
@@ -210,10 +210,15 @@ def write_files(
         httpx_files = []
         for file in files:
             file_path, file_data = file["path"], file["data"]
-            if isinstance(file_data, str) or isinstance(file_data, bytes):
+            if isinstance(file_data, (str, bytes)):
+                # str and bytes can be passed directly
                 httpx_files.append(("file", (file_path, file_data)))
-            elif isinstance(file_data, IOBase):
+            elif isinstance(file_data, TextIOBase):
+                # Text streams must be read first
```
ValentaTomas, line 217: Why text streams cannot be passed too?

packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -354,9 +355,11 @@ export class Filesystem {
 
     if (writeFiles.length === 0) return [] as WriteInfo[]
 
-    const blobs = await Promise.all(
-      writeFiles.map((f) => new Response(f.data).blob())
-    )
+    const formData = new FormData()
```
ValentaTomas, line 358: Can we add tests for all the types that can be passed, so we are sure they get properly processed?

packages/python-sdk/e2b/sandbox_sync/filesystem/filesystem.py
mishushakov: because it will attempt to consume the text stream and it would be a stream of str not bytes

PR #1101, opened 2026-01-27
Feature: Add Git Support

matthewlouisbrockman:
# Sandbox Git Commands

Adds Git support to the sandbox class. This allows the sandbox to manage git via standard clone, checkout, branch, add, pull, and push commands without needing to use commands.run. The API mirrors common Git workflows while handling sandbox-specific concerns like auth injection and safe remote handling.

**Python example**
```python
from e2b import Sandbox

sandbox = Sandbox.create()
repo_path = '/home/user/my-repo'

# Optional: set author for commits
sandbox.git.configure_user('Your Name', 'you@example.com')

# Clone or init
sandbox.git.clone('https://github.com/org/repo.git', path=repo_path)
# or
sandbox.git.init(repo_path, initial_branch='main')

# Make a change
sandbox.files.write(f'{repo_path}/README.md', '# Hello\n')

# Commit
sandbox.git.add(repo_path, files=['README.md'])
sandbox.git.commit(repo_path, message='Initial commit')

# Branching
sandbox.git.create_branch(repo_path, 'feature1')
# or
sandbox.git.checkout_branch(repo_path, 'main')

# Push
sandbox.git.remote_add(repo_path, 'origin', 'https://github.com/org/repo.git', overwrite=True)
sandbox.git.push(repo_path, remote='origin', branch='main', set_upstream=True)
```

**JavaScript / TypeScript example**
```ts
import { Sandbox } from 'e2b'

const repoPath = '/home/user/my-repo'

await sandbox.git.configureUser('Your Name', 'you@example.com')

await sandbox.git.clone('https://github.com/org/repo.git', { path: repoPath })
// or
await sandbox.git.init(repoPath, { initialBranch: 'main' })

await sandbox.files.write(`${repoPath}/README.md`, '# Hello\n')

await sandbox.git.add(repoPath, { files: ['README.md'] })
await sandbox.git.commit(repoPath, { message: 'Initial commit' })

await sandbox.git.createBranch(repoPath, 'feature1')
await sandbox.git.checkoutBranch(repoPath, 'main')

await sandbox.git.remoteAdd(repoPath, 'origin', 'https://github.com/org/repo.git', {
  overwrite: true,
})
await sandbox.git.push(repoPath, { remote: 'origin', branch: 'main', setUpstream: true })
```

**Main commands**
- `clone`: Clone a repo into the sandbox. Supports `branch`, `depth`, optional `username` + `password` for private repos, and `dangerously_store_credentials` / `dangerouslyStoreCredentials` to keep credentials in the remote URL.
- `init`: Initialize a new repo. Supports `initial_branch` / `initialBranch` and `bare`.
- `status`: Get parsed `git status --porcelain -b` info.
- `branches`: List branches and current branch.
- `create_branch` / `createBranch`: Create and check out a new branch.
- `checkout_branch` / `checkoutBranch`: Switch to an existing branch.
- `delete_branch` / `deleteBranch`: Delete a branch. Supports `force`.
- `add`: Stage files. Supports explicit files or `all`.
- `commit`: Create a commit. Supports author override and `allow_empty`.
- `reset` / `reset`: Reset `HEAD` (supports modes like `soft`, `mixed`, `hard`, etc.) and optional paths.
- `restore` / `restore`: Restore files or unstage changes (`worktree` / `staged`) from a source ref.
- `pull`: Pull from a remote. Supports `remote`, `branch`, and optional auth.
- `push`: Push to a remote. Supports `remote`, `branch`, `set_upstream`, and optional auth.
- `remote_add` / `remoteAdd`: Add a remote. Supports `overwrite` and `fetch`.
- `remote_get` / `remoteGet`: Read a remote URL.
- `set_config` / `setConfig`: Set a git config value. Supports `scope` (`global`, `local`, `system`), and `path` for local scope.
- `get_config` / `getConfig`: Read a git config value. Supports the same `scope` options and returns `None` / `undefined` if unset.
- `dangerously_authenticate` / `dangerouslyAuthenticate`: Persist credentials via the git credential helper (global).
- `configure_user` / `configureUser`: Set default `user.name` and `user.email` for commits.
- `create_github_repo` (Python only): Create a GitHub repo from inside the sandbox and optionally add it as a remote.

**Status shape**
- `status` returns a `GitStatus` with `current_branch` / `currentBranch`, `upstream`, `ahead`, `behind`, `detached`, and `file_status` / `fileStatus`.
- `file_status` entries include `name`, `status`, `index_status` / `indexStatus`, `working_tree_status` / `workingTreeStatus`, `staged`, and optional `renamed_from` / `renamedFrom`.
- Convenience helpers include: `is_clean` / `isClean`, `has_changes` / `hasChanges`, `has_staged` / `hasStaged`, `has_untracked` / `hasUntracked`, `has_conflicts` / `hasConflicts`, plus counts (`total_count` / `totalCount`, `staged_count` / `stagedCount`, `unstaged_count` / `unstagedCount`, `untracked_count` / `untrackedCount`, `conflict_count` / `conflictCount`). In Python these are properties on the `GitStatus` object; in JS they are fields on the returned object.

**Notes**
- For private HTTPS remotes, pass `username` + `password` (token) on `clone`, `pull`, or `push`.
- Use `remote_add` / `remoteAdd` with `overwrite=True` to update an existing remote URL and `fetch=True` to fetch after.
- Use `dangerously_authenticate` / `dangerouslyAuthenticate` only when you want to persist credentials globally on the sandbox.

### DIFF
--- a/packages/js-sdk/src/sandbox/git/index.ts
+++ b/packages/js-sdk/src/sandbox/git/index.ts
@@ -39,7 +39,7 @@ export interface GitRequestOpts
       CommandStartOpts,
       'envs' | 'user' | 'cwd' | 'timeoutMs' | 'requestTimeoutMs'
     >
-  > { }
+  > {}
 
 /**
  * Options for cloning a repository.
@@ -284,7 +284,7 @@ export interface GitDangerouslyAuthenticateOpts extends GitRequestOpts {
  * Module for running git operations in the sandbox.
  */
 export class Git {
-  constructor(private readonly commands: Commands) { }
+  constructor(private readonly commands: Commands) {}
 
   /**
    * Clone a git repository into the sandbox.
@@ -1044,7 +1044,6 @@ export class Git {
       return false
     }
   }
-
 }
 
 export type {
--- a/packages/python-sdk/e2b/sandbox_async/git.py
+++ b/packages/python-sdk/e2b/sandbox_async/git.py
@@ -7,21 +7,21 @@ from e2b.exceptions import (
 )
 from e2b.sandbox.commands.command_handle import CommandExitException
 from e2b.sandbox.git_utils import (
-  GitBranches,
-  GitStatus,
-  build_auth_error_message,
-  build_git_command,
-  build_push_args,
-  build_upstream_error_message,
-  is_auth_failure,
-  is_missing_upstream,
-  parse_git_branches,
-  parse_git_status,
-  resolve_config_scope,
-  shell_escape,
-  strip_credentials,
-  derive_repo_dir_from_url,
-  with_credentials,
+    GitBranches,
+    GitStatus,
+    build_auth_error_message,
+    build_git_command,
+    build_push_args,
+    build_upstream_error_message,
+    is_auth_failure,
+    is_missing_upstream,
+    parse_git_branches,
+    parse_git_status,
+    resolve_config_scope,
+    shell_escape,
+    strip_credentials,
+    derive_repo_dir_from_url,
+    with_credentials,
 )
 from e2b.sandbox_async.commands.command import Commands
 
@@ -324,9 +324,7 @@ class Git:
         except CommandExitException as err:
             if is_auth_failure(err):
                 raise GitAuthException(
-                    build_auth_error_message(
-                        "clone", bool(username) and not password
-                    )
+                    build_auth_error_message("clone", bool(username) and not password)
                 ) from err
             raise
 
@@ -776,9 +774,7 @@ class Git:
         except CommandExitException as err:
             if is_auth_failure(err):
                 raise GitAuthException(
-                    build_auth_error_message(
-                        "push", bool(username) and not password
-                    )
+                    build_auth_error_message("push", bool(username) and not password)
                 ) from err
             if is_missing_upstream(err):
                 raise GitUpstreamException(
@@ -867,9 +863,7 @@ class Git:
         except CommandExitException as err:
             if is_auth_failure(err):
                 raise GitAuthException(
-                    build_auth_error_message(
-                        "pull", bool(username) and not password
-                    )
+                    build_auth_error_message("pull", bool(username) and not password)
                 ) from err
             if is_missing_upstream(err):
                 raise GitUpstreamException(
--- a/packages/js-sdk/tests/sandbox/git/config.test.ts
+++ b/packages/js-sdk/tests/sandbox/git/config.test.ts
@@ -61,24 +61,27 @@ sandboxTest('git setConfig updates local config', async ({ sandbox }) => {
   }
 })
 
-sandboxTest('git configureUser sets global user config', async ({ sandbox }) => {
-  await sandbox.git.configureUser(AUTHOR_NAME, AUTHOR_EMAIL)
+sandboxTest(
+  'git configureUser sets global user config',
+  async ({ sandbox }) => {
+    await sandbox.git.configureUser(AUTHOR_NAME, AUTHOR_EMAIL)
 
-  const name = (
-    await sandbox.commands.run('git config --global --get user.name')
-  ).stdout.trim()
-  const email = (
-    await sandbox.commands.run('git config --global --get user.email')
-  ).stdout.trim()
-  const configuredName = await sandbox.git.getConfig('user.name', {
-    scope: 'global',
-  })
-  const configuredEmail = await sandbox.git.getConfig('user.email', {
-    scope: 'global',
-  })
+    const name = (
+      await sandbox.commands.run('git config --global --get user.name')
+    ).stdout.trim()
+    const email = (
+      await sandbox.commands.run('git config --global --get user.email')
+    ).stdout.trim()
+    const configuredName = await sandbox.git.getConfig('user.name', {
+      scope: 'global',
+    })
+    const configuredEmail = await sandbox.git.getConfig('user.email', {
+      scope: 'global',
+    })
 
-  expect(name).toBe(AUTHOR_NAME)
-  expect(email).toBe(AUTHOR_EMAIL)
-  expect(configuredName).toBe(AUTHOR_NAME)
-  expect(configuredEmail).toBe(AUTHOR_EMAIL)
-})
+    expect(name).toBe(AUTHOR_NAME)
+    expect(email).toBe(AUTHOR_EMAIL)
+    expect(configuredName).toBe(AUTHOR_NAME)
+    expect(configuredEmail).toBe(AUTHOR_EMAIL)
+  }
+)
--- a/packages/js-sdk/tests/sandbox/git/remote.test.ts
+++ b/packages/js-sdk/tests/sandbox/git/remote.test.ts
@@ -8,17 +8,20 @@ import {
   startGitDaemon,
 } from './helpers.js'
 
-sandboxTest('git remoteGet returns undefined for missing remote', async ({ sandbox }) => {
-  const baseDir = await createBaseDir(sandbox)
+sandboxTest(
+  'git remoteGet returns undefined for missing remote',
+  async ({ sandbox }) => {
+    const baseDir = await createBaseDir(sandbox)
 
-  try {
-    const repoPath = await createRepo(sandbox, baseDir)
-    const missingUrl = await sandbox.git.remoteGet(repoPath, 'origin')
-    expect(missingUrl).toBeUndefined()
-  } finally {
-    await cleanupBaseDir(sandbox, baseDir)
+    try {
+      const repoPath = await createRepo(sandbox, baseDir)
+      const missingUrl = await sandbox.git.remoteGet(repoPath, 'origin')
+      expect(missingUrl).toBeUndefined()
+    } finally {
+      await cleanupBaseDir(sandbox, baseDir)
+    }
   }
-})
+)
 
 sandboxTest('git remoteAdd adds remote', async ({ sandbox }) => {
   const baseDir = await createBaseDir(sandbox)
--- a/packages/python-sdk/e2b/sandbox_sync/git.py
+++ b/packages/python-sdk/e2b/sandbox_sync/git.py
@@ -1,21 +1,21 @@
 from typing import Dict, List, Optional
 
 from e2b.sandbox.git_utils import (
-  GitBranches,
-  GitStatus,
-  build_auth_error_message,
-  build_git_command,
-  build_push_args,
-  build_upstream_error_message,
-  is_auth_failure,
-  is_missing_upstream,
-  parse_git_branches,
-  parse_git_status,
-  resolve_config_scope,
-  shell_escape,
-  strip_credentials,
-  derive_repo_dir_from_url,
-  with_credentials,
+    GitBranches,
+    GitStatus,
+    build_auth_error_message,
+    build_git_command,
+    build_push_args,
+    build_upstream_error_message,
+    is_auth_failure,
+    is_missing_upstream,
+    parse_git_branches,
+    parse_git_status,
+    resolve_config_scope,
+    shell_escape,
+    strip_credentials,
+    derive_repo_dir_from_url,
+    with_credentials,
 )
 from e2b.exceptions import (
     GitAuthException,
@@ -322,9 +322,7 @@ class Git:
         except CommandExitException as err:
             if is_auth_failure(err):
                 raise GitAuthException(
-                    build_auth_error_message(
-                        "clone", bool(username) and not password
-                    )
+                    build_auth_error_message("clone", bool(username) and not password)
                 ) from err
             raise
 
@@ -860,9 +858,7 @@ class Git:
         except CommandExitException as err:
             if is_auth_failure(err):
                 raise GitAuthException(
-                    build_auth_error_message(
-                        "push", bool(username) and not password
-                    )
+                    build_auth_error_message("push", bool(username) and not password)
                 ) from err
             if is_missing_upstream(err):
                 raise GitUpstreamException(
@@ -948,9 +944,7 @@ class Git:
         except CommandExitException as err:
             if is_auth_failure(err):
                 raise GitAuthException(
-                    build_auth_error_message(
-                        "pull", bool(username) and not password
-                    )
+                    build_auth_error_message("pull", bool(username) and not password)
                 ) from err
             if is_missing_upstream(err):
                 raise GitUpstreamException(


packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1299 @@
...
+    } = opts
+
+    const resolvedToken =
+      token ??
+      requestOpts.envs?.GITHUB_PAT ??
+      requestOpts.envs?.GITHUB_TOKEN ??
+      requestOpts.envs?.GH_TOKEN ??
+      process.env.GITHUB_PAT ??
+      process.env.GITHUB_TOKEN ??
+      process.env.GH_TOKEN
```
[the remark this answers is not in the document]
matthewlouisbrockman: fixed, there's a getEnvVar helper I guess that's for this sort of thing

.changeset/README.md
mishushakov, line 1: can remain unchanged

packages/python-sdk/pytest.ini
```diff
@@ -6,3 +6,6 @@ markers =
 asyncio_mode=auto
 addopts = "--import-mode=importlib"
 timeout = 300
+filterwarnings =
+    ignore:'asyncio\.iscoroutinefunction' is deprecated.*:DeprecationWarning:pytest_asyncio\.plugin
+    ignore:'asyncio\.get_event_loop_policy' is deprecated.*:DeprecationWarning:pytest_asyncio\.plugin
```
mishushakov, line 11: good idea, maybe we could ask @djeebus to fix it so we don't have to suppress the warning?

DEV.md
mishushakov, line 1: no need to change here, but let's draft another PR for this

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+  /**
+   * Remote name to use.
+   *
+   * @default "origin"
+   */
+  remoteName?: string
+  /**
+   * Use the SSH URL instead of the HTTPS clone URL when `true`.
+   */
+  useSsh?: boolean
```
mishushakov: let's maybe do protocol: 'ssh' | 'https'

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+  /**
+   * Fetch after adding the remote when `true`.
+   */
+  fetch?: boolean
+  /**
+   * Overwrite the remote if it already exists when `true`.
+   */
+  overwrite?: boolean
```
mishushakov: we can just do & GitRemoteAddOpts, no need to duplicate

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+   */
+  authorName?: string
+  /**
+   * Commit author email.
+   */
+  authorEmail?: string
+  /**
+   * Allow empty commits when `true`.
+   */
+  allowEmpty?: boolean
```
mishushakov, line 119: let's explicitly not allow this

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+      }
+
+      return result
+    }
+
+    if (password && !username) {
+      throw new InvalidArgumentError(
+        'Username is required when using a password or token for git clone.'
+      )
+    }
```
mishushakov: we should do the check early, before run any commands

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+   * Create a new GitHub repository (remote).
+   *
+   * When `addRemote` is provided, the created repository is added as a remote
+   * in an existing sandbox repository.
+   * It does not initialize a local repository.
+   *
+   * @param opts GitHub repository creation options.
+   * @returns Minimal information about the created repository.
+   */
+  async createGitHubRepo(opts: GitHubCreateRepoOpts): Promise<GitHubRepoInfo> {
```
mishushakov: let's remove this part, it's not our business to create GitHub repositories (there's GitHub SDK/MCP that handles that)

we should only handle git

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+        args.push(branch)
+      }
+      return args
+    }
+
+    if (password && !username) {
+      throw new InvalidArgumentError(
+        'Username is required when using a password or token for git push.'
+      )
+    }
```
mishushakov, line 695: let's check this early

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+   * Push commits to a remote.
+   *
+   * @param path Repository path.
+   * @param opts Push options.
+   * @returns Command result from the command runner.
+   */
+  async push(path: string, opts?: GitPushOpts): Promise<CommandResult> {
+    const { remote, branch, setUpstream, username, password, ...rest } =
+      opts ?? {}
+    const buildArgs = (remoteName?: string) => {
```
mishushakov: one thing to keep in mind when you do this is that it will create a new function in JS heap every time you call .push. you can use function buildArgs instead to declare a function inside the block

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+   * Set a git config value.
+   *
+   * Use `scope: "local"` together with `path` to configure a specific repository.
+   *
+   * @param key Git config key (for example, `"pull.rebase"`).
+   * @param value Git config value.
+   * @param opts Config options.
+   * @returns Command result from the command runner.
+   */
+  async configSet(
```
mishushakov: let's do setConfig

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+  /**
+   * Get a git config value.
+   *
+   * Returns `undefined` when the key is not set in the requested scope.
+   *
+   * @param key Git config key (for example, `"pull.rebase"`).
+   * @param opts Config options.
+   * @returns The config value if present.
+   */
+  async configGet(
```
mishushakov: getConfig

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+
+    const targetHost = (host ?? 'github.com').trim()
+    const targetProtocol = (protocol ?? 'https').trim()
+    const credentialInput = [
+      `protocol=${targetProtocol}`,
+      `host=${targetHost}`,
+      `username=${username}`,
+      `password=${password}`,
+      '',
+      '',
```
mishushakov, line 874: what are these two?

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+    name: string,
+    email: string,
+    opts?: GitRequestOpts
+  ): Promise<CommandResult> {
+    if (!name || !email) {
+      throw new InvalidArgumentError('Both name and email are required.')
+    }
+
+    await this.configSet('user.name', name, { ...opts, scope: 'global' })
+    return this.configSet('user.email', email, { ...opts, scope: 'global' })
```
mishushakov: maybe make scope configurable too?

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+    return this.commands.run(cmd, {
+      ...rest,
+      envs: mergedEnvs,
+    })
+  }
+
+  /**
+   * Execute a raw shell command while applying default git environment variables.
+   */
+  private async runShell(
```
mishushakov, line 942: the function looks identical to run to me?

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+  ): '--global' | '--local' | '--system' {
+    switch (scope) {
+      case 'global':
+        return '--global'
+      case 'system':
+        return '--system'
+      case 'local':
+        return '--local'
+    }
+  }
```
mishushakov: not sure we need this function, you can just check if scope is one of global, system, local and return `--{scope}`

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+      )
+    }
+
+    try {
+      return await this.run(buildArgs(), path, rest)
+    } catch (err) {
+      if (this.isAuthFailure(err)) {
+        throw new InvalidArgumentError(
+          this.buildAuthErrorMessage('push', Boolean(username) && !password)
+        )
```
mishushakov: I would throw git auth error instead of invalid argument, since the arguments were correct, only the response from upstream raised the mismatch

packages/js-sdk/tests/sandbox/git/checkoutBranch.test.ts
mishushakov, line 1: I'd just add this to branches test

packages/js-sdk/tests/sandbox/git/clone.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    const repoPath = await createRepoWithCommit(sandbox, baseDir)
+    const daemon = await startGitDaemon(sandbox, baseDir)
+    const clonePath = `${baseDir}/clone`
+
+    try {
+      await sandbox.git.remoteAdd(repoPath, 'origin', daemon.remoteUrl)
+      await sandbox.git.push(repoPath, {
+        remote: 'origin',
+        branch: 'main',
+        setUpstream: true,
```
mishushakov: I am thinking if we should always setUpstream: true, unless user explicitly says otherwise

packages/js-sdk/tests/sandbox/git/commit.test.ts
```diff
@@ -0,0 +1,32 @@
...
+  createRepo,
+} from './helpers.js'
+
+sandboxTest('git commit creates commit', async ({ sandbox }) => {
+  const baseDir = await createBaseDir(sandbox)
+
+  try {
+    const repoPath = await createRepo(sandbox, baseDir)
+    await sandbox.files.write(`${repoPath}/README.md`, 'hello\n')
+    await sandbox.git.add(repoPath, { all: true })
```
mishushakov: all: true should be the default

packages/js-sdk/tests/sandbox/git/setConfig.test.ts
mishushakov, line 1: let's also move this one into config.test.ts

packages/js-sdk/tests/sandbox/git/configureUser.test.ts
```diff
@@ -0,0 +1,18 @@
+import { expect } from 'vitest'
+
+import { sandboxTest } from '../../setup.js'
+import { AUTHOR_EMAIL, AUTHOR_NAME } from './helpers.js'
+
+sandboxTest('git configureUser', async ({ sandbox }) => {
+  await sandbox.git.configureUser(AUTHOR_NAME, AUTHOR_EMAIL)
+
+  const name = (
+    await sandbox.commands.run('git config --global --get user.name')
```
mishushakov: can't we configGet here?

packages/js-sdk/tests/sandbox/git/createGitHubRepo.test.ts
mishushakov, line 1: let's remove this

packages/js-sdk/tests/sandbox/git/dangerouslyAuthenticate.test.ts
```diff
@@ -0,0 +1,18 @@
...
+    username: USERNAME,
+    password: PASSWORD,
+    host: HOST,
+    protocol: PROTOCOL,
+  })
+
+  const helper = (
+    await sandbox.commands.run('git config --global --get credential.helper')
+  ).stdout.trim()
+  expect(helper).toBe('store')
```
mishushakov, line 20: should we check credentials instead?

packages/js-sdk/tests/sandbox/git/deleteBranch.test.ts
```diff
@@ -0,0 +1,26 @@
...
+  createBaseDir,
+  createRepoWithCommit,
+} from './helpers.js'
+
+sandboxTest('git deleteBranch removes branch', async ({ sandbox }) => {
+  const baseDir = await createBaseDir(sandbox)
+
+  try {
+    const repoPath = await createRepoWithCommit(sandbox, baseDir)
+    await sandbox.commands.run(`git -C "${repoPath}" branch feature`)
```
mishushakov: we can use our createBranch instead of git ops

packages/js-sdk/tests/sandbox/git/deleteBranch.test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    await sandbox.git.deleteBranch(repoPath, 'feature', { force: true })
+
+    const branch = (
+      await sandbox.commands.run(`git -C "${repoPath}" branch --list feature`)
```
mishushakov: same here

packages/js-sdk/tests/sandbox/git/status.test.ts
mishushakov, line 1: can we also add tests for other statuses?

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+  /**
+   * Add (or update) a remote for a repository.
+   *
+   * @param path Repository path.
+   * @param name Remote name (for example, `"origin"`).
+   * @param url Remote URL.
+   * @param opts Remote add options.
+   * @returns Command result from the command runner.
+   */
+  async remoteAdd(
```
mishushakov, line 398: let's do also remote-get

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+      args.push('--allow-empty')
+    }
+
+    const hasAuthorName = Boolean(authorName)
+    const hasAuthorEmail = Boolean(authorEmail)
+    if (hasAuthorName !== hasAuthorEmail) {
+      throw new InvalidArgumentError(
+        'Both authorName and authorEmail are required to set commit author.'
+      )
+    }
```
mishushakov: I took another look and I think that we should not require author name and author e-mail if they are set globally in the config

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+   * @param message Commit message.
+   * @param opts Commit options.
+   * @returns Command result from the command runner.
+   */
+  async commit(
+    path: string,
+    message: string,
+    opts?: GitCommitOpts
+  ): Promise<CommandResult> {
+    const { authorName, authorEmail, allowEmpty, ...rest } = opts ?? {}
```
mishushakov, line 573: one thing to keep in mind here and elsewhere:

for required parameters we always use arguments and options are _optional_

example:

```ts
commit(path, message, name, email, options?)
```

if name, email not required:

```ts
commit(path, message, options?)
```

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+  }
+
+  /**
+   * Stage files for commit.
+   *
+   * @param path Repository path.
+   * @param opts Add options.
+   * @returns Command result from the command runner.
+   */
+  async add(path: string, opts?: GitAddOpts): Promise<CommandResult> {
```
mishushakov, line 547: the default should be "all"

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1430 @@
...
+      const cloneUrl =
+        authUsername && authPassword
+          ? withCredentials(url, authUsername, authPassword)
+          : url
+      const sanitizedUrl = stripCredentials(cloneUrl)
+      const shouldStripCredentials =
+        !dangerouslyStoreCredentials && sanitizedUrl !== cloneUrl
+      const repoPath = shouldStripCredentials
+        ? (path ?? deriveRepoDirFromUrl(url))
+        : path
```
mishushakov: let's make this one look a bit cleaner

.changeset/README.md
matthewlouisbrockman: ah, you saying not to push a version or leave the changeset as it wrote it

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: wait why not? e.g. "trigger CI" commits

packages/python-sdk/pytest.ini
matthewlouisbrockman: it was just a ton of warning showing up in the logs when running tests wanted to not deal with but yes will ask

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: aight.

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: removed per above

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: sounds good, although are you saying u want opts renamed to options? i think like, commands/index.ts, sandbox/index.ts they also use opts. Here path/message are required, rest are optional.

.changeset/README.md
mishushakov: I meant the change of the command in the README can be reverted (most of us run pnpm changeset btw., actually)

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: sounds good, ive always used . but reasonable default

packages/js-sdk/tests/sandbox/git/createGitHubRepo.test.ts
matthewlouisbrockman: removed

packages/js-sdk/src/sandbox/git/index.ts
mishushakov: okay, nevermind then

packages/js-sdk/src/sandbox/git/index.ts
mishushakov: all good in that case - yes for naming, we keep opts then, this was just pseudocode for illustrative purposes

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: all it does special right now is drop in `{"GIT_TERMINAL_PROMPT": "0"}` to autofail on interactive responses instead of timing out or something, can add that to the docstring

matthewlouisbrockman: oh whoops

matthewlouisbrockman: renaming that run to runGit to distinguish between terminal op vs git operation

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: changed

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: alright

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: alright

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: https://git-scm.com/docs/git-credential
input/output format expects newline after args

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: done

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: changed to all=True as default

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: added earlier checks

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: sounds good, left function, simplified and stuck in exception if the scope is invalid (like python has it already)

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: added

packages/js-sdk/tests/sandbox/git/commit.test.ts
matthewlouisbrockman: done

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: aight, keeping eye on it.

.changeset/README.md
matthewlouisbrockman: righto, reverted.

on a fresh box does the pnpm changeset work? wonder if i installed a bad version somehow on setup (was discussion here in office trying to figure out why pnpx was failing - possibly was a setup issue on my end)

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: moved to a helper

packages/js-sdk/tests/sandbox/git/clone.test.ts
matthewlouisbrockman: default is true now

packages/js-sdk/tests/sandbox/git/status.test.ts
matthewlouisbrockman: added

packages/js-sdk/tests/sandbox/git/configureUser.test.ts
matthewlouisbrockman: (for all of these last ones) realizing on the function usage for the tests using the sandbox git functions to help test them are a valid set of tests - I'm not really opinionated but worth pointing out that uh

I'm testing that the operation sets the state as expected and pulling the actual state. If I were to use the functions to check the state, it would make sure the outputs agree but not necessarily that the actual state is correct (although would indicate the state's correct). I guess it's like... it reduces the chance of a false passing test because it compares against ground truth instead of against another function having a correct answer. maybe can have both testing methods to make sure both the ground truth + helpers work as expected

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: not sure ssh'll work nice anyway actually, pain to add the ssh keys to the box possibly although hmm

matthewlouisbrockman: trying to avoid sticking credentials on box but then ssh keys would give the box credentials

matthewlouisbrockman: debating just yeeting ssh as an option

packages/js-sdk/tests/sandbox/git/setConfig.test.ts
matthewlouisbrockman: done

packages/js-sdk/tests/sandbox/git/checkoutBranch.test.ts
matthewlouisbrockman: done

packages/js-sdk/src/sandbox/git/index.ts
mishushakov: we can add it later, i think. https-only for now would work for most cases. also if we have the secrets proxy could be tough with ssh

packages/js-sdk/tests/sandbox/git/dangerouslyAuthenticate.test.ts
matthewlouisbrockman: added second check on it

packages/js-sdk/tests/sandbox/git/deleteBranch.test.ts
matthewlouisbrockman: for this one prefer to use the run(`git -C`) to ensure it's changing the real state

packages/js-sdk/tests/sandbox/git/configureUser.test.ts
mishushakov: i see the point. we can leave as is using git commands to verify the result of the git methods we're trying to test in the SDK. but then let's standardize it this way everywhere

matthewlouisbrockman: added second one on the getters that use the sandbox.git and make sure they match the raw getters

packages/js-sdk/tests/sandbox/git/deleteBranch.test.ts
matthewlouisbrockman: for this one prefer to use the run(git -C) to ensure it's changing the real state and not reliant on the sandbox create working

.changeset/README.md
mishushakov: we should def do smth about it and mise or nix might sound like a reasonable idea.

we sorta started to have this via a Dockerfile but it's currently only used for generating SDK API client.

need to put some more thought into it though.

mishushakov: i think we can ask @djeebus here as well

packages/js-sdk/src/sandbox/git/index.ts
matthewlouisbrockman: updated the variable names to more cleanly reflect intent; cloneUrl => urlWithCreds, shouldStrip => strip. it just needs to know if there's secrets on the credentials befroe storing them somewhere.

.changeset/README.md
matthewlouisbrockman: ah, issue is pnpx is deprecated
```
https://github.com/orgs/pnpm/discussions/5590
pnpx deprecated
[4:51](https://e2b-team.slack.com/archives/D0ABWR53FMW/p1769647860718159)
https://github.com/pnpm/pnpm/pull/3652 actual deprecation in 2021
```

so wonder if i have a different version of something than y'all

matthewlouisbrockman: joe's approach is update package.json to 
```
    "changeset": "pnpm dlx @changesets/cli"
```

and then the instruction to pnpm changeset; that works for me, thinking new PR with it so it's clean

djeebus: I think there are a few issues here:
- `npx changeset` doesn't work because it tries to find the `pnpx` executable as defined in `packages.json`, which doesn't exist.
- changing the script to `pnpm dlx @changesets/cli` still doesn't work, because pnpm isn't installed into `node_modules/`, so it still fails. this is just a theory, but it feels right, as the error states "could not determine executable to run"
- running `pnpm changeset` works because pnpm understands that a script that starts with "pnpm" refers to itself, so everything resolves correctly.
I'm guessing that it works on your machine because you've installed pnpm globally. not sure why mise isn't installing a pnpx script, but since pnpm says that "pnpx" is deprecated, I see no reason to rely on it when the change is just a few characters and makes it run for everyone.

Suggestion:
- change the md to instruct users to use `pnpm` instead of `npx`
- change the script to execute `pnpm dlx` instead of `pnpx`

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1157 @@
...
+        'or pass remote and branch explicitly.'
+      )
+    }
+
+    return (
+      'Git pull failed because no upstream branch is configured. ' +
+      'Pass remote and branch explicitly, or set upstream once (push with { setUpstream: true } ' +
+      'or run: git branch --set-upstream-to=origin/<branch> <branch>).'
+    )
+  }
```
mishushakov: i think we can move some of this to utils, since they don't actually depend on the methods here

packages/js-sdk/src/sandbox/git/index.ts
```diff
@@ -0,0 +1,1157 @@
...
+    if (scope !== 'local') {
+      return undefined
+    }
+    if (!path) {
+      throw new InvalidArgumentError(
+        'A repository path is required when using scope "local".'
+      )
+    }
+    return path
+  }
```
mishushakov: same as i said above

packages/js-sdk/tests/sandbox/git/branches.test.ts
```diff
@@ -0,0 +1,82 @@
...
+)
+
+sandboxTest('git deleteBranch removes branch', async ({ sandbox }) => {
+  const baseDir = await createBaseDir(sandbox)
+
+  try {
+    const repoPath = await createRepoWithCommit(sandbox, baseDir)
+    await sandbox.commands.run(`git -C "${repoPath}" branch feature`)
+
+    await sandbox.git.deleteBranch(repoPath, 'feature', { force: true })
```
mishushakov: is force required here?

packages/js-sdk/tests/sandbox/git/clone.test.ts
mishushakov: since setUpstream is the default now, we don't need it here explicitly

packages/js-sdk/tests/sandbox/git/dangerouslyAuthenticate.test.ts
mishushakov: mind elaborating where?

packages/js-sdk/tests/sandbox/git/remoteGet.test.ts
mishushakov, line 1: let's combine the get/add into remote.test.ts

packages/python-sdk/e2b/sandbox_async/git.py
```diff
@@ -0,0 +1,1173 @@
...
+        return await self._commands.run(
+            cmd,
+            envs=merged_envs,
+            user=user,
+            cwd=cwd,
+            timeout=timeout,
+            request_timeout=request_timeout,
+        )
+
+    def _is_auth_failure(self, err: Exception) -> bool:
```
mishushakov: same as i said in JS, we can move these into utils unless they depend on anything here

packages/python-sdk/tests/shared/git/conftest.py
```diff
@@ -0,0 +1,74 @@
...
+from uuid import uuid4
+
+import pytest
+
+BASE_DIR = "/tmp/test-git"
+
+
+@pytest.fixture
+def git_sandbox(sandbox_factory):
+    return sandbox_factory(secure=False, timeout=10)
```
mishushakov: we always want secure

packages/python-sdk/tests/shared/git/test_add.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+
+
+@pytest.mark.skip_debug()
+def test_add_stages_files(git_sandbox, git_repo):
+    git_sandbox.files.write(f"{git_repo}/README.md", "hello\n")
+
+    git_sandbox.git.add(git_repo, all=True)
```
mishushakov, line 8: all should be always True

packages/python-sdk/tests/shared/git/test_branches.py
```diff
@@ -0,0 +1,52 @@
...
+    assert "feature" in branches.branches
+    assert branches.current_branch == "feature"
+
+
+@pytest.mark.skip_debug()
+def test_delete_branch_removes_branch(git_sandbox, git_repo_with_commit):
+    repo_path = git_repo_with_commit
+
+    git_sandbox.commands.run(f'git -C "{repo_path}" branch feature')
+    git_sandbox.git.delete_branch(repo_path, "feature", force=True)
```
mishushakov: force may not be needed (same in JS)

packages/python-sdk/tests/shared/git/test_clone.py
```diff
@@ -0,0 +1,22 @@
...
+    repo_path = git_repo_with_commit
+    remote_url = git_daemon["remote_url"]
+    clone_path = f"{git_base_dir}/clone"
+
+    git_sandbox.git.remote_add(repo_path, "origin", remote_url)
+    git_sandbox.git.push(
+        repo_path,
+        remote="origin",
+        branch="main",
+        set_upstream=True,
```
mishushakov, line 17: may not be needed

packages/python-sdk/tests/shared/git/test_commit.py
```diff
@@ -0,0 +1,44 @@
+import pytest
+
+
+@pytest.mark.skip_debug()
+def test_commit_creates_commit(git_sandbox, git_repo, git_author):
+    author_name, author_email = git_author
+
+    git_sandbox.files.write(f"{git_repo}/README.md", "hello\n")
+    git_sandbox.git.add(git_repo, all=True)
```
mishushakov, line 9: all true by default

packages/python-sdk/tests/shared/git/test_sync.py
```diff
@@ -0,0 +1,49 @@
...
+    repo_path = git_repo_with_commit
+    remote_url = git_daemon["remote_url"]
+    clone_path = f"{git_base_dir}/clone"
+
+    git_sandbox.git.remote_add(repo_path, "origin", remote_url)
+    git_sandbox.git.push(
+        repo_path,
+        remote="origin",
+        branch="main",
+        set_upstream=True,
```
mishushakov, line 51: same here

packages/python-sdk/tests/shared/git/test_sync.py
```diff
@@ -0,0 +1,49 @@
...
+    )
+    git_sandbox.git.clone(remote_url, clone_path)
+
+    git_sandbox.files.write(f"{repo_path}/README.md", "hello\nmore\n")
+    author_name, author_email = git_author
+    git_sandbox.git.add(repo_path, all=True)
```
mishushakov, line 57: and here

packages/python-sdk/tests/shared/git/test_remote_get.py
mishushakov, line 1: let's combine the two with add

DEV.md
mishushakov, line 1: please undo for now, we will figure it out at later point

.changeset/README.md
mishushakov: @djeebus @matthewlouisbrockman can we make a PR

packages/python-sdk/tests/shared/git/test_sync.py
matthewlouisbrockman: done

packages/python-sdk/tests/shared/git/test_sync.py
matthewlouisbrockman: done

packages/python-sdk/tests/shared/git/test_commit.py
matthewlouisbrockman: done

packages/python-sdk/tests/shared/git/test_clone.py
matthewlouisbrockman: done

packages/python-sdk/tests/shared/git/test_add.py
matthewlouisbrockman: done

packages/python-sdk/tests/shared/git/conftest.py
matthewlouisbrockman: removed False but i think the python guy already has `kwargs.setdefault("secure", False)`, been there 2 months

packages/python-sdk/tests/shared/git/test_branches.py
matthewlouisbrockman: rm'd

packages/js-sdk/tests/sandbox/git/branches.test.ts
matthewlouisbrockman: fair enough, nocommits yet so don't need to force delete it

DEV.md
matthewlouisbrockman: not sure how i did this yeah

packages/js-sdk/tests/sandbox/git/remoteGet.test.ts
matthewlouisbrockman: combined

packages/python-sdk/tests/shared/git/test_remote_get.py
matthewlouisbrockman: combined

PR #1106, opened 2026-01-29
feat: validate copy src paths are relative and within context directory

mishushakov:
Add path validation to the copy method in both JS and Python SDKs to ensure source paths are always relative and don't escape the context directory.

This prevents:
- Absolute paths like /absolute/whatever (Unix) or C:\whatever (Windows)
- Path traversal attacks like ../whatever or ./foo/../../../bar

The validation works cross-platform using Node's path.isAbsolute/normalize and Python's os.path.isabs/normpath plus PureWindowsPath for detecting Windows paths on Unix.

---

> [!NOTE]
> **Medium Risk**
> Changes behavior of `copy`/`copy_items` to throw earlier for previously-accepted absolute or escaping paths, which could break some consumers; logic is localized and well-covered by tests.
> 
> **Overview**
> Prevents path traversal in template `copy` operations by validating `src` is *relative* and does not escape the context directory (rejects absolute paths and `..`-based escapes) in both the JS and Python SDKs.
> 
> Updates `copyItems`/`copy_items` error handling to preserve the caller’s stack trace when validation fails, adds unit coverage for the new path validator plus new stack-trace tests for absolute-path failures, and ships as patch releases via a changeset.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c1a8eb978e3fd99fa829d571e811bb7ee18cd40b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/js-sdk/src/errors.ts
```diff
@@ -41,10 +41,13 @@ export class TimeoutError extends SandboxError {
...
  */
-export class InvalidArgumentError extends SandboxError {
+export class InvalidArgumentError extends Error {
   constructor(message: string, stackTrace?: string) {
-    super(message, stackTrace)
+    super(message)
     this.name = 'InvalidArgumentError'
+    if (stackTrace) {
+      this.stack = stackTrace
+    }
```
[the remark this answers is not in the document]
mishushakov: this is supposed to be

packages/js-sdk/src/template/utils.ts
```diff
@@ -37,7 +100,7 @@ export function normalizeBuildArguments(
   }
 
   if (!name || name.length === 0) {
-    throw new TemplateError('Name must be provided')
+    throw new InvalidArgumentError('Name must be provided')
```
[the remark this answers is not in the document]
mishushakov: it's not a template error because there's no template yet

packages/js-sdk/src/template/utils.ts
```diff
@@ -7,6 +7,69 @@ import { BASE_STEP_NAME, FINALIZE_STEP_NAME } from './consts'
...
+  // Check for absolute paths (Unix-style starting with / or Windows-style with drive letter)
+  // path.isAbsolute uses platform-specific rules, so we also check path.win32.isAbsolute
+  // to catch Windows absolute paths (C:\foo, \\server\share) when running on Unix
+  if (path.isAbsolute(src) || path.win32.isAbsolute(src)) {
+    const error = new TemplateError(
+      `Invalid source path "${src}": absolute paths are not allowed. Use a relative path within the context directory.`,
+      stackTrace
+    )
+    throw error
+  }
```
jakubno: I still don't understand, why do we check windows when on unix? Is it just because of tests?

packages/python-sdk/e2b/template/main.py
```diff
@@ -64,9 +65,25 @@ def copy(
         """
         srcs = [src] if isinstance(src, (str, Path)) else src
 
+        # Get the caller frame for stack trace in validation errors
+        caller_frame = get_caller_frame(STACK_TRACE_DEPTH - 1)
+        stack_trace = None
+        if caller_frame is not None:
+            stack_trace = TracebackType(
```
jakubno: we could have util for this, I think I saw this multiple times

mishushakov: you're right, that's our pattern in Python

packages/js-sdk/src/template/utils.ts
mishushakov: in this case it's because path.absolute will not report paths like C:\Whatever\Whatever as absolute on Mac/Linux, unless you also specifically use the windows implementation to check. in other cases, i think path.isAbsolute is same as path.win32.isAbsolute on Windows

Issue #1108, opened 2026-01-30
[Bug]: (python-sdk) AsyncSandbox.git is missing reset and restore functions


bxxf:
### Environment

e2b==2.13.0
macOS 26.2 Tahoe

### Timestamp of the issue

2026-01-29 (when git feature was introduced)

### Frequency

Happens every time

### Expected behavior

Not sure if this is on purpose - but I believe the git instance should work the same both for the Sync and Async sandboxes with all the functions.

### Actual behavior

`AsyncSandbox.git` is missing `reset()` and `restore()` methods that exist on Sandbox.git and in the TypeScript SDK. Calling either raises AttributeError.

### Issue reproduction

Just try to use `.git.reset()` or `.git.restore()` on AsyncSandbox instance.

```python
import asyncio
from e2b import AsyncSandbox

async def main():
    sandbox = await AsyncSandbox.create()
    await sandbox.git.clone("https://github.com/e2b-dev/E2B.git", "/home/user/E2B")
    await sandbox.files.write("/home/user/E2B/README.md", "changed\n")
    await sandbox.git.reset("/home/user/E2B", mode="hard", target="HEAD")
    # AttributeError: 'Git' object has no attribute 'reset'
    await sandbox.kill()

asyncio.run(main())
```
Error:
```
Traceback (most recent call last):
  File "example.py", line 7, in main
    await sandbox.git.reset("/home/user/E2B", mode="hard", target="HEAD")
          ^^^^^^^^^^^^^^^^^
AttributeError: 'Git' object has no attribute 'reset'
```

### Additional context

If there's any additional details why this should not work this way, please lmk - I can provide a PR with the fix that I use locally. Parity check in unit tests for checking if any features are not missing could be helpful, or potentially some codegen / shared logic to prevent differences between sync/async env in the future:

```python
import asyncio
import inspect

from e2b.sandbox_async.git import Git as AsyncGit
from e2b.sandbox_sync.git import Git as SyncGit


def _public_methods(cls):
    return {
        name: getattr(cls, name)
        for name in sorted(dir(cls))
        if not name.startswith("_") and callable(getattr(cls, name))
    }


def test_identical_method_signatures():
    sync = _public_methods(SyncGit)
    async_ = _public_methods(AsyncGit)

    assert set(sync) == set(async_), (
        f"missing from async: {set(sync) - set(async_)}, "
        f"missing from sync: {set(async_) - set(sync)}"
    )

    for name in sync:
[...]

PR #1109, opened 2026-01-30
fix(python-sdk): missing functions in AsyncSandbox.git

bxxf:
Directly related to https://github.com/e2b-dev/E2B/issues/1108. Please check the issue description for more details.

tldr: `.restore()` and .`reset()` functions are missing in the `AsyncSandbox` instance

also added test to check parity of methods/signatures between sync and async git instances

other potential concerns: 
- tests are only ran for the sync variant across the python-sdk - if there are any mismatches in the async logic (not only git i believe), the tests won't catch it

### DIFF
--- a/packages/python-sdk/tests/shared/git/test_parity.py
+++ b/packages/python-sdk/tests/shared/git/test_parity.py
@@ -1,0 +1,34 @@
+import asyncio
+import inspect
+
+from e2b.sandbox_async.git import Git as AsyncGit
+from e2b.sandbox_sync.git import Git as SyncGit
+
+
+def _public_methods(cls):
+    return {
+        name: getattr(cls, name)
+        for name in sorted(dir(cls))
+        if not name.startswith("_") and callable(getattr(cls, name))
+    }
+
+
+def test_identical_method_signatures():
+    sync = _public_methods(SyncGit)
+    async_ = _public_methods(AsyncGit)
+
+    assert set(sync) == set(async_), (
+        f"missing from async: {set(sync) - set(async_)}, "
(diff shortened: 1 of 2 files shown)


packages/python-sdk/tests/shared/git/test_parity.py
```diff
@@ -0,0 +1,34 @@
...
+    for name in sync:
+        assert inspect.signature(sync[name]) == inspect.signature(async_[name]), (
+            f"{name}: sync{inspect.signature(sync[name])} "
+            f"!= async{inspect.signature(async_[name])}"
+        )
+
+
+def test_async_methods_are_coroutines():
+    for name, method in _public_methods(AsyncGit).items():
+        assert asyncio.iscoroutinefunction(method), f"AsyncGit.{name} is not async"
```
matthewlouisbrockman, line 34: Getting `DeprecationWarning: 'asyncio.iscoroutinefunction' is deprecated and slated for removal in Python 3.16; use inspect.iscoroutinefunction() instead` down to use the inspect function instead?

matthewlouisbrockman (approved):
trying to figure out right way to do this without ci/cd blocking; as a fix for now merging into a branch for ci/cd, adding the extra consolidation, and PRing from there.

packages/python-sdk/tests/shared/git/test_parity.py
bxxf: Thanks for flagging this! I wasn’t available at the time, but I see you’ve already resolved it - turns out my local dev env was using an older Python binary.

PR #1110, opened 2026-01-30
fix: Improve sandbox.git async/sync parity

matthewlouisbrockman:
Fixes issue with missing `restore` and `reset` functionality on the async git sandboxes.
 
Consolidate shared git helpers by moving remote URL argument construction and
parsing into the git utilities package. Sync and async git modules now reuse
the same builders where possible, with tests to ensure no drift.


packages/python-sdk/e2b/sandbox/git_utils.py
mishushakov, line 1: I think we don't need to re-export it here as they're already in __init__.py

packages/python-sdk/tests/shared/git/test_parity.py
mishushakov, line 1: can we do this test all of our classes? like Sandbox, SandboxAsync, Template, TemplateAsync

packages/python-sdk/e2b/sandbox/git_utils.py
matthewlouisbrockman: rmd

PR #1119, opened 2026-02-06
fix: resolve ty no-matching-overload on Sandbox.kill()

beran-t:
## Summary

- Fixes `ty` type checker reporting `error[no-matching-overload]` when calling `Sandbox.kill()` (and all other methods using the `class_method_variant` pattern)
- Single 2-line change: make `class_method_variant` inherit from `Generic[T]` instead of `object`

## Problem

The `class_method_variant` descriptor uses `cast(T, self)` to tell type checkers that the decorator preserves the original function's type. Without `Generic[T]`, `T` is only a method-level TypeVar — `ty` doesn't trust the cast and fails to resolve overloads at call sites. `mypy` and `pyright` are more lenient and accept it either way.

Affected methods (both `Sandbox` and `AsyncSandbox`): `kill`, `connect`, `set_timeout`, `get_info`, `get_metrics`, `beta_pause`.

## Fix

Adding `Generic[T]` makes `T` a class-level type parameter, so `ty` can track the type binding through the descriptor (`class_method_variant[(self, **opts) -> bool]`). The cast then makes sense to all three type checkers.

## Verification

Tested with a consumer repro (`sandbox.kill()`) against:

| Type Checker | Before | After |
|---|---|---|
| ty 0.0.15 | `error[no-matching-overload]` | All checks passed |
| mypy 1.19.1 | All checks passed | All checks passed |
| pyright 1.1.408 | All checks passed | All checks passed |

## Test plan

- [x] Verified `ty check` passes on consumer-side repro
- [x] Verified `mypy` and `pyright` still pass (no regressions)
- [x] Verified Python syntax is valid
- [x] No runtime behavior change (`Generic[T]` only affects type-level metadata)
- [ ] Existing SDK tests (require E2B API credentials)

### DIFF
--- a/.changeset/every-feet-attack.md
+++ b/.changeset/every-feet-attack.md
@@ -1,6 +1,0 @@
----
-'@e2b/python-sdk': patch
-'e2b': patch
----
-
-fix: validate copy src paths are relative and within context directory
--- a/.changeset/fix-empty-error-body-handling.md
+++ b/.changeset/fix-empty-error-body-handling.md
@@ -1,5 +1,0 @@
----
-"e2b": patch
----
-
-Fix error handling for API responses with empty body.
--- a/.changeset/fix-team-id-env-override.md
+++ b/.changeset/fix-team-id-env-override.md
@@ -1,0 +1,5 @@
+---
+"@e2b/cli": patch
+---
+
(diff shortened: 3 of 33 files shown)


mishushakov:
add changeset and merge/release please

PR #1132, opened 2026-02-14
feat: add getTags/get_tags to list all tags for a template

beran-t:
## Summary
- Add `GET /templates/{templateID}/tags` endpoint to the OpenAPI spec
- Add `Template.getTags()` to JS/TS SDK
- Add `Template.get_tags()` (sync) and `AsyncTemplate.get_tags()` (async) to Python SDK
- Returns a list of `TemplateTag` objects with `tag`, `buildId`, and `createdAt` fields

## Test plan
- Added unit tests for JS SDK (`Template.getTags` happy path + 404 error)
- Added unit tests for Python SDK (sync + async, happy path + error)

### DIFF
--- a/packages/js-sdk/src/api/index.ts
+++ b/packages/js-sdk/src/api/index.ts
@@ -74,7 +74,8 @@ class ApiClient {
 
     this.api = createClient<paths>({
       baseUrl: config.apiUrl,
-      // keepalive: true, // TODO: Return keepalive
+      // In HTTP 1.1, all connections are considered persistent unless declared otherwise
+      // keepalive: true,
       headers: {
         ...defaultHeaders,
         ...(config.apiKey && { 'X-API-KEY': config.apiKey }),
--- a/packages/js-sdk/src/envd/api.ts
+++ b/packages/js-sdk/src/envd/api.ts
@@ -113,7 +113,8 @@ class EnvdApiClient {
       baseUrl: config.apiUrl,
       fetch: config?.fetch,
       headers: config?.headers,
-      // keepalive: true, // TODO: Return keepalive
+      // In HTTP 1.1, all connections are considered persistent unless declared otherwise
+      // keepalive: true,
     })
     this.version = metadata.version
 


packages/js-sdk/src/template/index.ts
```diff
@@ -1279,5 +1306,6 @@ export type {
   TemplateBuildStatus,
   TemplateBuildStatusResponse,
   TemplateClass,
+  TemplateTag,
```
mishushakov, line 1309: you need to export this from index.ts as well

packages/python-sdk/e2b/template/types.py
```diff
@@ -73,6 +74,22 @@ class TemplateTagInfo:
...
+    """
+
+    tag: str
+    """Name of the tag."""
+
+    build_id: str
+    """Build identifier associated with this tag."""
+
+    created_at: datetime
+    """When this tag was assigned."""
```
mishushakov, line 90: you probably don't need this class

packages/python-sdk/e2b/template_async/build_api.py
```diff
@@ -336,3 +338,37 @@ async def remove_tags(client: AuthenticatedClient, name: str, tags: List[str]) -
...
+        raise TemplateException("Failed to get template tags")
+
+    return [
+        TemplateTag(
+            tag=item.tag,
+            build_id=str(item.build_id),
+            created_at=item.created_at,
+        )
+        for item in res.parsed
+    ]
```
mishushakov, line 374: it is safe to return res.parsed here, as Python will be returning a correctly parsed response

dobrac: note, we want to retype all responses to a custom - SDK defined types to not expose/export the generated openapi spec types

mishushakov: @dobrac why that?

dobrac: the reason is to not expose the openapi types as they might have different field names than we want, or contain additional fields that we might not want to expose to the users.<br><br>We're doing this for all types

packages/js-sdk/src/template/index.ts
beran-t: In index.ts there is 
94: `export * from './template'`

Which probably takes care of this? Everything works for me when testing it.
```
import pkg from './packages/js-sdk/dist/index.js'
const { Template, TemplateTag } = pkg

console.log('TemplateTag:', TemplateTag)

const tags = await Template.getTags('w8h2fkcozzr3bz100iyl')
console.log(`Found ${tags.length} tags:`)
for (const tag of tags) {
  console.log(`  Tag: ${tag.tag}, Build: ${tag.buildId}, Created: ${tag.createdAt}`)
}
```

PR #1152, opened 2026-02-23
fix(python-sdk): apply read/write timeout to server-stream requests

### LINKED ISSUE #1128: [Bug]: commands.run hangs indefinitely when sandbox becomes unreachable — request_timeout does not set read timeout on streaming calls
### Sandbox ID or Build ID

_No response_

### Environment

e2b==2.8.1
python-sdk

### Timestamp of the issue

2026-02-11 15:00 PST

### Frequency

One-time occurrence

### Expected behavior

if sandbox is unreachable, commands.run should error out

### Actual behavior

if sandbox is unreachable, commands.run hangs forever

### Issue reproduction

1. Create a sandbox and run some commands successfully
2. Wait for the sandbox to become unreachable (e.g., sandbox times out or dies)
3. Call commands.run() — it hangs indefinitely because there is no read timeout on the HTTP stream

### Additional context

There was a blip in e2b of some sort and we saw couple of our sandboxes die or timeout at the same time that a long running command had been issued; the command was retried automatically with the commands.run, but then all of those tasks in our system hung indefinitely getting stuck at `commands.run`.

My (claude's) best guess:
When a sandbox becomes unreachable (e.g., sandbox dies or network connectivity is lost), commands.run() hangs indefinitely because the underlying call_server_stream does not set a read timeout on the HTTP stream. The request_timeout parameter only sets connect and pool timeouts for streaming calls, unlike unary calls which correctly set all four (connect, pool, read, write).

In e2b_connect/client.py, _prepare_unary_request (used by files.make_dir, files.write, etc.) correctly sets all timeouts:
```
  extensions = {
      "timeout": {
          "connect": request_timeout,
          "pool": request_timeout,
          "read": request_timeout,
          "write": request_timeout,
      }
  }
```

But _prepare_server_stream_request (used by commands.run) omits read and write:
```
  extensions = {
      "timeout": {"connect": request_timeout, "pool": request_timeout}
  }
```

The timeout parameter is sent as a connect-timeout-ms header for server-side enforcement, but if the server is unreachable, the client never gets a response and httpcore's ConnectionPool.stream() blocks on read forever.

VintLin:
## Summary
- `commands.run()` can hang indefinitely on server-stream calls when the sandbox becomes unreachable.
- Root cause: `_prepare_server_stream_request` only applied `connect`/`pool` timeout, but did not apply `read`/`write`.
- This PR applies `read`/`write` timeout as well, aligning stream behavior with unary request timeout handling.

## Changes
- Update `packages/python-sdk/e2b_connect/client.py`:
  - in `_prepare_server_stream_request`, when `request_timeout` is set, include:
    - `connect`
    - `pool`
    - `read`
    - `write`
- Add focused tests in `packages/python-sdk/tests/e2b_connect/test_client.py`:
  - `test_server_stream_timeout_extensions_include_read_write`
  - `test_server_stream_timeout_extensions_none_when_timeout_unset`

## Minimal reproduction
A stream request path using `commands.run()` can block when the stream becomes unresponsive if only connect/pool timeouts are configured.
With this change, stream read/write phases now honor `request_timeout`, so the call exits instead of hanging.

## Validation
- [x] `pytest tests/e2b_connect/test_client.py -q`
- [x] `ruff check e2b_connect/client.py tests/e2b_connect/test_client.py`

## Scope / Risk
- Small, targeted fix (2 files changed in the final diff).
- No API surface change.
- Risk is low: only timeout extension fields for server-stream requests are aligned with existing unary timeout semantics.

Fixes #1128

### DIFF
--- a/DEV.md
+++ b/DEV.md
@@ -1,3 +1,3 @@
 # Releasing e2b cli
 
-to create a changeset run `npx changeset`
+to create a changeset run `pnpm run changeset`
--- a/apps/web/src/app/(docs)/docs/sdk-reference/cli/v2.8.1/auth/page.mdx
+++ b/apps/web/src/app/(docs)/docs/sdk-reference/cli/v2.8.1/auth/page.mdx
@@ -1,0 +1,58 @@
+## e2b auth
+
+
+authentication commands
+
+### Usage
+
+```bash
+e2b auth [options] [command]
+```
+## e2b auth login
+
+
+log in to CLI
+
+### Usage
+
+```bash
+e2b auth login [options]
+```
+
+
+## e2b auth logout
+
+
+log out of CLI
+
+### Usage
+
+```bash
+e2b auth logout [options]
+```
+
+
+## e2b auth info
+
+
+get information about the current user
+
+### Usage
+
+```bash
+e2b auth info [options]
+```
+
+
+## e2b auth configure
+
+
+configure user
+
+### Usage
+
+```bash
+e2b auth configure [options]
+```
+
+
--- a/apps/web/src/app/(docs)/docs/sdk-reference/cli/v2.8.1/sandbox/page.mdx
+++ b/apps/web/src/app/(docs)/docs/sdk-reference/cli/v2.8.1/sandbox/page.mdx
@@ -1,0 +1,181 @@
+## e2b sandbox
+
+
+work with sandboxes
+
+### Usage
+
+```bash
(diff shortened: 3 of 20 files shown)


ValentaTomas:
Extracted and modified in https://github.com/e2b-dev/E2B/pull/1197 as we need to take care of the specifics of the server streaming timeout.

PR #1156, opened 2026-02-25
fix(js-sdk): expand ENV variables and strip quotes in fromDockerfile()

beran-t:
## Summary

`fromDockerfile()` parses `ENV` instructions but stores values as raw strings without stripping quotes. This causes `ENV GOPATH="/go"` to produce `"/go"` (with literal quote characters) instead of `/go`.

Variable expansion (e.g. `${GOPATH}`) is handled by the backend, so the SDK only needs to strip quotes.

### Changes

- Add `stripQuotes()` helper — removes surrounding `"` or `'` from ENV values
- Apply `stripQuotes()` to all value extraction points in `handleEnvInstruction()`
- Add unit tests for quote stripping behavior

### Before/After

| Dockerfile | Before (broken) | After (fixed) |
|---|---|---|
| `ENV GOPATH="/go"` | `"/go"` | `/go` |
| `ENV FOO='bar'` | `'bar'` | `bar` |
| `ENV PLAIN=hello` | `hello` | `hello` (unchanged) |

Only `packages/js-sdk/src/template/dockerfileParser.ts` is modified.

## Test plan

- [x] Unit tests for quote stripping (double, single, mismatched, unquoted, multi-pair)
- [x] Integration test confirmed backend handles variable expansion correctly
- [x] Existing `fromDockerfile.test.ts` tests pass
- [x] `pnpm run format`, `pnpm run lint`, `pnpm run typecheck` all pass

### DIFF
--- a/packages/js-sdk/src/template/dockerfileParser.ts
+++ b/packages/js-sdk/src/template/dockerfileParser.ts
@@ -97,6 +97,9 @@ export function parseDockerfile(
   templateBuilder.setUser('root')
   templateBuilder.setWorkdir('/')
 
+  // Track resolved ENV values for variable expansion across instructions
+  const envContext: Record<string, string> = {}
+
   // Process all other instructions
   for (const instruction of instructions) {
     const keyword = instruction.getKeyword()
@@ -130,7 +133,7 @@ export function parseDockerfile(
 
       case 'ENV':
       case 'ARG':
-        handleEnvInstruction(instruction, templateBuilder)
+        handleEnvInstruction(instruction, templateBuilder, envContext)
         break
 
       case 'EXPOSE':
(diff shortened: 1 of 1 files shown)


packages/js-sdk/src/template/dockerfileParser.ts
```diff
@@ -220,9 +223,79 @@ function handleUserInstruction(
...
+  )
+  // Handle $VAR
+  result = result.replace(
+    /\$([a-zA-Z_][a-zA-Z0-9_]*)/g,
+    (_, name) => {
+      return envContext[name] !== undefined ? envContext[name] : ''
+    }
+  )
+  return result
+}
```
mishushakov, line 282: I think we do this on the backend already, probably just stripping quotes is enough - can you test?

packages/js-sdk/src/template/dockerfileParser.ts
```diff
@@ -97,6 +97,9 @@ export function parseDockerfile(
   templateBuilder.setUser('root')
   templateBuilder.setWorkdir('/')
 
+  // Track resolved ENV values for variable expansion across instructions
+  const envContext: Record<string, string> = {}
```
mishushakov, line 101: not required

PR #1157, opened 2026-02-25
fix(js-sdk): return from iterateEvents on end instead of break

beran-t:
## Summary
- Replace `break` with `return` in the `"end"` case of `CommandHandle.iterateEvents()` so the async generator exits immediately once the command result is captured, instead of continuing to wait for the transport-level gRPC stream to close.

## Test plan
- [x] `pnpm run typecheck` passes
- [x] `pnpm run lint` passes
- [x] `pnpm run format` passes (no changes)


mishushakov:
Unnecessary, closing this.

PR #1176, opened 2026-03-03
fix(js-sdk): strip quotes from ENV values in fromDockerfile()

beran-t:
## Summary

`fromDockerfile()` parses `ENV` instructions but stores values as raw strings without stripping quotes. This causes `ENV GOPATH="/go"` to produce `"/go"` (with literal quote characters) instead of `/go`.

Variable expansion (e.g. `${GOPATH}`) is handled by the backend, so the SDK only needs to strip quotes.

### Changes

- Add `stripQuotes()` helper — removes surrounding `"` or `'` from ENV values
- Apply `stripQuotes()` to all value extraction points in `handleEnvInstruction()`
- Add unit tests for quote stripping behavior

### Before/After (tested)

| Test | Input | Output |
|---|---|---|
| strips double quotes | `ENV GOPATH="/go"` | `/go` |
| strips single quotes | `ENV GOPATH='/go'` | `/go` |
| does not strip mismatched quotes | `ENV GOPATH="/go'` | `"/go'` (unchanged) |
| handles unquoted values | `ENV GOPATH=/go` | `/go` (unchanged) |
| preserves variable references as-is | `ENV PATH="/usr/bin:${GOPATH}/bin"` | `/usr/bin:${GOPATH}/bin` |
| strips quotes from multiple pairs | `ENV A="hello" B="world"` | `hello`, `world` |
| ARG with default value | `ARG MY_ARG="hello"` | `hello` |
| ARG without default | `ARG MY_ARG` | `""` |

Only `packages/js-sdk/src/template/dockerfileParser.ts` is modified (+97/-4 lines).

## Test plan

- [x] Unit tests for quote stripping (double, single, mismatched, unquoted, multi-pair)
- [x] Integration build test: Dockerfile with `ENV GOPATH="/go"` + `RUN test "$GOPATH" = "/go"` — build succeeded, confirming quotes are stripped and backend expands variables
- [x] Existing `fromDockerfile.test.ts` tests pass
- [x] `pnpm run format`, `pnpm run lint`, `pnpm run typecheck` all pass

### DIFF
--- a/packages/js-sdk/src/template/dockerfileParser.ts
+++ b/packages/js-sdk/src/template/dockerfileParser.ts
@@ -220,6 +220,22 @@ function handleUserInstruction(
   }
 }
 
+/**
+ * Strip surrounding double or single quotes from an ENV value.
+ * Docker's ENV instruction strips quotes from values like ENV KEY="value".
+ */
+function stripQuotes(value: string): string {
+  if (value.length >= 2) {
+    if (
+      (value[0] === '"' && value[value.length - 1] === '"') ||
+      (value[0] === "'" && value[value.length - 1] === "'")
+    ) {
+      return value.slice(1, -1)
+    }
+  }
+  return value
+}
+
 function handleEnvInstruction(
   instruction: DockerfileInstruction,
   templateBuilder: DockerfileParserInterface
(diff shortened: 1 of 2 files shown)


mishushakov (approved):

mishushakov:
@claude does it need to be changed in Python as well?

mishushakov:
I will re-do this one in a new Docker parser/converter.

PR #1183, opened 2026-03-07
fix: add read/write timeouts to streaming requests

### LINKED ISSUE #1128: [Bug]: commands.run hangs indefinitely when sandbox becomes unreachable — request_timeout does not set read timeout on streaming calls

MaxwellCalkin:
## Summary

Fixes #1128

`_prepare_server_stream_request` only sets `connect` and `pool` timeouts when `request_timeout` is provided:

```python
# Before (line 328)
{"timeout": {"connect": request_timeout, "pool": request_timeout}}
```

But `_prepare_unary_request` correctly sets all four:

```python
# Unary (line 222-228)
{"timeout": {"connect": request_timeout, "pool": request_timeout, "read": request_timeout, "write": request_timeout}}
```

When a sandbox becomes unreachable during a streaming call (e.g. `commands.run()`), the client hangs indefinitely because httpcore's `ConnectionPool.stream()` blocks on read with no timeout set. The `connect-timeout-ms` header only enforces server-side timeout, not client-side read timeout.

**Fix:** Add `"read": request_timeout, "write": request_timeout` to the streaming extensions to match unary behavior.

```python
# After
{"timeout": {"connect": request_timeout, "pool": request_timeout, "read": request_timeout, "write": request_timeout}}
```

## Test plan
- [ ] Verify `commands.run()` with `request_timeout` raises an error instead of hanging when sandbox is unreachable
- [ ] Verify normal streaming operations still complete successfully with adequate timeout

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/.changeset/fix-streaming-read-write-timeout.md
+++ b/.changeset/fix-streaming-read-write-timeout.md
@@ -1,0 +1,7 @@
+---
+"e2b": patch
+---
+
+fix: add read/write timeouts to streaming requests to prevent indefinite hangs
+
+Add read and write timeouts to _prepare_server_stream_request to match _prepare_unary_request, preventing commands.run() from hanging indefinitely when a sandbox becomes unreachable.


MaxwellCalkin:
Closing as duplicate of #1194, which has the same fix with a clearer description. Apologies for the duplicate.

PR #1216, opened 2026-03-19
fix(sandbox): enhance pause method to merge connection config with options

### LINKED ISSUE #1215: [Bug]: Passed Credential in Sandbox.connect, does not get used in `.pause` method like in `.kill`
Version of JS SDK : 2.14.1

When making use of the SDK you can connect to an existing sandbox using the connect method and passing in an API key.

This allows for the use of the `.kill()` command normally, but `.pause` does not work and requires the apiKey field passed in again.

```ts
const mySandbox = Sandbox.connect(externalId, {
      apiKey,
    })

mySandbox.kill() // Works as should

mySandbox.pause() // Fails with auth issue, only works when apiKey is passed as a param
```

It would be cool to have pause work without passing in the apiKey again like `.kill`.

Please close the Issue and my PR if this is desired behaviour, I made this issue as a user, it was confusing and decided to go ahead and make a fix. Thank you!

OpeOginni:
Updated the `pause` method in the `Sandbox` class to merge the existing `connectionConfig` with the provided options (`opts`). This change ensures that connection settings are consistently applied when pausing the sandbox, improving flexibility and usability. Just like it is for the `kill` method.

Fixes #1215

Please ignore and close if this behaves like so on purpose.

### DIFF
--- a/packages/js-sdk/src/sandbox/index.ts
+++ b/packages/js-sdk/src/sandbox/index.ts
@@ -585,7 +585,10 @@ export class Sandbox extends SandboxApi {
    * ```
    */
   async pause(opts?: ConnectionOpts): Promise<boolean> {
-    return await SandboxApi.pause(this.sandboxId, opts)
+    return await SandboxApi.pause(this.sandboxId, {
+      ...this.connectionConfig,
+      ...opts,
+    })
   }
 
   /**


mishushakov:
Hey, just fyi. we're already working on a fix here:
https://github.com/e2b-dev/E2B/pull/1179

PR #1229, opened 2026-03-23
fix: disable stream timeout for background commands to allow reconnect by PID

### LINKED ISSUE #1074: Reconnect to running command by PID fails after 2 minutes
Using E2B Cloud Service: 

When trying to reconnect to a running command in the sandbox using its PID after 2 minutes, the command cannot be found. The reconnect functionality using Commands.connect(pid) does not work consistently, causing running background commands to appear lost.

To Reproduce

Start a long-running command in the sandbox using Commands.run(cmd, { background: true }).

Store the returned CommandHandle PID.

Wait for 2 minutes (or more).

Attempt to reconnect using Commands.connect(pid) with the stored PID.

Observe that the connection fails and the PID is reported as not found.

Expected behavior
I should be able to reconnect to any running command by its PID at any time, even after several minutes, and continue receiving stdout/stderr streams or send stdin.

Browser console output
[error] process with PID 12345 not found

Terminal commands & output
Example code used:
const handle = await sandbox.commands.run("long-running-script.sh", { background: true });
console.log("PID:", handle.pid);

// After 2 minutes
await sandbox.commands.connect(handle.pid);

Output:
Error: process with PID 12345 not found

Screenshots
N/A

Desktop (please complete the following information):

OS: Windows 11

Browser: Chrome 143

Version: 143.0.0.0

Smartphone (please complete the following information):

Device: N/A

Additional context
This bug breaks reconnecting to background commands, making it impossible to stream logs or interact with long-running tasks. It seems like the sandbox may be cleaning up the PID after 2 minutes, even though the command is still running.

majiayu000:
Fixes #1074

## Summary

`Commands.run()` passes a default 60-second timeout to the internal `start()` RPC call. This timeout sets a Connect-RPC deadline on the server-streaming call. For background commands, when the deadline expires the stream terminates with `DeadlineExceeded`, which signals envd to clean up its process tracking — making the PID unreachable for subsequent `connect(pid)` calls.

This fix changes the default timeout for background commands to `0` (unlimited) while preserving the 60-second default for foreground commands:

- **JS SDK**: When `background` is true and `timeoutMs` is not explicitly provided, override to `0` before calling `start()`.
- **Python SDK** (sync + async): Change the `timeout` parameter default from `60` to `None` (sentinel), then resolve to `0` for background commands or `60` for foreground commands at runtime.

Users who explicitly pass a timeout value still get that exact timeout regardless of background mode.

## Test plan

- Verified `pnpm run typecheck`, `pnpm run lint`, and `pnpm run format` all pass
- Existing connect tests continue to pass
- Manual verification: start a background command, wait >60s, then `connect(pid)` succeeds

### DIFF
--- a/packages/python-sdk/e2b/sandbox_async/commands/command.py
+++ b/packages/python-sdk/e2b/sandbox_async/commands/command.py
@@ -144,7 +144,7 @@ class Commands:
         on_stdout: Optional[OutputHandler[Stdout]] = None,
         on_stderr: Optional[OutputHandler[Stderr]] = None,
         stdin: Optional[bool] = None,
-        timeout: Optional[float] = 60,
+        timeout: Optional[float] = None,
         request_timeout: Optional[float] = None,
     ) -> CommandResult:
         """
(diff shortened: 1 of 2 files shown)


mishushakov:
This would be a breaking change - we have a better implementation coming soon. Thanks 🙏

PR #1261, opened 2026-04-11
fix: await async callbacks in CommandHandle.wait()

### LINKED ISSUE #1259: JS SDK `wait()` resolves before async stdout/stderr/PTY callbacks finish
## Bug Description

`CommandHandle` in the JS SDK advertises `onStdout`, `onStderr`, and `onPty` callbacks that may return `Promise<void>`, but `handleEvents()` invokes them **without** awaiting. As a result, `wait()` only waits for the event stream to end, not for user-provided async handlers to finish.

Any state those handlers persist can still be in flight after `await handle.wait()` returns, and rejected callback promises bypass `iterationError` because they are detached from the `try/catch`.

## Root Cause

`packages/js-sdk/src/sandbox/commands/commandHandle.ts` L149-164:

```typescript
constructor(
  readonly pid: number,
  ...
  private readonly onStdout?: (stdout: string) => void | Promise<void>,
  private readonly onStderr?: (stderr: string) => void | Promise<void>,
  private readonly onPty?: (pty: Uint8Array) => void | Promise<void>
) {
  this._wait = this.handleEvents()
}

// In handleEvents():
if (stdout !== null) {
  this.onStdout?.(stdout)      // ← not awaited, even though it can return Promise
} else if (stderr !== null) {
  this.onStderr?.(stderr)      // ← not awaited
} else if (pty) {
  this.onPty?.(pty)            // ← not awaited
}
```

## Steps to Reproduce

```typescript
const handle = await sandbox.commands.run("echo hello", {
  onStdout: async (data) => {
    await saveToDatabase(data); // async operation
  }
});
await handle.wait();
// At this point, saveToDatabase() may still be running!
```

## Expected Behavior

`wait()` should resolve only after all async callbacks have completed, matching the Python async SDK's `inspect.isawaitable` handling.

## Suggested Fix

Await callback results when they return promises:

```typescript
if (stdout !== null) {
  await this.onStdout?.(stdout)
}
```

This keeps `wait()` aligned with the documented callback types and ensures callback failures surface through `wait()` instead of becoming unhandled rejections.

YizukiAme:
Fixes #1259

## Problem

The `CommandHandle.wait()` method fires `onStdout`, `onStderr`, and `onPty` callbacks without `await`, so async callbacks run as fire-and-forget microtasks. If a callback performs I/O (e.g. writing to a file, sending over network), `wait()` can resolve before the callback finishes, leading to lost data or race conditions.

## Fix

Add `await` before each optional-chain callback invocation:

```diff
-this.onStdout?.(stdout)
+await this.onStdout?.(stdout)
```

This is fully backwards compatible — `await`-ing a sync function's return value is a no-op.

## Tests

3 parameterized test cases (`stdout`, `stderr`, `pty`) verify that `wait()` does not resolve until an async callback's promise settles.

_This fix was developed with AI assistance and reviewed by a human._

### DIFF
--- a/.changeset/await-async-callbacks.md
+++ b/.changeset/await-async-callbacks.md
@@ -1,0 +1,5 @@
+---
+"e2b": patch
+---
+
+fix: await async callbacks in CommandHandle.wait()
--- a/packages/js-sdk/src/sandbox/commands/commandHandle.ts
+++ b/packages/js-sdk/src/sandbox/commands/commandHandle.ts
@@ -226,11 +226,11 @@ export class CommandHandle
     try {
       for await (const [stdout, stderr, pty] of this.iterateEvents()) {
         if (stdout !== null) {
-          this.onStdout?.(stdout)
+          await this.onStdout?.(stdout)
         } else if (stderr !== null) {
-          this.onStderr?.(stderr)
+          await this.onStderr?.(stderr)
         } else if (pty) {
-          this.onPty?.(pty)
+          await this.onPty?.(pty)
         }
       }
     } catch (e) {
--- a/packages/js-sdk/tests/sandbox/commands/commandHandle.test.ts
+++ b/packages/js-sdk/tests/sandbox/commands/commandHandle.test.ts
@@ -1,0 +1,97 @@
+import { describe, expect, it, vi } from 'vitest'
+
+import { CommandHandle } from '../../../src/sandbox/commands/commandHandle'
+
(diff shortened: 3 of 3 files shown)


mishushakov (approved):

mishushakov:
we can merge once the CLA is signed

YizukiAme:
@mishushakov CLA is signed now. Ready to merge whenever you get a chance. Thanks!

PR #1263, opened 2026-04-12
fix(js-sdk): forward name param in createSnapshot and return names

### LINKED ISSUE #1249: JS SDK: createSnapshot does not forward the 'name' parameter to the REST API
## Problem

The REST API at `POST /sandboxes/{sandboxID}/snapshots` supports an optional `name` field in the request body, which allows creating named snapshots addressable via `Sandbox.create('team-slug/my-snapshot')`. The API reference documents this [here](https://e2b.dev/docs/api-reference/sandboxes/post-sandboxes-snapshots).

However, the JS SDK's `createSnapshot()` method:

1. Sends an empty `body: {}` — the `name` is never forwarded even if the caller passes it
2. Does not expose a `name` parameter in its options type (`SandboxApiOpts`)
3. Does not return the `names` array from the API response in `SnapshotInfo`

This means SDK users cannot create named snapshots and must use raw snapshot IDs.

## Expected behavior

```ts
const snapshot = await sandbox.createSnapshot({ name: 'my-snapshot' })
console.log(snapshot.snapshotId)  // 'team-slug/my-snapshot:default'
console.log(snapshot.names)       // ['team-slug/my-snapshot:default']

// Later, create a sandbox from the named snapshot
const sbx = await Sandbox.create('team-slug/my-snapshot')
```

## Actual behavior

```ts
// name is silently ignored — not in the type, not forwarded to the API
const snapshot = await sandbox.createSnapshot({ name: 'my-snapshot' })
console.log(snapshot.snapshotId)  // 'abc123xyz:default' (raw ID)
// snapshot.names is undefined
```

## Affected code

- `packages/js-sdk/src/sandbox/sandboxApi.ts` — `createSnapshot()` static method sends `body: {}`
- `packages/js-sdk/src/sandbox/sandboxApi.ts` — `SnapshotInfo` interface missing `names` field
- `packages/js-sdk/src/sandbox/index.ts` — instance `createSnapshot()` method
- The generated schema (`schema.gen.ts`) already includes both `name` in the request and `names` in the response

## Environment

- `e2b` JS SDK version: 2.14.1 through 2.18.0 (tested both)

Yanhu007:
## Summary

Fixes #1249

`createSnapshot()` sends `body: {}`, silently ignoring the `name` parameter. The `SnapshotInfo` interface also omits the `names` array from the API response. This means SDK users cannot create named snapshots.

### `sandboxApi.ts`
- **`CreateSnapshotOpts`**: new interface extending `SandboxApiOpts` with optional `name`
- **`SnapshotInfo`**: added `names: string[]` field
- **`createSnapshot()`**: forwards `name` to request body, returns `names` from response
- **`SnapshotPaginator.nextItems()`**: returns `names` from list response

### `index.ts` (sandbox)
- Updated instance `createSnapshot()` to accept `CreateSnapshotOpts`

### `index.ts` (package)
- Exported `CreateSnapshotOpts`

## Before / After

```ts
// Before: name silently ignored, names not returned
const snapshot = await sandbox.createSnapshot({ name: 'my-snapshot' })
// snapshot.snapshotId = 'abc123:default'
// snapshot.names = undefined

// After: name forwarded, names returned
const snapshot = await sandbox.createSnapshot({ name: 'my-snapshot' })
// snapshot.snapshotId = 'team/my-snapshot:default'
// snapshot.names = ['team/my-snapshot:default']
```

### DIFF
--- a/packages/js-sdk/src/index.ts
+++ b/packages/js-sdk/src/index.ts
@@ -61,6 +61,7 @@ export type {
   SandboxLifecycle,
   SandboxInfoLifecycle,
   SnapshotInfo,
+  CreateSnapshotOpts,
   SnapshotListOpts,
   SnapshotPaginator,
 } from './sandbox/sandboxApi'
--- a/packages/js-sdk/src/sandbox/index.ts
+++ b/packages/js-sdk/src/sandbox/index.ts
@@ -21,6 +21,7 @@ import {
   SandboxPaginator,
   SandboxBetaCreateOpts,
   SandboxApiOpts,
+  CreateSnapshotOpts,
   SnapshotListOpts,
   SnapshotInfo,
   SnapshotPaginator,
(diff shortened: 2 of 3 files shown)


mishushakov:
we're moving this to https://github.com/e2b-dev/E2B/pull/1277

PR #1267, opened 2026-04-13
feat: allow passing template as option in Sandbox.create()

mishushakov:
## Summary
- Adds `template` as an optional property on `SandboxOpts` in the JS SDK, enabling `Sandbox.create({ template: 'my-template' })` syntax
- Updates both `create` and `betaCreate` to check `opts.template` before falling back to the default template
- Python SDK already supports `Sandbox.create(template='template')` via named parameters, so no changes needed there

## Test plan
- [ ] Verify `Sandbox.create({ template: 'base' })` works
- [ ] Verify `Sandbox.create('base')` still works (backwards compatible)
- [ ] Verify `Sandbox.create()` still defaults to `'base'`
- [ ] Verify MCP template fallback still works when no template is specified

### DIFF
--- a/.changeset/sandbox-template-option.md
+++ b/.changeset/sandbox-template-option.md
@@ -1,0 +1,5 @@
+---
+'e2b': patch
+---
+
+allow passing template as an option in Sandbox.create()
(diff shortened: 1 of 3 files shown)


jakubno (approved):
maybe worth mentioning that if you specify mcp, it won't have any effect

PR #1013, opened 2025-11-06
Support overriding envd API URL

djeebus:
This requires [infra#1448](https://github.com/e2b-dev/infra/pull/1448) first. 

---

> [!NOTE]
> Adds support to override the sandbox API URL (E2B_SANDBOX_URL) across JS/Python SDKs, centralizes sandbox host/url logic with headers, and updates CI to build the SDK before CLI.
> 
> - **SDKs (JS & Python)**
>   - Add `sandboxUrl` support in `ConnectionConfig` (env var `E2B_SANDBOX_URL`), with new helpers `getSandboxUrl`/`getHost` and shared `envdPort`.
>   - Refactor sandbox initialization to use `ConnectionConfig.getSandboxUrl(...)` and `getHost(...)`.
>   - Always attach sandbox headers `E2b-Sandbox-Id` and `E2b-Sandbox-Port` to sandbox and connect requests.
>   - Python: thread `sandbox_url` through opts; update async/sync connect calls to pass headers; minor fix to default `headers=None` in `e2b_connect.client.Client` stream prep.
> - **CI**
>   - Build `packages/js-sdk` before `packages/cli`; set step `working-directory` for build/test.
> - **Dependencies**
>   - Point `e2b` dependency in lockfile to local `../js-sdk` link.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5dc58171af6c170f8640f49d736dbe9c571f2b21. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #1272, opened 2026-04-14
fix: move E2b-Sandbox-Id/Port headers from control plane connect to data plane ConnectionConfig

WoodenStone:
### Problem

The Python SDK's `Sandbox.connect()` currently injects `E2b-Sandbox-Id` and `E2b-Sandbox-Port` headers into the **control plane** request (`POST /sandboxes/{sandboxID}/connect`).

This causes issues for any proxy/gateway that uses these headers to distinguish between control plane and data plane traffic. Since these headers are the canonical signal for "this request targets a specific sandbox instance (data plane)", their presence on a control plane API call causes the proxy to misroute the request to the sandbox's envd endpoint instead of the API server — resulting in authentication failures (401).

The JS/TS SDK does **not** have this problem: `connectSandbox()` sends a clean control plane request without these headers, and only injects them on data plane requests via the sandbox instance's connection config.

### Root Cause

Introduced in [#1013](https://github.com/e2b-dev/E2B/pull/1013) (*Support overriding envd API URL*), the `_cls_connect` method was updated to pass `E2b-Sandbox-Id` / `E2b-Sandbox-Port` directly to `get_api_client()` as extra headers. This means they end up on the **control plane** HTTP client, polluting the `POST /sandboxes/{id}/connect` request with data plane routing signals.

### Fix

Move `E2b-Sandbox-Id` and `E2b-Sandbox-Port` from the control plane API client (`get_api_client()` in `sandbox_api.py`) to `extra_sandbox_headers` in the `ConnectionConfig` (set in `main.py`'s `_cls_connect_sandbox`).

This ensures:

- **Control plane requests** (via `config.headers` → `ApiClient`) do **not** carry these headers
- **Data plane requests** (via `config.sandbox_headers` → filesystem / commands / pty / health) **do** carry them, as intended

The fix applies to both sync and async SDK paths (4 files).

### Why this matters

These two headers serve as a **data plane routing signal** in the E2B protocol. Any transparent proxy, API gateway, or load balancer that multiplexes control plane and data plane traffic on a single endpoint will use their presence/absence to decide where to forward the request.

When a control plane API like `connect` carries these headers, it creates an ambiguous signal — the path says "control plane" but the headers say "data plane". This is not specific to any single proxy implementation; it affects any gateway that follows the protocol semantics.

### Verification

- Confirmed that after this fix, `connect` requests no longer carry `E2b-Sandbox-Id` / `E2b-Sandbox-Port`
- Confirmed that subsequent data plane requests (file ops, commands, health checks) still carry these headers via `sandbox_headers`
- Behavior now matches the JS/TS SDK

### DIFF
--- a/.changeset/fix-connect-unset-sanitization.md
+++ b/.changeset/fix-connect-unset-sanitization.md
@@ -1,0 +1,5 @@
+---
+'@e2b/python-sdk': patch
+---
+
+fix connect path: move sandbox headers to data plane and sanitize UNSET values
--- a/packages/python-sdk/e2b/sandbox_async/main.py
+++ b/packages/python-sdk/e2b/sandbox_async/main.py
@@ -861,8 +861,22 @@ class AsyncSandbox(SandboxApi):
         )
 
         sandbox_headers = {}
-        envd_access_token = sandbox.envd_access_token
-        if envd_access_token is not None and not isinstance(envd_access_token, Unset):
+        domain = (
+            sandbox.domain
+            if isinstance(sandbox.domain, str)
+            else None
+        )
+        envd_access_token = (
+            sandbox.envd_access_token
+            if isinstance(sandbox.envd_access_token, str)
+            else None
+        )
+        traffic_access_token = (
+            sandbox.traffic_access_token
+            if isinstance(sandbox.traffic_access_token, str)
+            else None
+        )
+        if envd_access_token is not None:
             sandbox_headers["X-Access-Token"] = envd_access_token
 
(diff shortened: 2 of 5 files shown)


mishushakov:
@claude review
@codex review
@cursor review

mishushakov:
@matthewlouisbrockman would like to ask you to check this one?

mishushakov:
Thanks for submitting the pull request - I think we can drop these altogether.

I am closing this PR and will proceed with removal here: https://github.com/e2b-dev/E2B/pull/1402

PR #1273, opened 2026-04-15
feat(js-sdk): add native wallet/payment support via MPP

decofe:
## Summary

Add native wallet support to the E2B JS SDK so AI agents can pay for sandbox usage with crypto instead of API keys.

## Motivation

With [MPP support in the E2B API](https://github.com/e2b-dev/infra/pull/2398), agents can now pay per request via the [Machine Payments Protocol](https://mpp.dev). This PR adds client-side support so the SDK handles the HTTP 402 flow automatically.

### `package.json`
- Add `mppx` as optional peer dependency

### `src/connectionConfig.ts`
- Add `payment` option to `ConnectionOpts` — accepts an mppx client instance
- Store payment config on `ConnectionConfig`

### `src/api/index.ts`
- Skip API key requirement when `payment` is configured
- Pass mppx `fetch` to `openapi-fetch` — handles 402 automatically
- Add 402 error handling in `handleApiError`

## Usage

```ts
import { Sandbox } from 'e2b'
import { Mppx, tempo } from 'mppx/client'
import { privateKeyToAccount } from 'viem/accounts'

// Create an mppx client with a Tempo wallet
const mppx = Mppx.create({
  methods: [tempo({ account: privateKeyToAccount('0x...') })],
})

// Create sandbox with crypto payment — no API key needed
  payment: { client: mppx },
})

await sandbox.commands.run('echo hello')
```

## Backward compatibility

Fully backward compatible. API key auth works exactly as before. The `payment` option is entirely additive and `mppx` is an optional peer dep.

Prompted by: georgios

### DIFF
--- a/packages/js-sdk/package.json
+++ b/packages/js-sdk/package.json
@@ -99,6 +99,14 @@
     "platform": "^1.3.6",
     "tar": "^7.5.11"
   },
+  "peerDependencies": {
+    "mppx": ">=0.0.1"
+  },
+  "peerDependenciesMeta": {
+    "mppx": {
+      "optional": true
+    }
+  },
   "engines": {
     "node": ">=20"
   },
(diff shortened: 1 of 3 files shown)


mishushakov:
We're not adding this

PR #1284, opened 2026-04-22
fix(js-sdk): forward name param and expose names in createSnapshot

### LINKED ISSUE #1249: JS SDK: createSnapshot does not forward the 'name' parameter to the REST API

FisherXZ:
## Summary

Closes #1249.

Named snapshots let users write \`Sandbox.create('team/my-snapshot')\` instead of tracking raw template IDs like \`abc123:default\`. But passing \`name\` to \`createSnapshot()\` / \`create_snapshot()\` silently had no effect in both SDKs — the generated REST clients were already correct, but the hand-written wrapper layer never forwarded the value or surfaced the response field.

Three layers had the same gap in both SDKs:

| Layer | Problem |
|---|---|
| Options type | No \`name\` field — callers had nowhere to pass it |
| Request body | \`body: {}\` / \`PostSandboxesSandboxIDSnapshotsBody()\` — \`name\` never included |
| \`SnapshotInfo\` return type | Missing \`names\` field — API response aliases silently dropped |
| Paginators | \`listSnapshots\` / \`list_snapshots\` had the same \`names\` omission |

**JS SDK changes** (\`packages/js-sdk/\`):
- Added \`SnapshotCreateOpts\` interface with \`name?: string\`
- Added \`names: string[]\` to \`SnapshotInfo\`
- Forwarded \`name\` in \`SandboxApi.createSnapshot()\` using \`!== undefined\` guard (preserves \`name: ''\`)
- Fixed \`SnapshotPaginator\` to map \`names\`
- Exported \`SnapshotCreateOpts\` from the package index

**Python SDK changes** (\`packages/python-sdk/\`):
- Added \`names: List[str]\` to \`SnapshotInfo\` dataclass (\`default_factory=list\`)
- Added \`name: Optional[str] = None\` to \`_cls_create_snapshot()\` in both sync and async, forwarded as \`name if name is not None else UNSET\`
- Updated \`create_snapshot()\` overloads and implementations (sync + async)
- Fixed both \`SnapshotPaginator.next_items()\` to map \`names\`

Backwards compatible: \`name\` is optional everywhere, \`names\` defaults to \`[]\`.

## Usage

**JS:**
\`\`\`ts
const snap = await sandbox.createSnapshot({ name: 'my-snapshot' })
console.log(snap.snapshotId)  // 'team-slug/my-snapshot:default'
console.log(snap.names)       // ['team-slug/my-snapshot:default']

// Now usable by name instead of raw ID
const sbx = await Sandbox.create('team-slug/my-snapshot')
\`\`\`

**Python:**
\`\`\`python
snap = sandbox.create_snapshot(name='my-snapshot')
print(snap.snapshot_id)  # 'team-slug/my-snapshot:default'
print(snap.names)        # ['team-slug/my-snapshot:default']

sbx = Sandbox.create('team-slug/my-snapshot')
\`\`\`

## Test plan

- [x] \`pnpm run typecheck\` — zero errors
- [x] \`pnpm run lint\` — zero errors
- [x] Python \`make lint\` (ruff check + ruff format) — all pass
- [x] Changeset added: \`patch\` bump for \`e2b\` (JS) and \`@e2b/python-sdk\`
- [x] Backwards compatible: \`name\` is optional, \`names\` defaults to \`[]\`
- [ ] Live integration: \`createSnapshot({ name: ... })\` returns alias in \`names\` — pending backend round-trip

### DIFF
--- a/.changeset/fix-snapshot-name.md
+++ b/.changeset/fix-snapshot-name.md
@@ -1,0 +1,6 @@
+---
+"e2b": patch
+"@e2b/python-sdk": patch
+---
+
+Forward optional `name` parameter in `createSnapshot` / `create_snapshot` and return `names` array in `SnapshotInfo` for both JS and Python SDKs.
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ vale-styles
 .vercel
 .vscode
 .DS_Store
+.worktrees
 
 # Logs
 logs
--- a/packages/js-sdk/src/index.ts
+++ b/packages/js-sdk/src/index.ts
@@ -62,6 +62,7 @@ export type {
   SandboxInfoLifecycle,
   SnapshotInfo,
   SnapshotListOpts,
+  SnapshotCreateOpts,
   SnapshotPaginator,
 } from './sandbox/sandboxApi'
 
--- a/packages/js-sdk/src/sandbox/index.ts
+++ b/packages/js-sdk/src/sandbox/index.ts
@@ -20,8 +20,8 @@ import {
   SandboxListOpts,
   SandboxPaginator,
   SandboxBetaCreateOpts,
-  SandboxApiOpts,
   SnapshotListOpts,
+  SnapshotCreateOpts,
   SnapshotInfo,
   SnapshotPaginator,
 } from './sandboxApi'
@@ -621,7 +621,7 @@ export class Sandbox extends SandboxApi {
    * const newSandbox = await Sandbox.create(snapshot.snapshotId)
    * ```
    */
-  async createSnapshot(opts?: SandboxApiOpts): Promise<SnapshotInfo> {
+  async createSnapshot(opts?: SnapshotCreateOpts): Promise<SnapshotInfo> {
     return await SandboxApi.createSnapshot(
       this.sandboxId,
       this.resolveApiOpts(opts)
--- a/packages/js-sdk/src/sandbox/sandboxApi.ts
+++ b/packages/js-sdk/src/sandbox/sandboxApi.ts
@@ -722,7 +722,7 @@ export class SandboxApi {
         },
       },
       body: {
-        ...(opts?.name && { name: opts.name }),
+        ...(opts?.name !== undefined && { name: opts.name }),
       },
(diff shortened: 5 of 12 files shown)


mishushakov:
we're working on it here
https://github.com/e2b-dev/E2B/pull/1277

PR #1285, opened 2026-04-22
fix(js-sdk): buffer template upload to avoid S3 chunked PUT 501

### LINKED ISSUE #1243: [JS SDK] uploadFile uses chunked transfer encoding, causing 501 NotImplemented on S3 presigned PUT URLs
## Bug Report

  ### Summary

  When building a template using the JS SDK in a self-hosted deployment, uploading files via `.copy()` fails with `501 NotImplemented` from S3. The root
  cause is that `uploadFile` sends the tar stream with `Transfer-Encoding: chunked`, which S3 presigned PUT URLs do not support.

  ### Environment

  - e2b JS SDK: v2.18.0 (latest)
  - Node.js: v22
  - Storage: Self-hosted with S3-compatible storage (presigned PUT URL returned directly to SDK)

  ### Error

  501 NotImplemented
  A header you provided implies functionality that is not implemented

  ### Root Cause

  In `dist/index.js`, `uploadFile` passes a Node.js `Readable` stream directly as the `fetch` body:

  ```js
  const res = await fetch(url, {
    method: "PUT",
    body: uploadStream,  // Node.js Readable, no Content-Length
    duplex: "half"
  });

  Since Content-Length is unknown, Node.js undici falls back to Transfer-Encoding: chunked. S3 presigned PUT URLs reject chunked requests with 501
  NotImplemented — this is an S3 protocol-level restriction that cannot be resolved via IAM or bucket policy.

  Comparison with Python SDK

  The Python SDK (v2.19.0) already handles this correctly — it buffers the tar archive into io.BytesIO first, then uploads bytes via httpx, which sets
  Content-Length automatically:

  tar_buffer = tar_file_stream(...)
  response = client.put(url, content=tar_buffer.getvalue())  # bytes, not stream

  ┌───────────────────┬────────────────────┬───────────────────┐
  │                   │       JS SDK       │    Python SDK     │
  ├───────────────────┼────────────────────┼───────────────────┤
  │ Body type         │ ReadableStream     │ bytes             │
  ├───────────────────┼────────────────────┼───────────────────┤
  │ Content-Length    │ Not set            │ Set automatically │
  ├───────────────────┼────────────────────┼───────────────────┤
  │ Transfer-Encoding │ chunked            │ fixed-length      │
  ├───────────────────┼────────────────────┼───────────────────┤
  │ S3 presigned PUT  │ 501 NotImplemented │ Works             │
  └───────────────────┴────────────────────┴───────────────────┘

  Suggested Fix

  Buffer the stream into memory before uploading, consistent with how the Python SDK works:

  // Before
  const res = await fetch(url, {
    method: "PUT",
    body: uploadStream,
    duplex: "half"
  });

  // After
  const _buf = await new Promise((resolve, reject) => {
    const chunks = [];
    uploadStream.on("data", (chunk) => chunks.push(Buffer.from(chunk)));
    uploadStream.on("end", () => resolve(Buffer.concat(chunks)));
    uploadStream.on("error", reject);
  });
  const res = await fetch(url, {
    method: "PUT",
    body: _buf,
    headers: { "Content-Length": String(_buf.length) }
  });

  This is a 5-line change and aligns the JS SDK behavior with the Python SDK. Happy to submit a PR if helpful.

  ---

truffle-dev:
Fixes #1243.

`uploadFile` passed the gzipped tar `Readable` directly to `fetch`. With no known `Content-Length`, undici falls back to `Transfer-Encoding: chunked`, which S3 presigned PUT URLs reject with `501 NotImplemented`. Template uploads against S3-compatible storage in self-hosted deployments were broken.

Buffer the archive with `stream/consumers.buffer()` before PUT so `fetch` sets `Content-Length` automatically. This aligns the JS SDK with the Python SDK, which already materializes the tar into `io.BytesIO` and uploads bytes via httpx (`packages/python-sdk/e2b/template_sync/build_api.py:upload_file`).

## Verification

Added `tests/template/uploadFile.test.ts`: spins up a local HTTP server, runs `uploadFile`, asserts `Content-Length` is set and matches the received body length, and asserts no chunked transfer encoding.

Stash-bisected against the fix:

```
# pre-fix (body = Readable)
× expected undefined to be defined  (Content-Length was never sent)

# post-fix (body = Buffer)
✓ sets Content-Length and does not use chunked transfer encoding  (68ms)
```

All 51 JS SDK unit tests (`tests/template/utils` + new file) pass. The 22 failures in the full `tests/template` run are integration tests that require `E2B_API_KEY`; they are unrelated.

## Trade-off note

Like the Python SDK, this materializes the archive in memory before PUT. Template build contexts are bounded (the SDK targets Docker-build-sized payloads) and aligning with the Python SDK keeps behavior consistent across clients. If memory pressure becomes an issue for very large uploads, a streaming alternative would need coordinated changes on both SDKs and the server to compute Content-Length up front.

Issue: https://github.com/e2b-dev/E2B/issues/1243

### DIFF
--- a/.changeset/fix-js-sdk-upload-content-length.md
+++ b/.changeset/fix-js-sdk-upload-content-length.md
@@ -1,0 +1,5 @@
+---
+"e2b": patch
+---
+
+fix(js-sdk): buffer template tar archive before upload so `fetch` sets `Content-Length` instead of falling back to `Transfer-Encoding: chunked`. S3 presigned PUT URLs reject chunked requests with `501 NotImplemented`, breaking template uploads in self-hosted deployments backed by S3-compatible storage. Aligns the JS SDK with the Python SDK, which already buffers via `io.BytesIO`.
--- a/packages/js-sdk/src/template/buildApi.ts
+++ b/packages/js-sdk/src/template/buildApi.ts
@@ -1,3 +1,4 @@
+import { buffer } from 'node:stream/consumers'
 import { ApiClient, handleApiError, paths, components } from '../api'
 import { stripAnsi } from '../utils'
 import { BuildError, FileUploadError, TemplateError } from '../errors'
@@ -119,12 +120,16 @@ export async function uploadFile(
       resolveSymlinks
     )
 
-    // The compiler assumes this is Web fetch API, but it's actually Node.js fetch API
+    // Buffer the archive before uploading so fetch sets Content-Length.
+    // S3 presigned PUT URLs reject Transfer-Encoding: chunked with 501
+    // NotImplemented, which is what Node's fetch falls back to when the
+    // body is a Readable without a known length. See e2b-dev/e2b#1243.
+    // The Python SDK takes the same approach (build_api.py:upload_file).
+    const uploadBody = await buffer(uploadStream)
+
     const res = await fetch(url, {
       method: 'PUT',
-      // @ts-expect-error
-      body: uploadStream,
-      duplex: 'half',
+      body: uploadBody,
     })
 
     if (!res.ok) {
--- a/packages/js-sdk/src/template/buildApi.ts
+++ b/packages/js-sdk/src/template/buildApi.ts
@@ -1,6 +1,5 @@
-import { buffer } from 'node:stream/consumers'
 import { ApiClient, handleApiError, paths, components } from '../api'
-import { stripAnsi } from '../utils'
+import { dynamicImport, stripAnsi } from '../utils'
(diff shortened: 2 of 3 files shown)


mishushakov (approved):

mishushakov:
@cursor review
@codex review
@claude review

mishushakov (approved):

mishushakov:
moved to https://github.com/e2b-dev/E2B/pull/1285

PR #1286, opened 2026-04-22
feat(sdk): support structured network rules with per-host transforms

mishushakov:
## Summary
- Adds a `network.rules` map keyed by host (or CIDR / IP) for per-host outbound request transforms (e.g. header injection). Registering a host in `rules` does **not** grant egress on its own — it must still be referenced via `allowOut`. JS accepts either a plain object or a `Map`.
- Extends `network.allowOut` / `network.denyOut` to accept either a static list or a **selector callback** receiving `{ allTraffic, rules }` (Python: `ctx.all_traffic`, `ctx.rules`). `allTraffic` is `'0.0.0.0/0'`; `rules` is a `Map` (Python `Mapping`) view of `network.rules`, so policies can be composed against the registered rule hosts without duplicating them.
- `transform` on a rule accepts a static `{ headers: ... }` object. The header values are sent on the wire as-is and injected by the egress proxy on matching requests.
- Updates the OpenAPI spec, regenerates JS and Python API clients, and surfaces new types: `SandboxNetworkRule`, `SandboxNetworkRules`, `SandboxNetworkTransform`, `SandboxNetworkSelector`, `SandboxNetworkSelectorContext`, `SandboxNetworkInfo`.
- Adds contract tests (JS + async/sync Python) that create a sandbox with a `transform.headers` rule for `httpbin.e2b.team`, run `curl https://httpbin.e2b.team/headers`, and assert the injected header is reflected back.

## Follow-up
The callback variant — `transform: ({ identity, sandboxId }) => (...)` — and its placeholder context (`${e2b.sandboxId}`, `${e2b.teamId}`, `${e2b.executionId}`, `${e2b.identity.jwt}`) are split out into #1335, stacked on top of this PR. That lets us land the static shape first and wire up the proxy-side placeholder resolution separately.

### Inject a static header on requests to a specific host (TypeScript)

```ts
import { Sandbox } from 'e2b'

await Sandbox.create({
  network: {
    // Only allow egress to hosts that have rules registered.
    allowOut: ({ rules }) => [...rules.keys()],
    rules: {
      'api.openai.com': [
        {
          transform: {
            headers: { 'X-Header': 'Content' },
          },
        },
      ],
    },
  },
})
```

### Same thing in Python

```python
from e2b import Sandbox

sandbox = Sandbox(
    network={
        "allow_out": lambda ctx: list(ctx.rules.keys()),
        "rules": {
            "api.openai.com": [
                {
                    "transform": {
                        "headers": {"X-Header": "Content"},
                    },
                },
            ],
        },
    },
)
```

### `network.rules` as a `Map` (TypeScript)

```ts
const rules = new Map([
  ['api.openai.com', [{ transform: { headers: { 'X-Trace': 'on' } } }]],
])

await Sandbox.create({
  network: { allowOut: ({ rules }) => [...rules.keys()], rules },
})
```

### Block all egress except an explicit allowlist

```ts
await Sandbox.create({
  network: {
    denyOut: ({ allTraffic }) => [allTraffic],   // allTraffic === '0.0.0.0/0'
    allowOut: ['1.1.1.1', '8.8.8.0/24'],
  },
})
```

```python
Sandbox(
    network={
        "deny_out": lambda ctx: [ctx.all_traffic],
        "allow_out": ["1.1.1.1", "8.8.8.0/24"],
    },
)
```

### Static list (unchanged, still supported)

```ts
await Sandbox.create({
  network: { allowOut: ['1.1.1.1', '8.8.8.0/24'] },
})
```

```python
Sandbox(network={"allow_out": ["1.1.1.1", "8.8.8.0/24"]})
```

## Notes
- SDK/spec-only; the sandbox egress proxy still needs to honor `transform.headers` for the httpbin assertion to pass end-to-end — the new test locks the wire shape as a contract.
- `ALL_TRAFFIC` is still exported but the selector form (`({ allTraffic }) => [allTraffic]` / `lambda ctx: [ctx.all_traffic]`) is the recommended way to express "everything".

## Test plan
- [ ] Spec + generated clients reviewed for shape correctness
- [ ] Existing `network.test.ts` / `test_network.py` cases still pass once backend auth is available
- [ ] New `firewall transform injects headers` cases pass once egress proxy injects headers


nauxliu:
Thanks for working on this. I’m trying to understand the intended pattern for credential brokering and request logging.

For our use case, we want connectors to integrate with third-party platforms like CRMs, Jira, GitHub, etc. Sandboxes need to make outbound API calls to these services, while the actual credentials should stay outside the sandbox. The credential to inject may depend on the actual sandbox, tenant, and user.

Ideally, this would be an egress-time callback where we can access the original outbound request, inspect the URL / method / headers, inject the right token, and log or audit the request.

From my reading, `transform` is evaluated at sandbox creation time and produces a static config. It does not run at egress time, cannot do async lookup, cannot access the original request, and cannot use the real runtime sandbox context. That makes dynamic credential injection and request logging difficult.

If the design is intentionally static, could E2B at least support updating the network config after sandbox creation, similar to Vercel’s `updateNetworkPolicy()`? That would allow us to create the sandbox, resolve credentials based on the real sandbox ID, and then patch the transform.

Is that the expected workaround, or is there a better recommended pattern?

mishushakov:
@nauxliu thanks for the feedback!

We will consider making it possible to update the networking configuration after the Sandbox creation - I will check in with the team!

packages/js-sdk/src/sandbox/sandboxApi.ts
```diff
@@ -402,6 +500,50 @@ export interface SandboxMetrics {
   diskTotal: number
 }
 
+function resolveNetworkSelector(
```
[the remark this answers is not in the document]
mishushakov: will fix in a follow-up PR

spec/openapi.yml
```diff
@@ -361,12 +391,12 @@ components:
           description: Logs of the sandbox
           type: array
           items:
-            $ref: '#/components/schemas/SandboxLog'
+            $ref: "#/components/schemas/SandboxLog"
```
djeebus: Are these changes required in this PR? Can we ensure that this *won't* happen again (because they're required by linting rules), or revert these changes?

mishushakov: I have same questions for the infra repo - can we have same rules?

matthewlouisbrockman: we don't have linters on any of the yamls it looks like and yeah, it does look like infra's double quote.

let's just split it then because this is combining a couple different things; can we get 1 PR to do the openapi change (adds the models, reformat the quotes) and one for this stuff?

matthewlouisbrockman: then we can do a 3rd+4th pr to see if we can get yaml linters on both repos

matthewlouisbrockman (approved):

Issue #888, opened 2025-09-02
Add `py.typed` file for type checking


draftcode:
**Is your feature request related to a problem? Please describe.**

Python type checkers expect that type-annotated packages to have `py.typed` file. Without this, type checkers do not do type checking for that package. E2B Python SDK doesn't have this file, and because of this, type checkers do not type-check code around e2b related APIs. Developers using the SDK won't receive warnings about type mismatches, which can lead to runtime errors.

See https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/ or https://typing.python.org/en/latest/spec/distributing.html#packaging-type-information for the official instruction on distributing type-annotated libraries.

**Describe the solution you'd like**

Add an empty file at following path:

* `packages/python-sdk/e2b/py.typed`
* `packages/python-sdk/e2b_connect/py.typed`

Considering that the Python SDK is using Poetry, adding these two files to the repository would be enough to enable type-checkers check user code against E2B SDK.

**Describe alternatives you've considered**

This is more about E2B Python SDK not conforming to the Python library distribution standard. Need to be fixed on the distribution side.

**Additional context**

PR #1298, opened 2026-04-28
feat(python-sdk): add py.typed marker for PEP 561 type checking support

FisherXZ:
## Context

Python packages can opt into PEP 561 by shipping a `py.typed` marker file alongside their source. Without it, type checkers like `mypy --strict` and `pyright` treat the package as untyped and suppress all type errors at the import boundary — meaning users get no type safety from `e2b` even though the SDK ships full annotations.

Issue: #888

## Investigation

The `e2b` Python SDK already ships complete type annotations across all public APIs (`sandbox/`, `sandbox_async/`, `sandbox_sync/`, etc.). The only missing piece was the `py.typed` marker file that signals to type checkers that the package is typed.

Checked `pyproject.toml` — no additional configuration needed. The existing:

```toml
packages = [{ include = "e2b" }, { include = "e2b_connect" }]
```

tells Poetry to include the entire `e2b/` directory recursively in the built wheel, which picks up `py.typed` automatically.

## Fix

Added an empty `py.typed` marker file at `packages/python-sdk/e2b/py.typed` per the PEP 561 spec. No `pyproject.toml` or `__init__.py` changes required.

Also includes a pre-existing `ruff` formatting fix in `test_stacktrace.py` surfaced by `make format`.

## Test Plan

- [x] `make lint` (`ruff check` + `ruff format --check`) — passes
- [ ] `make typecheck` — local Poetry venv has a broken Python 3.13 dylib (pre-existing env issue unrelated to this change); CI should confirm
- [ ] Downstream verification: add `e2b` to a `mypy --strict` or `pyright` project and confirm no `Cannot find implementation or library stub` errors

### DIFF
--- a/.changeset/python-sdk-py-typed.md
+++ b/.changeset/python-sdk-py-typed.md
@@ -1,0 +1,5 @@
+---
+"@e2b/python-sdk": patch
+---
+
+add py.typed marker for PEP 561 type checking support
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ vale-styles
 .vercel
 .vscode
 .DS_Store
+.worktrees
 
 # Logs
 logs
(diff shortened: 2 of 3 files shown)


mishushakov:
closing, unnecessary change

PR #1300, opened 2026-04-28
fix(cli): prevent shell command injection in template build commands

johnpippett:
## Summary

Replaces `child_process.execSync` with string interpolation for `docker login`, `docker build`, and `docker push` commands with `child_process.spawnSync` using explicit array arguments. This eliminates shell metacharacter injection vectors in the CLI template build flow.

## Security Impact

- **Critical finding:** Shell command injection in `packages/cli/src/commands/template/build.ts:397,380` (identified in security audit).
- **Attack vectors closed:**
  - `--build-arg` CLI options (user-controlled key/value pairs) could inject shell metacharacters.
  - `accessToken` passed via `echo \"\" | docker login ...` could break out of quotes.
  - `dockerfileRelativePath` and `imageUrl` interpolated into shell strings.

## Changes

- `docker login`: uses `spawnSync` with `input` piped to stdin instead of `echo |` shell pipeline.
- `docker build`: builds argument array directly; `--build-arg` values are passed as discrete array elements.
- `docker push`: uses `spawnSync` with array arguments.
- Non-zero exit statuses are manually checked and thrown to preserve existing error-handling behavior.

## Verification

- `tsc --noEmit --skipLibCheck` passes.
- All existing CLI tests pass (73 passed, 7 skipped).

### DIFF
--- a/packages/cli/src/commands/template/build.ts
+++ b/packages/cli/src/commands/template/build.ts
@@ -376,13 +376,20 @@ Migration guide: ${asPrimary('https://e2b.dev/docs/template/migration-v2')}`
 
         if (imageUriMask == undefined) {
           try {
-            child_process.execSync(
-              `echo "${accessToken}" | docker login docker.${connectionConfig.domain} -u _e2b_access_token --password-stdin`,
+            const loginResult = child_process.spawnSync(
+              'docker',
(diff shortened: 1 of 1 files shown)


mishushakov:
Irrelevant fix - we have switched to new build system, this is only for legacy support.

PR #1309, opened 2026-04-30
test(python-sdk): parallelize async sandbox teardown

mishushakov:
## Summary
- Replace the sequential `for sandbox in sandboxes: await sandbox.kill()` teardown in `async_sandbox_factory` with `asyncio.gather(..., return_exceptions=True)` so a failure on one kill no longer blocks cleanup of the rest, and teardown runs concurrently.
- Hoist the `_test_failed` check above the loop so the failure log is printed once per fixture rather than per sandbox.

## Test plan
- [ ] `pnpm run format`, `pnpm run lint`, `pnpm run typecheck` (python-sdk) all pass
- [ ] `pnpm run test` against affected python async sandbox tests

### DIFF
--- a/packages/python-sdk/tests/conftest.py
+++ b/packages/python-sdk/tests/conftest.py
@@ -83,9 +83,12 @@ async def async_sandbox_factory(request, template, sandbox_test_id):
         for sandbox in sandboxes:
             print(f"\n[TEST FAILED] Sandbox ID: {sandbox.sandbox_id}")
 
-    await asyncio.gather(
+    results = await asyncio.gather(
         *(sandbox.kill() for sandbox in sandboxes), return_exceptions=True
     )
+    for sandbox, result in zip(sandboxes, results):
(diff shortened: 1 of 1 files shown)


packages/python-sdk/tests/conftest.py
```diff
@@ -79,13 +79,13 @@ async def factory(*, template_name: str = template, **kwargs):
...
-        if getattr(request.node, "_test_failed", False):
+    if getattr(request.node, "_test_failed", False):
+        for sandbox in sandboxes:
             print(f"\n[TEST FAILED] Sandbox ID: {sandbox.sandbox_id}")
-        try:
-            await sandbox.kill()
-        except Exception:
-            pass
+
+    await asyncio.gather(
```
jakubno: do we want to raise the error here? It changes the behavior a little

jakubno (approved):

PR #1315, opened 2026-05-07
fix: Fix incorrect gating of ACCESS_TOKEN for template commands

wj-e2b:


packages/cli/src/commands/template/build.ts
```diff
@@ -377,7 +378,7 @@ Migration guide: ${asPrimary('https://e2b.dev/docs/template/migration-v2')}`
         if (imageUriMask == undefined) {
           try {
             child_process.execSync(
-              `echo "${accessToken}" | docker login docker.${connectionConfig.domain} -u _e2b_access_token --password-stdin`,
+              `echo "${dockerAuthToken}" | docker login docker.${connectionConfig.domain} -u _e2b_access_token --password-stdin`,
```
jakubno, line 381: I don't think this is supported by docker reverse proxy

jakubno (changes requested):

mishushakov:
I think we should not touch the legacy build system code here

mishushakov:
closing the PR, as the changes are now in https://github.com/e2b-dev/E2B/pull/1361

PR #1320, opened 2026-05-10
fix(cli): restrict ~/.e2b/config.json permissions to owner-only

andesyteoss:
## Summary

The CLI stores credentials (E2B access token and team API key) in plaintext at `~/.e2b/config.json`. Today the file is created with the process default umask, which on most Linux distributions and macOS results in mode `0644` — readable by every other local user and by any process running as a different UID on the same machine.

This PR routes all three write sites through a single `writeUserConfig()` helper that creates `~/.e2b` as `0700` and `config.json` as `0600`, matching the convention used by the AWS CLI (`~/.aws/credentials`), `kubectl` (`~/.kube/config`), and `gh` (`~/.config/gh/hosts.yml`).

- **CWE:** CWE-312 (Cleartext Storage of Sensitive Information) — partial mitigation. The file remains plaintext on disk (the existing `// TODO` in `user.ts` already acknowledges that keychain storage is the proper long-term fix); this change reduces exposure to other local users / less-privileged processes, which is the standard industry mitigation while plaintext storage remains.
- **Affected file:** `packages/cli/src/user.ts` and the three writers in `packages/cli/src/commands/`.
- **Severity:** Moderate on shared / multi-user machines (CI runners, dev VMs, jump boxes); low on single-user workstations.

## What's in `~/.e2b/config.json`

```ts
{
  email, accessToken,           // user access token
  teamName, teamId, teamApiKey  // team API key
}
```

`accessToken` authenticates the user against the E2B control plane; `teamApiKey` authorizes sandbox creation against the team. Either is sufficient to impersonate the user / spend on the team's account.

## Fix

A new helper in `packages/cli/src/user.ts`:

```ts
export function writeUserConfig(configPath: string, config: UserConfig): void {
  const dir = path.dirname(configPath)
  fs.mkdirSync(dir, { recursive: true, mode: 0o700 })
  fs.chmodSync(dir, 0o700)
  fs.writeFileSync(configPath, JSON.stringify(config, null, 2), { mode: 0o600 })
  fs.chmodSync(configPath, 0o600)
}
```

The explicit `chmodSync` calls are intentional: `mkdirSync({ mode })` and `writeFileSync({ mode })` only set permissions when the path is created. If the directory or file already exists with looser permissions (the common case for users upgrading), `chmodSync` corrects them on the next write.

Call sites updated:
- `packages/cli/src/commands/auth/login.ts`
- `packages/cli/src/commands/auth/configure.ts`
- `packages/cli/src/commands/template/buildWithProxy.ts`

`logout` uses `unlinkSync` and is unaffected. I grep'd the package for any other writers to `USER_CONFIG_PATH` — these three are the complete set.

Behavior on Windows: `chmodSync` only manipulates the read-only bit on Windows, which is consistent with how the AWS/kubectl/gh CLIs behave. ACL hardening on Windows is out of scope for this change.

## Tests

Added `packages/cli/tests/user_config_permissions.test.ts`, which writes a config to a temporary path and asserts the resulting directory is `0700` and file is `0600`, plus that the JSON round-trips correctly.

Manually verified before/after on Linux:

```
# before this PR
$ ls -l ~/.e2b/config.json
-rw-r--r-- 1 user user 234 ... config.json
# after
$ ls -l ~/.e2b/config.json
-rw------- 1 user user 234 ... config.json
```

## Why this is worth fixing

The exploitable scenario is a multi-tenant or shared-account host: another local user (or a process running as `nobody`, a CI worker UID, a sandboxed app, etc.) can `cat ~/<victim>/.e2b/config.json` and lift live credentials. No privilege escalation, no race, no special tooling — the file is simply world-readable today.

Before submitting, I tried to disprove the finding: I checked whether E2B sets a restrictive umask anywhere in the CLI bootstrap (it doesn't), whether the tokens are short-lived enough to make disclosure low-impact (the access token isn't visibly rotated and the team API key is long-lived), and whether the directory itself was being created restrictively elsewhere (it wasn't — `mkdirSync` was called with default mode). None of those mitigations are in place, so the permission tightening is doing real work.

This doesn't close out CWE-312 — that requires moving the secrets out of plaintext entirely, which the existing TODO acknowledges. It does close the "any local user can read it" gap, which is the cheap, high-value half of the mitigation.


andesyteoss:
Thanks for the review — you're right, the POSIX bit assertions would fail on `windows-latest` since Node reports Windows-derived broad modes regardless of `chmod`.

Fixed in 43e169c: the `0o700`/`0o600` mask checks are now gated on `process.platform !== 'win32'`, while the JSON content assertion still runs on every platform. The underlying `writeUserConfig` chmod calls are unchanged (they're effectively no-ops on Windows, which is the existing platform behavior).

mishushakov:
@sebastiondev can you sign the CLA please? would like to proceed with the merge

mishushakov (approved):

andesyteoss:
Thanks @mishushakov — I've signed the CLA at https://e2b.dev/docs/cla.

@cla-bot check

mishushakov:
@cla-bot check

mishushakov:
@sebastiondev please sign the CLA, otherwise we cannot proceed with the PR

mishushakov:
@sebastiondev  reminder to sign CLA!

andesyteoss:
Hi @mishushakov — apologies for the delay on this. I've re-signed the CLA at https://e2b.dev/docs/cla just now using the same email associated with my GitHub account.

@cla-bot check

If the bot still doesn't pick it up, there may be a mismatch between the email I signed with and the one on my commits. Happy to coordinate on the correct email if needed — just let me know what the CLA system expects.

andesyteoss:
Thanks for the ping — will get the CLA signed today.

Signed-
sebastiondev

mishushakov:
@cla-bot check

PR #1338, opened 2026-05-20
PoC: Support profiles

sitole:
Support for having multiple configuration for E2B CLI. This helps with multi-environemtn and mult-team setups.

- Support `~/.e2b/config` TOML file with different profiles
- Every command accepts `--profile` flag for the profile that should be loaded
- Deprecates config.json and replaces it with a TOML file (easier for humans to edit) - this is an option and we can go back


ValentaTomas:

PR #1339, opened 2026-05-21
feat(js-sdk): support AbortSignal for Template operations

mishushakov:
## Summary

- Extends AbortSignal support (introduced for Sandbox in #1328) to `Template.build`, `buildInBackground`, `getBuildStatus`, `exists`, `assignTags`, `removeTags`, and `getTags`. Aborting the signal cancels in-flight requests and, for `Template.build`, the status-polling loop.
- Refactors signal+timeout plumbing: `ConnectionConfig` now stores `signal`, and `ApiClient` / `VolumeApiClient` auto-apply it (plus `requestTimeoutMs`) to every request via a custom fetch wrapper. The 22 explicit \`signal: config.getSignal(...)\` lines across sandboxApi.ts and volume/index.ts are dropped.
- Volume's \`FILE_TIMEOUT_MS\` overrides move one level up into the \`VolumeConnectionConfig\` constructor. Dead \`VolumeConnectionConfig.getSignal()\` removed.

## Test plan

- [x] \`pnpm run typecheck\` / \`lint\` / \`format\`
- [x] New \`tests/template/abortSignal.test.ts\` covering all 8 Template entry points (MSW-based)
- [x] Existing \`tests/sandbox/abortSignal.test.ts\`, \`tests/template/uploadFile.test.ts\`, \`tests/connectionConfig.test.ts\` still pass


mishushakov:
@claude can we add FILETIMEOUT_MS for uploadFile as well

packages/js-sdk/src/connectionConfig.ts
```diff
@@ -191,6 +221,7 @@ export class ConnectionConfig {
     this.logger = opts?.logger
     this.headers = opts?.headers || {}
     this.headers['User-Agent'] = `e2b-js-sdk/${version}`
+    this.signal = opts?.signal
```
matthewlouisbrockman: not sure keeping this signal in the config is safe; this config is also used by the sandbox which is persistent and if it gets aborted it'll abort all later operations from that client to that sandbox. might make sense to keep it just to the volumes/templates?

matthewlouisbrockman (changes requested):
i think abort signal should be per operation; if this is on the sandbox, if you do sandbox.create(), abort after getting it back, then later try anything everything's aborted.

packages/js-sdk/src/connectionConfig.ts
matthewlouisbrockman: looks resolved with latest commit, thanks.

matthewlouisbrockman (approved):

PR #1345, opened 2026-05-22
fix(cli): handle missing xdg-open on headless machines during login

mishushakov:
## Summary
- `e2b auth login` previously crashed on headless machines (no `xdg-open`) with an unhandled `error` event from the spawned browser process.
- Attach an `error` listener (and `.catch`) to the `open` call; on failure, print the login URL so the user can open it manually.
- Added a changeset for `@e2b/cli` (patch).

## Test plan
- [ ] On a headless Linux box without `xdg-open`, run `e2b auth login` and confirm the CLI prints the manual URL instead of crashing.
- [ ] On macOS/Linux with a desktop, confirm the browser still opens automatically and login completes.

### DIFF
--- a/packages/cli/src/utils/openBrowser.ts
+++ b/packages/cli/src/utils/openBrowser.ts
@@ -1,0 +1,29 @@
+import { spawn } from 'child_process'
+
+// Spawn the platform's URL opener ourselves so the 'error' listener is attached
+// synchronously. The `open` package (v9.x) only attaches its listener after a
+// microtask, by which point a `spawn` ENOENT (e.g. missing `xdg-open` on
+// headless Linux) has already been emitted and crashes the process — see
+// sindresorhus/open#144.
+export function openUrlInBrowser(url: string, onError: () => void): void {
+  let command: string
+  let args: string[]
+  if (process.platform === 'darwin') {
+    command = 'open'
+    args = [url]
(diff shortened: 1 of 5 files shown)


packages/cli/src/utils/openBrowser.ts
```diff
@@ -0,0 +1,29 @@
...
+    command = 'cmd'
+    args = ['/c', 'start', '""', url.replace(/&/g, '^&')]
+  } else {
+    command = 'xdg-open'
+    args = [url]
+  }
+
+  try {
+    const child = spawn(command, args, { stdio: 'ignore', detached: true })
+    child.once('error', onError)
```
matthewlouisbrockman, line 24: approving, but edge case for later; if you have xdg but no browser, it'll stall/timeout.

e.g. DEBIAN_FRONTEND=noninteractive apt-get install -y xdg-utils and then run, it just hangs at `attempting to login`

matthewlouisbrockman (approved):

packages/cli/src/utils/openBrowser.ts
mishushakov: https://linear.app/e2b/issue/ENG-4124/cli-auth-should-handle-browser-being-unavailable-when-xgd-is-installed

PR #1347, opened 2026-05-22
feat(python-sdk): allow opting out of HTTP/2 in `get_transport`

jakubno:


ValentaTomas (approved):

mishushakov:
pls on JS too

jakubno:
> pls on JS too

it isn't exported, so it doesn't make sense

PR #1351, opened 2026-05-26
chore(js): add max concurrency limits

jakubno:
Set limit for max concurrent inflight connection, with burst traffic it could happen that the number of connection overwhelms the underlaying infrastructure or at least saturate it to the point each request is too slow to succeed and blocking the rest

### DIFF
--- a/packages/js-sdk/src/api/http2.ts
+++ b/packages/js-sdk/src/api/http2.ts
@@ -1,4 +1,6 @@
 import { runtime } from '../utils'
+import { getEnvVar } from './metadata'
+import { limitConcurrency } from './inflight'
 import {
   loadUndici,
   toUndiciRequestInput,
@@ -6,7 +8,10 @@ import {
   type UndiciRequestInit,
 } from '../undici'
 
-const API_CONNECTION_LIMIT = 100
+const DEFAULT_API_CONNECTION_LIMIT = 100
+// 1000 = ~10 streams per connection (with the 100-conn default).
+// Override via env if your workload needs different.
+const DEFAULT_API_INFLIGHT_LIMIT = 1000
 const API_UNDICI_FALLBACK_WARNING =
   'Failed to load undici for API HTTP/2 transport; falling back to global fetch.'
 
@@ -27,8 +32,9 @@ export function createApiFetchForRuntime(
   currentRuntime = runtime,
   options: {
     connectionLimit?: number
+    inflightLimit?: number
     loadUndici?: () => Promise<UndiciModule | undefined>
-  } = { connectionLimit: API_CONNECTION_LIMIT }
+  } = {}
 ): typeof fetch {
   if (currentRuntime !== 'node') {
     return fetch
@@ -46,6 +52,7 @@ export function createApiFetchForRuntime(
 
 async function buildApiFetcher(options: {
   connectionLimit?: number
+  inflightLimit?: number
   loadUndici?: () => Promise<UndiciModule | undefined>
 }): Promise<typeof fetch> {
   const undici = await (options.loadUndici ?? loadUndici)()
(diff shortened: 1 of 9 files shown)


packages/js-sdk/src/api/http2.ts
```diff
@@ -46,34 +52,75 @@ export function createApiFetchForRuntime(
...
+  return limitConcurrency(fetcher, limit)
+}
+
+export function getApiConnectionLimit(): number {
+  const raw = getEnvVar('E2B_API_CONNECTIONS')
+  if (!raw) return DEFAULT_API_CONNECTION_LIMIT
+
+  const parsed = Number.parseInt(raw, 10)
+  if (!Number.isFinite(parsed) || parsed < 1)
+    return DEFAULT_API_CONNECTION_LIMIT
```
arkamar: Not sure if we should silently ignore the user value. They might be surprised that something does not work as they expect and it would be hard to debug.

packages/js-sdk/src/api/http2.ts
```diff
@@ -46,34 +52,75 @@ export function createApiFetchForRuntime(
...
+ * Defaults to {@link DEFAULT_API_INFLIGHT_LIMIT} ({@link 1000}). Override via
+ * `E2B_API_INFLIGHT_REQUESTS` env var; set to `0` (or any non-positive value)
+ * to disable the cap entirely.
+ */
+export function getApiInflightLimit(): number | null {
+  const raw = getEnvVar('E2B_API_INFLIGHT_REQUESTS')
+  if (!raw) return DEFAULT_API_INFLIGHT_LIMIT
+
+  const parsed = Number.parseInt(raw, 10)
+  if (!Number.isFinite(parsed) || parsed < 1) return null
```
arkamar: The same here.

packages/js-sdk/src/envd/http2.ts
```diff
@@ -117,3 +134,41 @@ export function getEnvdRpcConnectionLimit() {
...
+ * Defaults to {@link DEFAULT_ENVD_INFLIGHT_LIMIT} ({@link 2000}). Override
+ * via `E2B_ENVD_INFLIGHT_REQUESTS` env var; set to `0` (or any non-positive
+ * value) to disable the cap entirely.
+ */
+export function getEnvdInflightLimit(): number | null {
+  const raw = getEnvVar('E2B_ENVD_INFLIGHT_REQUESTS')
+  if (!raw) return DEFAULT_ENVD_INFLIGHT_LIMIT
+
+  const parsed = Number.parseInt(raw, 10)
+  if (!Number.isFinite(parsed) || parsed < 1) return null
```
arkamar: and here

packages/js-sdk/src/envd/http2.ts
```diff
@@ -117,3 +134,41 @@ export function getEnvdRpcConnectionLimit() {
...
+ * Defaults to {@link DEFAULT_ENVD_RPC_INFLIGHT_LIMIT} ({@link 2000}). Override
+ * via `E2B_ENVD_RPC_INFLIGHT_REQUESTS` env var; set to `0` (or any
+ * non-positive value) to disable the cap entirely.
+ */
+export function getEnvdRpcInflightLimit(): number | null {
+  const raw = getEnvVar('E2B_ENVD_RPC_INFLIGHT_REQUESTS')
+  if (!raw) return DEFAULT_ENVD_RPC_INFLIGHT_LIMIT
+
+  const parsed = Number.parseInt(raw, 10)
+  if (!Number.isFinite(parsed) || parsed < 1) return null
```
arkamar: and here.

arkamar (approved):
The main intent looks good. But maybe we should not just silently ignore user preference on error? I'm not going to block it, though.

packages/js-sdk/src/api/inflight.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  const sem = new Semaphore(max)
+
+  return (async (input, init) => {
+    const signal =
+      init?.signal ?? (input instanceof Request ? input.signal : undefined)
+    const release = await sem.acquire(signal)
+    try {
+      return await fetcher(input, init)
```
matthewlouisbrockman, line 73: I think this is releasing on the stream start, not end, so it's gonna allow a bit more concurrency than intended

matthewlouisbrockman (approved):

packages/js-sdk/src/api/inflight.ts
```diff
@@ -0,0 +1,78 @@
...
+    } finally {
+      release()
+    }
```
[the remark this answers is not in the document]
jakubno: It's mentioned in comment above

packages/js-sdk/src/api/metadata.ts
```diff
@@ -27,3 +27,59 @@ export function getEnvVar(name: string) {
...
+export function parseIntEnv(name: string, defaultValue: number): number {
+  const raw = getEnvVar(name)
+  if (!raw) return defaultValue
+
+  const parsed = Number.parseInt(raw, 10)
+  if (!Number.isFinite(parsed)) {
+    throw new Error(
+      `Invalid ${name}=${JSON.stringify(raw)}: expected an integer.`
+    )
+  }
```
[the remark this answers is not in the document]
jakubno: Let's not overengineer this

PR #1354, opened 2026-05-27
fix(js-sdk): preserve multi-source COPY in fromDockerfile

### LINKED ISSUE #1349: fromDockerfile: multi-source COPY silently drops all sources except the first
## Description

`Template().fromDockerfile()` silently drops intermediate sources from multi-source `COPY` instructions. Only the first source is copied; all others are ignored without warning.

## Reproduction

```js
import { Template } from "e2b";

const dockerfile = `FROM ubuntu:22.04
COPY file1.txt file2.txt /dest/`;

const t = Template().fromDockerfile(dockerfile);
const json = JSON.parse(await t.toJSON());

// Expected: two COPY steps (file1.txt → /dest/, file2.txt → /dest/)
//   or one COPY step with both sources
// Actual: one COPY step with only file1.txt → /dest/
console.log(json.steps.filter(s => s.type === "COPY"));
// [{ type: "COPY", args: ["file1.txt", "/dest/", "", ""], force: false }]
```

`file2.txt` is silently dropped.

## Root cause

`handleCopyInstruction` takes `args[0]` as source and `args[args.length - 1]` as destination, ignoring everything in between:

```js
// node_modules/e2b/dist/index.mjs ~line 5025
function handleCopyInstruction(instruction, templateBuilder) {
  const argumentsData = instruction.getArguments();
  if (argumentsData && argumentsData.length >= 2) {
    const src = argumentsData[0].getValue();                       // only first
    const dest = argumentsData[argumentsData.length - 1].getValue(); // last
    templateBuilder.copy(src, dest, { user });
  }
}
```

## Impact

This is a silent data loss — builds succeed but produce broken images. In our case, `COPY gateway/package.json gateway/package-lock.json /opt/qumis/gateway/` resulted in only `package.json` being copied, causing `npm ci` to fail at runtime with a confusing "no package-lock.json" error.

## Expected behavior

Either:
1. Emit a separate `copy()` call for each source (matching Docker semantics), or
2. Throw an error telling the user multi-source COPY is not supported

## Workaround

Split into one COPY per source file:

```dockerfile
COPY gateway/package.json /opt/qumis/gateway/
COPY gateway/package-lock.json /opt/qumis/gateway/
```

## Environment

- `e2b` SDK version: `2.19.2`
- Node.js: 22.x

serejaris:
## Summary
- Fixes `Template().fromDockerfile()` dropping all but the first source in multi-source Dockerfile `COPY` instructions.
- Emits one template `COPY` step per source while preserving the shared destination and COPY flags like `--chown`.
- Adds regression coverage for multi-source COPY and a patch changeset for the JS SDK.

Fixes #1349

## Test plan
- [x] `pnpm --filter e2b exec vitest run tests/template/methods/fromDockerfile.test.ts -t "multi-source COPY"`
- [x] `pnpm --filter e2b exec vitest run tests/template/methods/fromDockerfile.test.ts`
- [x] `pnpm --filter e2b run typecheck`
- [x] `pnpm --filter e2b run lint`
- [x] `pnpm --filter e2b run build`
- [x] `pnpm --filter e2b exec prettier --check src/template/dockerfileParser.ts tests/template/methods/fromDockerfile.test.ts`

### DIFF
--- a/.changeset/fromdockerfile-multi-source-copy.md
+++ b/.changeset/fromdockerfile-multi-source-copy.md
@@ -1,0 +1,5 @@
+---
+"e2b": patch
+---
+
+Preserve all sources from multi-source Dockerfile COPY instructions in fromDockerfile.
--- a/packages/js-sdk/src/template/dockerfileParser.ts
+++ b/packages/js-sdk/src/template/dockerfileParser.ts
@@ -184,8 +184,8 @@ function handleCopyInstruction(
 ): void {
   const argumentsData = instruction.getArguments()
   if (argumentsData && argumentsData.length >= 2) {
-    const src = argumentsData[0].getValue()
     const dest = argumentsData[argumentsData.length - 1].getValue()
+    const sources = argumentsData.slice(0, -1).map((arg) => arg.getValue())
 
     let user: string | undefined
     const flags = instruction.getFlags()
@@ -194,7 +194,9 @@ function handleCopyInstruction(
       user = chownFlag.getValue() ?? undefined
     }
 
-    templateBuilder.copy(src, dest, { user })
+    for (const src of sources) {
+      templateBuilder.copy(src, dest, { user })
+    }
   }
 }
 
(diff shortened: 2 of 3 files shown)


serejaris:
Closing this as superseded by #1355.\n\nI compared the two PRs: this branch only covers the JS  path and is still CLA-blocked, while #1355 covers both JS  and Python , includes the multi-source COPY/ADD regressions in both SDKs, and has the maintainer-owned checks/CLA path.\n\nThanks for picking this up; I will track #1355 instead.

serejaris:
Small correction to the comparison above: this branch covers the JS `Template.fromDockerfile` path, while #1355 covers both JS `Template.fromDockerfile` and Python `Template.from_dockerfile`.

PR #1355, opened 2026-05-27
fix(sdks): handle multi-source COPY/ADD in fromDockerfile

### LINKED ISSUE #1349: fromDockerfile: multi-source COPY silently drops all sources except the first

mishushakov:
## Summary

- Fixes #1349: `Template.fromDockerfile` (JS) and `Template.from_dockerfile` (Python) silently dropped intermediate sources from multi-source `COPY`/`ADD`, keeping only the first one and producing broken images without warning.
- Both parsers now emit one `copy()` call per source to the same destination (matching Docker semantics), preserving `--chown` across all calls.
- Added tests in both SDKs (multi-source COPY, and multi-source COPY with `--chown`), plus changesets for `e2b` and `@e2b/python-sdk`.

## Test plan

- [x] `pnpm run test tests/template/methods/fromDockerfile.test.ts` (JS)
- [x] `poetry run pytest tests/{async,sync}/template_*/methods/test_from_dockerfile.py` (Python)
- [x] `pnpm run format` / `pnpm run lint`

### DIFF
--- a/.changeset/js-sdk-multi-source-copy.md
+++ b/.changeset/js-sdk-multi-source-copy.md
@@ -1,0 +1,5 @@
+---
+'e2b': patch
+---
+
+fix(js-sdk): `Template.fromDockerfile` now correctly handles multi-source `COPY`/`ADD` instructions. Previously, only the first source was kept and all intermediate sources were silently dropped; now each source is emitted as its own `copy()` call to the same destination.
--- a/.changeset/js-sdk-multi-source-copy.md
+++ b/.changeset/js-sdk-multi-source-copy.md
@@ -1,5 +1,0 @@
----
-'e2b': patch
----
-
-fix(js-sdk): `Template.fromDockerfile` now correctly handles multi-source `COPY`/`ADD` instructions. Previously, only the first source was kept and all intermediate sources were silently dropped; now each source is emitted as its own `copy()` call to the same destination.
--- a/.changeset/multi-source-copy.md
+++ b/.changeset/multi-source-copy.md
@@ -1,0 +1,6 @@
+---
+'e2b': patch
+'@e2b/python-sdk': patch
+---
+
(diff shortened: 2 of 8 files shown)


matthewlouisbrockman (approved):
tests only cover COPY not ADD but not blocking

PR #1356, opened 2026-05-27
feat(sdk): validate E2B API key format client-side

mishushakov:
## Summary

- Both JS and Python SDKs now validate that the configured E2B API key matches `e2b_` followed by 40 hex characters (mirroring the server-side check in [`infra/.../keys/key.go`](https://github.com/e2b-dev/infra/blob/main/packages/shared/pkg/keys/key.go#L66)) and throw `AuthenticationError` / `AuthenticationException` with an example token (`e2b_0000…`) and a link to the API Keys dashboard tab.
- Validation runs inside `ApiClient` / `ApiClient.__init__` whenever an API key is present, so callers get immediate, actionable feedback instead of a generic 401 from the server.
- Added unit tests (`validateApiKey.test.ts`, `test_validate_api_key.py`) and updated existing fixtures that used placeholder keys like `'test-key'` / `'base-api-key'` to use the valid format.

## Test plan

- [x] `pnpm run format`, `pnpm run lint`, `pnpm run typecheck`
- [x] `pnpm exec vitest run tests/api/validateApiKey.test.ts tests/api/handleApiError.test.ts tests/sandbox/abortSignal.test.ts tests/template/abortSignal.test.ts tests/sandbox/configPropagation.test.ts tests/connectionConfig.test.ts`
- [x] `poetry run pytest tests/test_validate_api_key.py tests/test_api_client_transport.py tests/sync/sandbox_sync/test_config_propagation.py tests/async/sandbox_async/test_config_propagation.py tests/test_connection_config.py`

### DIFF
--- a/.changeset/validate-api-key-format.md
+++ b/.changeset/validate-api-key-format.md
@@ -1,0 +1,6 @@
+---
+'e2b': patch
+'@e2b/python-sdk': patch
+---
+
+Validate the E2B API key format client-side. SDKs now throw an `AuthenticationError` / `AuthenticationException` with an example token (e.g. `e2b_0000000000000000000000000000000000000000`) when the key does not start with `e2b_` or is not followed by 40 hex characters.
--- a/.changeset/validate-api-key-format.md
+++ b/.changeset/validate-api-key-format.md
@@ -3,4 +3,4 @@
 '@e2b/python-sdk': patch
 ---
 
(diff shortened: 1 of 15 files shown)


matthewlouisbrockman (approved):
looks fine; it looks like we have the base api key though, can we use that const everywhere if possible instead of hardcoding it in places?

PR #1359, opened 2026-05-28
fix(python-sdk): remove dead code in ConnectionConfig.__init__

devteamaegis:
## Problem

In `ConnectionConfig.__init__`, `_get_request_timeout` is called and its
result assigned to `self.request_timeout`, but that assignment is
immediately overwritten by an inline `if/elif/else` block that duplicates
the same logic:

```python
# lines 102-105 — result is never used
self.request_timeout = ConnectionConfig._get_request_timeout(
    REQUEST_TIMEOUT,
    request_timeout,
)

# lines 107-112 — always overwrites the above
if request_timeout == 0:
    self.request_timeout = None
elif request_timeout is not None:
    self.request_timeout = request_timeout
else:
    self.request_timeout = REQUEST_TIMEOUT
```

`_get_request_timeout` already implements all three cases identically, so
lines 107–112 are unreachable dead code. This creates a misleading
impression that the static method is *checked* before the inline fallback,
and silently breaks if someone later changes `_get_request_timeout`
without noticing the duplicate.

## Fix

Remove the inline `if/elif/else` block and rely solely on
`_get_request_timeout`, which is already the method used by
`get_request_timeout` for per-call overrides.

## Tests

Added six unit tests to `tests/test_connection_config.py` covering:
- default timeout (60 s)
- explicit timeout value
- `0` → `None` (no limit)
- `get_request_timeout()` with no argument inherits configured default
- `get_request_timeout(override)` returns the override
- `get_request_timeout(0)` returns `None` regardless of config

### DIFF
--- a/.changeset/remove-dead-request-timeout-code.md
+++ b/.changeset/remove-dead-request-timeout-code.md
@@ -1,0 +1,5 @@
+---
+'@e2b/python-sdk': patch
+---
+
+Remove dead code in `ConnectionConfig.__init__` that duplicated the logic of `_get_request_timeout` and immediately overwrote its result. Behavior is unchanged.
--- a/packages/python-sdk/e2b/connection_config.py
+++ b/packages/python-sdk/e2b/connection_config.py
@@ -104,13 +104,6 @@ class ConnectionConfig:
             request_timeout,
         )
 
-        if request_timeout == 0:
-            self.request_timeout = None
-        elif request_timeout is not None:
-            self.request_timeout = request_timeout
-        else:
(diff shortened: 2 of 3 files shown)


mishushakov (approved):

mishushakov:
hey @devteamaegis looks good!
could you please sign the CLA?

would like to proceed with merge

mishushakov:
@devteamaegis reminder to sign CLA!

mishushakov:
CLA not signed, closing stale PR.

PR #1360, opened 2026-05-28
feat(sdk): allow disabling client-side API key validation

mishushakov:
## Summary

Allow disabling client-side API key **format** validation. Previously the SDKs hard-required keys to match the `e2b_<hex>` pattern, which blocked deployments that issue API keys with a different format. Instead of a custom-prefix override, this adds a simple on/off toggle.

The default behaviour is unchanged (validation stays **on**).

## Configuration

| Form | JS | Python |
| --- | --- | --- |
| Env var | `E2B_VALIDATE_API_KEY=false` | `E2B_VALIDATE_API_KEY=false` |
| Connection option | `validateApiKey: false` | `validate_api_key=False` |

The connection option takes priority over the environment variable.

## Usage

**JavaScript / TypeScript**

```ts
import { Sandbox } from 'e2b'

// Via connection option
  apiKey: 'custom_key_format',
  validateApiKey: false,
})

// Or via env var: E2B_VALIDATE_API_KEY=false
```

**Python**

```python
from e2b import Sandbox

# Via connection option
sandbox = Sandbox(
    api_key="custom_key_format",
    validate_api_key=False,
)

# Or via env var: E2B_VALIDATE_API_KEY=false
```


huv1k (commented):
Why do we need this?

mishushakov:
@huv1k some self-hosted setups have a different API Key formats - we want to support it.

huv1k (approved):

PR #1361, opened 2026-05-28
feat(cli): accept E2B_API_KEY in template list & create

mishushakov:
## Summary

`E2B_ACCESS_TOKEN` is deprecated, so CLI commands whose endpoints accept either credential now authenticate with `E2B_API_KEY` instead of requiring an access token.

- `e2b template list` now uses `ensureAPIKey()`. The underlying `GET /templates` endpoint accepts both `ApiKeyAuth` and `AccessTokenAuth`, and since the access token is deprecated we standardize on the API key.
- `e2b template create` no longer calls `ensureAccessToken()`. Its only API calls — `POST /v3/templates` and `POST /v2/templates/{id}/builds/{bid}` (via the SDK's `Template.build`) — accept only `ApiKeyAuth`, so requiring an access token locked out API-key-only environments for no reason.
- `e2b template build` is intentionally left as-is: its v1 endpoints and docker-registry login are access-token-only at the API level.
- Removed the now-unused `ensureAccessTokenOrAPIKey()` helper and the `'BOTH'` variant of the auth-error box that an earlier iteration of this PR introduced.
- Adds a real backend-integration test in `tests/commands/template/create.test.ts` that mirrors the existing `backend_integration.test.ts` pattern: use the real `E2B_DOMAIN` and assert end-to-end that `template create` succeeds with only `E2B_API_KEY` set (no `E2B_ACCESS_TOKEN`). Uses a unique template name per run and cleans up the created template in `afterAll`.

## Test plan

- [x] `pnpm --filter @e2b/cli run typecheck`
- [x] `pnpm --filter @e2b/cli run lint`
- [x] `pnpm --filter @e2b/cli run format`
- [x] `pnpm --filter @e2b/cli run test` (local, with real `E2B_API_KEY` — new test passes; create succeeds without `E2B_ACCESS_TOKEN`)


huv1k (approved):
Shouldn't we migrate everything in one go? I think it's not patch 😅

mishushakov:
@huv1k I have looked at other methods as well, but this was the only incorrect one.

PR #1362, opened 2026-05-28
refactor(sdk): import API types via components instead of paths

mishushakov:
## Summary
- Replace verbose `paths['/route']['method']['responses'|'requestBody'][...]` traversal with direct `components['schemas'][...]` references in `packages/js-sdk/src/template/buildApi.ts` and `packages/cli/src/commands/template/build.ts`.
- Matches the existing convention used throughout `sandboxApi.ts` and reads at a glance.

## Test plan
- [x] `pnpm run format` / `lint` / `typecheck` pass for `e2b` and `@e2b/cli`


huv1k (approved):
Looks good

PR #1372, opened 2026-05-31
fix(parse): guard against branch names containing "..." in parse_git_status

### LINKED ISSUE #1371: BUG: parse_git_status crashes with ValueError when branch name contains "..."
When git reports a branch whose name itself contains the substring `...` (e.g. `feat...v2`) with a tracking upstream, the porcelain output looks like `## feat...v2...origin/feat...v2`. The line `branch, upstream_branch = normalized_branch.split("...")` in `packages/python-sdk/e2b/sandbox/_git/parse.py` splits on every occurrence of `...`, producing more than two parts, and the two-variable unpacking immediately raises `ValueError: too many values to unpack (expected 2)`.

Traceback:
```
  File "parse.py", line 135, in parse_git_status
    branch, upstream_branch = normalized_branch.split("...")
ValueError: too many values to unpack (expected 2)
```

To reproduce, run: `python3 -c "from e2b.sandbox._git.parse import parse_git_status; parse_git_status('## feat...v2...origin/feat...v2\n')"`

devteamaegis:
## What's broken

`parse_git_status` in `packages/python-sdk/e2b/sandbox/_git/parse.py` crashes with `ValueError: too many values to unpack (expected 2)` when git reports a branch whose name contains the substring `...`. For example, a branch named `feat...v2` tracking `origin/feat...v2` produces porcelain output `## feat...v2...origin/feat...v2`. The call `normalized_branch.split("...")` splits on every occurrence and yields more than two parts, causing the two-variable unpacking to fail immediately.

## Why it happens

`str.split(sep)` with no `maxsplit` argument splits on every occurrence of the separator, so branch names that contain `...` produce more than two segments.

## Fix

Changed `normalized_branch.split("...")` to `normalized_branch.split("...", 1)` so only the first `...` is used to separate the local branch name from the upstream tracking ref, regardless of how many `...` appear in the branch name.

## Test

Added `tests/test_git_parse.py` with one test that feeds porcelain output for a `...`-containing branch into `parse_git_status` and asserts it returns without raising and produces the correct `current_branch` and `upstream` values.

Fixes #1371

### DIFF
--- a/packages/python-sdk/e2b/sandbox/_git/parse.py
+++ b/packages/python-sdk/e2b/sandbox/_git/parse.py
@@ -132,7 +132,7 @@ def parse_git_status(output: str) -> GitStatus:
         if is_detached or normalized_branch.startswith("HEAD"):
             detached = True
         elif "..." in normalized_branch:
-            branch, upstream_branch = normalized_branch.split("...")
+            branch, upstream_branch = normalized_branch.split("...", 1)
             current_branch = branch or None
             upstream = upstream_branch or None
         else:
--- a/packages/python-sdk/tests/test_git_parse.py
+++ b/packages/python-sdk/tests/test_git_parse.py
@@ -1,0 +1,11 @@
+from e2b.sandbox._git.parse import parse_git_status
+
+
+def test_parse_git_status_branch_name_with_ellipsis():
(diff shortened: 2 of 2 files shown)


ValentaTomas:

PR #1378, opened 2026-06-02
feat(sdk): retry transient network errors and rate limits

jakubno:
Automatically retry requests on transient failures across the JS and Python SDKs. Retries connection errors and 429/502/503/504 responses using exponential backoff with jitter, and honor a server-provided Retry-After header so rate limiting (e.g. listing sandboxes) is handled transparently.

Retries are idempotency-aware: idempotent methods retry on any transient failure, while non-idempotent ones (e.g. Sandbox.create) only retry on "rejected" failures where the server provably did not process the request (throttling, connection-refused, DNS), avoiding duplicate side effects.

Configure via the new `retries` option or E2B_MAX_RETRIES env var (default 3, set 0 to disable). The Python envd RPC retry now also uses backoff between attempts.


packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+ * {@link DEFAULT_MAX_RETRIES}.
+ */
+export function resolveMaxRetries(retries?: number): number {
+  if (retries !== undefined) {
+    if (!Number.isInteger(retries) || retries < 0) {
+      throw new Error(
+        `Invalid retries=${retries}: expected a non-negative integer.`
+      )
+    }
+    return retries
```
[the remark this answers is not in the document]
jakubno: This looks like a false positive. `Number.isInteger(false) and Number.isInteger(true) are both false (booleans aren't numbers), so resolveMaxRetries already throws on true/false:
resolveMaxRetries(false) // throws: Invalid retries=false: expected a non-negative integer.
resolveMaxRetries(true)  // throws: Invalid retries=true: expected a non-negative integer.
So JS and Python behave the same (both reject booleans). No change needed — please resolve unless I'm missing a case.

packages/python-sdk/e2b_connect/client.py
```diff
@@ -281,10 +334,29 @@ async def acall_unary(
...
+                continue
+
+            if res.status == _REJECTED_STATUS and attempt < retries:
+                retry_after = _retry_after_seconds(res.headers)
+                delay = (
+                    retry_after if retry_after is not None else _backoff_delay(attempt)
+                )
+                await asyncio.sleep(delay)
+                attempt += 1
+                continue
```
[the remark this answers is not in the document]
jakubno: This is a false positive for this code path. The RPC client uses httpcore's convenience pool.request() / async_pool.request(), which read and close the response before returning (httpcore 1.0.9):
```py
response = self.handle_request(request)
try:
    response.read()
finally:
    response.close()
return response
```

So by the time the 429 branch runs, the body is already drained and the connection released to the pool — sleeping and continuing can't stall it. The explicit-drain pattern in e2b._retry/withRetry is needed there because those layers receive un-drained httpx/fetch responses; the connect client sits above request(), which drains for it. No change needed — please resolve.

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+    if (typeof e.code === 'string') codes.push(e.code.toLowerCase())
+    if (typeof e.name === 'string') codes.push(e.name.toLowerCase())
+    current = e.cause
+  }
+
+  for (const code of codes) {
+    const kind = RETRYABLE_ERROR_CODES.get(code)
+    if (kind) return kind
+    // undici low-level socket/transport errors are ambiguous mid-flight drops.
+    if (code.startsWith('und_err_') || code === 'fetch failed') {
```
wj-e2b, line 181: Should probably check `.message` here since `fetch failed` isn't set in `.code`.

https://github.com/nodejs/undici/blob/c995513094903c67151907213296e91179279b50/lib/web/fetch/index.js#L261

wj-e2b (approved):

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+export function computeDelayMs(
+  attempt: number,
+  policy: RetryPolicy,
+  response?: Response
+): number {
+  const retryAfter = parseRetryAfter(
+    response?.headers.get('retry-after') ?? null
+  )
+  if (retryAfter !== undefined) {
+    return Math.min(retryAfter, policy.backoffCapMs * 4)
```
mishushakov, line 147: it doesn't make much sense?<br>if server says you should retry after x, but it's > `policy.backoffCapMs * 4` when we retry it will fail again - maybe makes sense to mark the request as not retryable at the time if retryAfter > `policy.backoffCapMs * 4` - wdyt?

.changeset/retry-transient-errors-js.md
mishushakov, line 1: merge python changeset into it

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+function retryableStatusKind(status: number): FailureKind | undefined {
+  return RETRYABLE_STATUS.get(status)
+}
+
+/**
+ * Classify a thrown error as a transient, connection-level failure (returning
+ * its {@link FailureKind}) or `undefined` when it is not retryable. User/timeout
+ * aborts are explicitly not retryable.
+ */
+export function retryableErrorKind(err: unknown): FailureKind | undefined {
```
mishushakov, line 164: : bool, no undefined

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
+import { parseIntEnv } from './api/metadata'
+import { wait } from './utils'
+
+/** Default number of *retries* (i.e. attempts after the first). */
+export const DEFAULT_MAX_RETRIES = 3
```
mishushakov, line 5: it's exported but not imported anywhere else?

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+ * its {@link FailureKind}) or `undefined` when it is not retryable. User/timeout
+ * aborts are explicitly not retryable.
+ */
+export function retryableErrorKind(err: unknown): FailureKind | undefined {
+  if (isAbortError(err)) return undefined
+
+  const codes: string[] = []
+  let current: unknown = err
+  // Walk the `cause` chain (undici wraps the real error in `cause`).
+  for (let i = 0; i < 5 && current; i++) {
```
mishushakov, line 170: i will have to check undici, but the for loop looks weird

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
+import { parseIntEnv } from './api/metadata'
+import { wait } from './utils'
+
+/** Default number of *retries* (i.e. attempts after the first). */
+export const DEFAULT_MAX_RETRIES = 3
+
+/** Base for the exponential backoff, in milliseconds. */
+const DEFAULT_BACKOFF_BASE_MS = 100
```
mishushakov, line 8: should be higher imo, 500ms?

packages/js-sdk/src/retry.ts
mishushakov, line 181: have to recheck, fetch failed is correct

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+ * so the request can be replayed across retries. Larger bodies (e.g. file
+ * uploads) are sent once and not retried.
+ */
+const MAX_REPLAYABLE_BODY_BYTES = 1024 * 1024 // 1 MiB
+
+/**
+ * HTTP methods that are idempotent per the HTTP spec and can therefore be
+ * retried on any transient failure.
+ */
+const IDEMPOTENT_METHODS = new Set(['GET', 'HEAD', 'OPTIONS', 'PUT', 'DELETE'])
```
mishushakov, line 22: PUT might not be idempotent, check: PUT /volumecontent/{volumeID}/file

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+  return undefined
+}
+
+/**
+ * Single decision point for both the response and error paths: retry only when
+ * attempts remain, the failure is transient, and the failure is safe to replay
+ * for this request. `rejected` failures are always safe; `ambiguous` failures
+ * are only safe for idempotent requests.
+ */
+function shouldRetry(
```
mishushakov, line 195: rm

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+/**
+ * A transient failure is either:
+ *  - `rejected`: the server demonstrably did NOT process the request (e.g. it
+ *    was throttled, or the connection never reached the server). Replaying is
+ *    always safe, even for non-idempotent requests.
+ *  - `ambiguous`: the request may or may not have been processed (e.g. the
+ *    connection dropped after sending). Replaying is only safe for idempotent
+ *    requests.
+ */
+type FailureKind = 'rejected' | 'ambiguous'
```
mishushakov, line 33: i think this is completely unnecessary, if it's retryable - we retry, if not - we don't

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+  kind: FailureKind | undefined,
+  attempt: number,
+  policy: RetryPolicy,
+  idempotent: boolean
+): boolean {
+  if (kind === undefined || attempt >= policy.retries) return false
+  return kind === 'rejected' || idempotent
+}
+
+function isAbortError(err: unknown): boolean {
```
mishushakov, line 205: confusing naming we also check for timeout error here

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+ *
+ * Keep in sync with `_RETRYABLE_STATUS` in the Python SDK (`e2b/_retry.py`).
+ */
+export const RETRYABLE_STATUS: ReadonlyMap<number, FailureKind> = new Map([
+  [408, 'ambiguous'], // request timeout
+  [429, 'rejected'], // throttled — not processed
+  [502, 'ambiguous'], // bad gateway
+  [503, 'ambiguous'], // service unavailable — may be returned after processing
+  [504, 'ambiguous'], // gateway timeout
+])
```
mishushakov, line 51: these statuses are not documented in our client API spec, so you're basically guessing which status code google cloud lb / proxies could emit

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+  if (retryAfter !== undefined) {
+    return Math.min(retryAfter, policy.backoffCapMs * 4)
+  }
+
+  const exp = Math.min(policy.backoffCapMs, policy.backoffBaseMs * 2 ** attempt)
+  // Full jitter: a random value in [0, exp].
+  return Math.random() * exp
+}
+
+function retryableStatusKind(status: number): FailureKind | undefined {
```
mishushakov, line 155: the function only has one caller

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+  if (retries !== undefined) {
+    if (!Number.isInteger(retries) || retries < 0) {
+      throw new Error(
+        `Invalid retries=${retries}: expected a non-negative integer.`
+      )
+    }
+    return retries
+  }
+
+  const fromEnv = parseIntEnv('E2B_MAX_RETRIES', DEFAULT_MAX_RETRIES)
```
mishushakov, line 91: can't you just remove the above if block and instead do:

```
const fromEnv = parseIntEnv('E2B_MAX_RETRIES', retries || DEFAULT_MAX_RETRIES)
```

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+function defaultPolicy(retries: number): RetryPolicy {
+  return {
+    retries,
+    backoffBaseMs: DEFAULT_BACKOFF_BASE_MS,
+    backoffCapMs: DEFAULT_BACKOFF_CAP_MS,
+  }
+}
+
+/**
+ * Parse a `Retry-After` header value (either delta-seconds or an HTTP date)
```
mishushakov, line 109: not in the spec as well

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+  if (kind === undefined || attempt >= policy.retries) return false
+  return kind === 'rejected' || idempotent
+}
+
+function isAbortError(err: unknown): boolean {
+  const name = (err as { name?: unknown } | null)?.name
+  return name === 'AbortError' || name === 'TimeoutError'
+}
+
+function isStreamLike(body: unknown): boolean {
```
mishushakov, line 210: prob unnecessary (explain below)

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+  // delta-seconds form
+  if (/^\d+$/.test(trimmed)) {
+    return Number.parseInt(trimmed, 10) * 1000
+  }
+
+  // HTTP-date form
+  const date = Date.parse(trimmed)
+  if (!Number.isNaN(date)) {
+    return Math.max(0, date - now)
+  }
```
mishushakov, line 129: why are there two different types of dates? pls document where both could come from (eg. google cloud lb returns x, client proxy returns y)

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+
+  if (policy.retries <= 0) {
+    return innerFetch
+  }
+
+  return (async (input, init) => {
+    // A raw streaming body (e.g. connect-web server/bidi streams) cannot be
+    // replayed and would throw when constructing a Request without `duplex`.
+    // Send it once, unwrapped.
+    if (isStreamLike(init?.body)) {
```
mishushakov, line 303: can just check `body instanceof ReadableStream`

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+ * Compute the delay before the next attempt. A server-provided `Retry-After`
+ * takes precedence; otherwise an exponential backoff with full jitter is used.
+ */
+export function computeDelayMs(
+  attempt: number,
+  policy: RetryPolicy,
+  response?: Response
+): number {
+  const retryAfter = parseRetryAfter(
+    response?.headers.get('retry-after') ?? null
```
mishushakov, line 144: ?? undefined

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+
+/**
+ * Buffer a request body up to {@link MAX_REPLAYABLE_BODY_BYTES} so the request
+ * can be replayed across retries. Reads the (cloned) body incrementally and
+ * bails out as soon as the cap is exceeded so we never buffer large uploads in
+ * memory. Returns `undefined` body for bodyless requests, and
+ * `replayable: false` when the body is too large to buffer safely (in which
+ * case the request is sent once, without retries).
+ */
+async function bufferBody(
```
mishushakov, line 226: this might be a stretch, would just retry as-is, even if it's larger bodies.

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+ * @param retries the number of retries, resolved via {@link resolveMaxRetries}
+ *   by the caller; `0` disables retries (the inner fetch is returned unwrapped).
+ */
+export function withRetry(
+  innerFetch: typeof fetch,
+  retries: number
+): typeof fetch {
+  const policy = defaultPolicy(retries)
+
+  if (policy.retries <= 0) {
```
mishushakov, line 295: retries can't be <0

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+    // input's body.
+    if (!replayable) {
+      return innerFetch(request)
+    }
+
+    const buildAttempt = (): Request =>
+      body !== undefined ? new Request(request, { body }) : request.clone()
+
+    let attempt = 0
+    for (;;) {
```
mishushakov, line 325: is it just a while loop

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+      try {
+        const response = await innerFetch(buildAttempt())
```
mishushakov, line 327: keep in mind, this might pollute the memory with Request objects

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+
+        const kind = retryableStatusKind(response.status)
+        if (!shouldRetry(kind, attempt, policy, idempotent)) {
+          return response
+        }
+
+        // Drain and discard the body so the connection can be reused.
+        await response.body?.cancel().catch(() => {})
```
mishushakov, line 335: isn't this handled by the V8 GC, not sure you need to manually drain it

packages/js-sdk/src/retry.ts
```diff
@@ -0,0 +1,352 @@
...
+  backoffBaseMs: number
+  backoffCapMs: number
+}
+
+/**
+ * Resolve the configured number of retries, falling back to the
+ * `E2B_MAX_RETRIES` environment variable and finally
+ * {@link DEFAULT_MAX_RETRIES}.
+ */
+export function resolveMaxRetries(retries?: number): number {
```
mishushakov, line 81: this should perhaps live on connectionConfig directly, not here

packages/js-sdk/src/volume/client.ts
```diff
@@ -110,6 +124,9 @@ class VolumeApiClient {
   constructor(config: VolumeConnectionConfig) {
     this.api = createClient<paths>({
       baseUrl: config.apiUrl,
+      // Volume content uploads/downloads can be large; withRetry only retries
+      // small, replayable bodies.
```
mishushakov, line 128: i would check specific URL patterns, for example PUTs on the sandbox fs  / volumes and mark these as non-retryable instead.<br>would also remove the need for the buffer logic altogether.

packages/js-sdk/tests/retry.test.ts
```diff
@@ -0,0 +1,338 @@
...
+  RETRYABLE_ERROR_CODES,
+} from '../src/retry'
+
+const policy = { retries: 3, backoffBaseMs: 500, backoffCapMs: 8_000 }
+
+/**
+ * Build a fake `fetch` that returns/throws the queued outcomes in order and
+ * records the requests it received.
+ */
+function fakeFetch(
```
mishushakov, line 18: this feels kinda cursed, i think should instead use our actual fetch and mock the responses w MSW

mishushakov (commented):
JS SDK reviewed

packages/js-sdk/src/retry.ts
mishushakov: also, why have this function when you only call it from one place

ValentaTomas:

PR #1380, opened 2026-06-02
ci: skip CI jobs when unaffected by changed paths

mishushakov:
Adds a `dorny/paths-filter` change-detection job to the PR-triggered workflows so jobs only run when relevant paths change: Lint/Typecheck run only when package code, spec, or lint configs change; Generated files runs only when codegen inputs/outputs change; and the JS/Python/CLI SDK tests run only when the respective SDK changes (CLI also runs on JS SDK changes since it builds against it). The SDK test jobs are gated *inside* the reusable workflows via a new `run` input rather than by skipping the caller, so the required matrix status checks still report (skipped jobs report success) and branch protection stays satisfied. Shared paths (spec, lockfiles, `package.json`, `.tool-versions`) and `workflow_dispatch` runs still trigger everything.


matthewlouisbrockman (approved):
i like it! if we start seeing weird issues from missing issues on loosely related code can revisit but hopefully makes faster

PR #1383, opened 2026-06-04
feat(sdks): expose user-defined file metadata on sandbox.files

mishushakov:
Adds a `metadata` option to file uploads and surfaces persisted metadata on every `EntryInfo` / `WriteInfo` returned by `getInfo`, `list`, `rename`, and write responses, across the JS and Python (sync + async) SDKs.

Metadata is sent as `X-Metadata-<key>: <value>` request headers and persisted by envd as `user.e2b.*` extended attributes; the same map is applied to every file in a multi-file upload. Keys and values must be printable US-ASCII and keys are lowercased by the sandbox, so they may differ in case when read back. Requires **envd 0.6.2 or later**.

This syncs the envd OpenAPI spec and filesystem proto with [infra#2732](https://github.com/e2b-dev/infra/pull/2732) and regenerates the JS/Python clients.

## Usage

**JavaScript / TypeScript**
```ts
// Single file
const info = await sandbox.files.write('report.txt', 'hello', {
  metadata: { author: 'mish', purpose: 'demo' },
})
console.log(info.metadata) // { author: 'mish', purpose: 'demo' }

// Multiple files (same metadata applied to each)
await sandbox.files.writeFiles(
  [
    { path: 'a.txt', data: 'A' },
    { path: 'b.txt', data: 'B' },
  ],
  { metadata: { source: 'import' } }
)

// Read it back
const stat = await sandbox.files.getInfo('report.txt')
console.log(stat.metadata) // { author: 'mish', purpose: 'demo' }
```

**Python**
```python
# Single file
info = sandbox.files.write("report.txt", "hello", metadata={"author": "mish"})
print(info.metadata)  # {"author": "mish"}

# Multiple files (same metadata applied to each)
sandbox.files.write_files(
    [
        WriteEntry(path="a.txt", data="A"),
        WriteEntry(path="b.txt", data="B"),
    ],
    metadata={"source": "import"},
)

# Read it back
stat = sandbox.files.get_info("report.txt")
print(stat.metadata)  # {"author": "mish"}
```

The async Python API is identical with `await`.

## Tests

Integration tests cover the round-trip across `write` / `getInfo` / `list` / `rename`, octet-stream uploads, multi-file uploads, overwrite-clears-stale-metadata, and metadata written directly as `user.e2b.*` xattrs via `sandbox.commands.run` surfacing in `getInfo`. They require a sandbox running envd 0.6.2+.


packages/js-sdk/src/sandbox/filesystem/index.ts
```diff
@@ -431,13 +466,27 @@ export class Filesystem {
...
 
+    const metadata = writeOpts?.metadata
+    if (
+      metadata &&
+      Object.keys(metadata).length > 0 &&
+      compareVersions(this.envdApi.version, ENVD_FILE_METADATA) < 0
+    ) {
+      throw new TemplateError(
+        'File metadata requires envd 0.5.26 or later. ' +
+          'You can update the template by running `e2b template build` in the directory with the template.'
```
mishushakov: remove this part of the comment

packages/python-sdk/e2b/sandbox_async/filesystem/filesystem.py
```diff
@@ -257,9 +274,17 @@ async def write_files(
         if len(files) == 0:
             return []
 
+        if metadata and self._envd_version < ENVD_FILE_METADATA:
+            raise TemplateException(
+                "File metadata requires envd 0.6.2 or later. "
+                "You can update the template by running `e2b template build` in the directory with the template."
```
mishushakov: remove this part of the comment

packages/python-sdk/e2b/sandbox_async/filesystem/filesystem.py
```diff
@@ -58,6 +61,15 @@ async def _ahandle_filesystem_envd_api_exception(r):
     return await ahandle_envd_api_exception(r, _FILESYSTEM_HTTP_ERROR_MAP)
 
 
+def _write_info_from_dict(payload: Dict) -> WriteInfo:
```
mishushakov: i think you can instead add method WriteInfo.from_dict()

packages/js-sdk/tests/sandbox/files/metadata.test.ts
```diff
@@ -0,0 +1,136 @@
...
+sandboxTest(
+  'writeFiles applies metadata to every file',
+  async ({ sandbox }) => {
+    const metadata = { source: 'test-suite' }
+    const files: WriteEntry[] = [
+      { path: 'metadata_multi_1.txt', data: 'File 1' },
+      { path: 'metadata_multi_2.txt', data: 'File 2' },
+    ]
+
+    const infos = await sandbox.files.writeFiles(files, { metadata })
```
mishushakov, line 71: we should set metadata on the WriteEntry instead, instead on writeFiles method

packages/js-sdk/tests/sandbox/files/metadata.test.ts
mishushakov, line 1: we need to also add a test that would set metadata using xattrs via sandbox.commands.run in the user.e2b namespace and then these should surface when retrieving via sandbox.files.getInfo

mishushakov: same for python tests

packages/js-sdk/tests/sandbox/files/metadata.test.ts
mishushakov: not doing this - in case of multipart upload - we cannot have metadata per part

matthewlouisbrockman (approved):
approving, may want to address claudes comment, also the instructions for the allowed values aren't quite accurate because it's not all ascii allowed because it's passed in the headers

mishushakov:
I've added validation

mishushakov:
@cla-bot check

PR #1385, opened 2026-06-04
feat(sdk): add includeEntry option to filesystem watch

mishushakov:
Client-side counterpart to [e2b-dev/infra#2930](https://github.com/e2b-dev/infra/pull/2930): adds an `includeEntry`/`include_entry` option to filesystem directory watching across the JS and Python (sync + async) SDKs, so each `FilesystemEvent` can carry the affected entry's `EntryInfo` (best-effort — unset for remove/rename-away events where the path no longer exists). This regenerates the filesystem proto code from the updated spec, threads the flag through `watchDir`/`watch_dir` (streaming `WatchDir` and polling `CreateWatcher`), maps the new `entry` field onto the event, and extracts a shared entry-mapping helper reused by `list`/`getInfo`/`rename`. The option degrades gracefully: older sandboxes (< envd 0.6.2) ignore it and leave `entry` unset, so there's no hard version gate. Includes new watch tests for all three SDKs and a minor-bump changeset for `e2b` and `@e2b/python-sdk`.

> Note: the entry-info tests require envd 0.6.2 (shipped by the infra PR), so this should land with/after that deploy.

### Usage

**JavaScript**
```ts
const handle = await sandbox.files.watchDir(
  'my-dir',
  (event) => {
    console.log(event.type, event.name, event.entry?.path, event.entry?.type)
  },
  { includeEntry: true }
)
```

**Python (async)**
```python
def on_event(e):
    print(e.type, e.name, e.entry.path if e.entry else None)

handle = await sandbox.files.watch_dir("my-dir", on_event=on_event, include_entry=True)
```

**Python (sync)**
```python
handle = sandbox.files.watch_dir("my-dir", include_entry=True)
for e in handle.get_new_events():
    print(e.type, e.name, e.entry.path if e.entry else None)
```


matthewlouisbrockman (approved):
approving _but_ there's probably going to be a merge conflict with 6.2; whatever gets merged 2nd is gonna need to bump to 6.3

PR #1386, opened 2026-06-05
feat(js-sdk): add proxy connection param

mishushakov:
Adds a `proxy` connection parameter to the JS SDK, mirroring the Python SDK. When set, requests are routed through the given HTTP proxy via an undici `ProxyAgent` dispatcher (fetchers are cached per-proxy so non-proxy traffic is unaffected). It applies to control-plane API requests, all requests made to the returned sandbox (REST plus filesystem/commands/pty RPC), and volume requests. Behavior is unchanged when no proxy is provided, and unit tests cover both the API and envd fetch paths.

## Usage

```ts
import { Sandbox } from 'e2b'

// Routes API + all sandbox requests through the proxy
  proxy: 'http://user:pass@127.0.0.1:8080',
})
await sandbox.files.write('/hello.txt', 'world')

// Also works when connecting to an existing sandbox
const sbx = await Sandbox.connect(sandboxId, { proxy: 'http://127.0.0.1:8080' })
```

> Proxying relies on the optional `undici` package and the Node runtime; in browser/edge runtimes requests use global `fetch`, which has no proxy support (same as the existing HTTP/2 dispatcher).


packages/js-sdk/src/volume/client.ts
```diff
@@ -110,6 +120,9 @@ class VolumeApiClient {
   constructor(config: VolumeConnectionConfig) {
     this.api = createClient<paths>({
       baseUrl: config.apiUrl,
+      // Only route through the proxy-aware fetch when a proxy is set, so the
+      // default fetch behavior is preserved otherwise.
+      ...(config.proxy && { fetch: createApiFetch(config.proxy) }),
```
mishushakov: i think we should just use createApiFetch for consistency here

mishushakov: Done — now using `createApiFetch(config.proxy)` unconditionally in 12603e4d6 for consistency with the main API client.

packages/js-sdk/src/volume/client.ts
```diff
@@ -79,6 +88,7 @@ export class VolumeConnectionConfig {
     this.logger = opts?.logger
     this.requestTimeoutMs = opts?.requestTimeoutMs
     this.signal = opts?.signal
+    this.proxy = opts?.proxy
```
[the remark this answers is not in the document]
mishushakov: this is a separate issue, will be handled in a separate PR

matthewlouisbrockman (changes requested):
Not sure the env auto detection is a patch change here; it's possibly breaking and there's no off switch.

I've seen it before where when you have _PROXY env supported you also have a no_proxy bypass which we probably want to at least be able to opt out, although if want to keep a patch, we should probably make the env reading opt in rather than opt out? IMO lighter touch of opt in probably better but not too opinionated.

Tangentially, reading HTTP_PROXY as cursor pointed out above is risky; some servers have vulns there

mishushakov:
@matthewlouisbrockman I've dropped the env var for now, also fixed the Volume instance proxy inheritance.

matthewlouisbrockman (approved):

PR #1387, opened 2026-06-05
fix(sdk): consistent rate limit error for envd 429 responses

mishushakov:
## Summary

The main API client already raised `RateLimitError` (JS) / `RateLimitException` (Python) for HTTP 429, but the lower-level **envd** HTTP and RPC layers fell through to a generic sandbox error, so the same rate-limit condition surfaced as a different type depending on which request path hit it. This maps envd 429 (and the equivalent gRPC `ResourceExhausted` code) to the dedicated rate-limit error consistently across the JS SDK and the Python sync/async SDKs. The JS RPC layer was also missing the `ResourceExhausted` mapping entirely, which is now added for parity with Python.

Also includes a small Python-SDK cleanup: `ConnectionConfig` set `request_timeout` via `_get_request_timeout` and then immediately overwrote it with an inline copy of the same logic — the duplicate block is removed (no behavior change).

## Changes
- `js-sdk/src/envd/api.ts`, `python-sdk/e2b/envd/api.py` — envd HTTP 429 → `RateLimitError`/`RateLimitException`
- `js-sdk/src/envd/rpc.ts` — added gRPC `Code.ResourceExhausted` → `RateLimitError`
- `python-sdk/e2b/connection_config.py` — removed redundant `request_timeout` reassignment
- Added unit tests for both the envd HTTP and RPC error mappers in JS and Python
- Changeset (`e2b`: patch)

## Usage example
```ts
import { Sandbox, RateLimitError } from 'e2b'

try {
  await sandbox.files.write('/tmp/file.txt', 'data')
} catch (err) {
  if (err instanceof RateLimitError) {
    // now reliably caught regardless of which request path was rate limited
  }
}
```

```python
from e2b import RateLimitException

try:
    sandbox.files.write("/tmp/file.txt", "data")
except RateLimitException:
    ...
```

### DIFF
--- a/.changeset/rate-limit-error-consistency.md
+++ b/.changeset/rate-limit-error-consistency.md
@@ -1,0 +1,5 @@
+---
+"e2b": patch
+---
+
+Return a dedicated rate limit error for HTTP 429 responses from the envd API. Previously these were surfaced as a generic sandbox error, unlike the main API client which already raised `RateLimitError` (JS) / `RateLimitException` (Python). Rate limit errors are now consistent across all SDK request paths.
--- a/.changeset/rate-limit-error-consistency.md
+++ b/.changeset/rate-limit-error-consistency.md
@@ -1,5 +1,6 @@
 ---
 "e2b": patch
+"@e2b/python-sdk": patch
 ---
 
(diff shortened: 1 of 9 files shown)


matthewlouisbrockman (approved):
lgtm... i think there's an edge case in js-sdk/envd/api.ts where we're checking against !res.error instead of res.error === undefined but can do that in future PR cause not related to the 429 fix

PR #1390, opened 2026-06-05
feat(cli): fully deprecate template build command (v1)

mishushakov:
## Summary

- Strips all v1 build logic from \`e2b template build\` (\`bd\`): Docker build/push, API calls, config-loading, and retry/proxy handling are removed
- The command now only displays the existing yellow deprecation warning (pointing to the v2 migration guide) and exits with code 1
- Deletes \`buildWithProxy.ts\` which is no longer referenced anywhere
- Moves \`getDockerfile\` helper (used by \`template create\` and \`template migrate\`) to a new shared \`dockerfile.ts\` module, leaving \`build.ts\` as a clean stub

## Test plan

- [ ] Run \`e2b template build\` — confirm deprecation warning is shown and the command exits immediately
- [ ] Run \`e2b template create\` and \`e2b template migrate\` — confirm they still work (both use the moved \`getDockerfile\` helper)


matthewlouisbrockman (approved):
lgtm! v1 was deprecated even before I joined so rming code makes sense

Issue #1034, opened 2025-11-25
Executing AsyncCommandHandle.kill() will cause blocking and cannot kill the process.


fsqinghuayu:
**Describe the bug**
When I also use asyncio.create_task(run_command()) to start a task and get output in real-time, upon catching an exception (such as request disconnection), I want to abort the task, but kill() does not take effect. After pause sandbox and get_info() shows the sandbox status as paused, but the process is still running（It's also possible that I misjudged the situation. The Claude model token costs were still being incurred after the sandbox was paused — this might be because the request had already been sent out when I paused the sandbox.）.

**Expected behavior**
1. Be able to successfully kill the process
2. When the sandbox is paused, the process should stop executing
3. I need an effective method to kill processes. Now when I reconnect to the sandbox, the previous task starts running again.This is causing me to incur a lot of unnecessary token expenses.

**Main code**
```
 async def run_command():
                nonlocal command_handle
                try:
                    logger.info(f"[COMMAND] Starting command with background=True...")
                    command_handle = await sdx.commands.run(
                        command,
                        timeout=600,
                        on_stdout=on_stdout,
                        on_stderr=on_stderr,
                        cwd=work_path,
                        envs=envs,
                        user=user,
                        background=True,
                    )
                    logger.info(
                        f"[COMMAND] Got command handle: {command_handle}, pid: {getattr(command_handle, 'pid', 'N/A')}")
                    logger.info(f"[COMMAND] Waiting for command to complete...")
                    result = await command_handle.wait()
                    logger.info(f"[COMMAND] Command completed with exit_code: {getattr(result, 'exit_code', 'N/A')}")
                except asyncio.CancelledError:
                    logger.warning(f"[COMMAND] run_command task was cancelled")
                    raise
                except Exception as e:
                    logger.error(f"[COMMAND] Command error: {e}")
                finally:
                    logger.info(f"[COMMAND] run_command finally block, setting done_event")
                    done_event.set()

        try:

            run_command_task = asyncio.create_task(run_command())
            logger.info(f"[STREAM] Started run_command_task: {run_command_task}")

            while not done_event.is_set() or not output_queue.empty():
                if await request.is_disconnected():
                    logger.warning("[STREAM] Client disconnected detected!")
                    logger.info(f"[STREAM] command_handle exists: {command_handle is not None}")
                    logger.info(
                        f"[STREAM] run_command_task done: {run_command_task.done() if run_command_task else 'N/A'}")

                    if command_handle:
[...]

PR #1392, opened 2026-06-05
fix(python-sdk): stop command event iterator at terminal end event

mishushakov:
## Problem

In the Python SDK, a background command handle reads the command's output stream in a loop. When the command finishes, envd sends a terminal `end` event (carrying the exit code) and then closes the HTTP stream. The iterator recorded the result on the `end` event but kept awaiting the next stream event, so it only stopped once envd closed the stream, and the underlying `acall_server_stream`/`call_server_stream` generator held its `async with http_resp` block (and the connection) open until garbage collection.

## Changes (Python only)

- Stop the iterator as soon as the terminal `end` event is received.
- Release the HTTP stream/connection deterministically in all paths (normal completion, exception, `disconnect()`) via `aclose()`/`close()` in a `finally`. In the async SDK, `disconnect()` cancels the task and lets that `finally` close the stream, avoiding the concurrent-access `RuntimeError` that previously disabled `aclose()` there.
- Added async + sync unit tests for the stop-at-end behavior.

## Scope note

This originally also touched the JS SDK, but the JS command handle already drains its event stream to completion (`done`), so it has no leak and no practical `wait()` hang — the JS changes were unnecessary and caused CI failures (a Deno timer leak; then connection-pool exhaustion under concurrent load due to connect-es's abort-before-read behavior). The JS SDK has been reverted to match `main`; this PR is Python-only.

`kill()` semantics are unchanged; the child-process-survival facet of #1034 is handled separately.


matthewlouisbrockman (approved):

mishushakov:
Dropping as this fix might be actually unnecessary and introduce more foot guns.

PR #1403, opened 2026-06-08
fix(python-sdk): align sync and async implementations

mishushakov:
Reconciles divergences found while auditing the sync and async Python SDK trees, keeping behavior equivalent across both.

- **Parameter ordering:** aligned `_create` and `Commands._start` signatures to the public API and to each other, and reordered the `Commands.connect` rpc args (`headers` before `timeout`) to match the `_start` convention.
- **`pause` return:** the public `pause()` / `beta_pause()` are now annotated `-> str` and actually return the sandbox ID (matching `_cls_pause` and the class-method form, which already returned it) instead of `-> None`; the `:return:` docstrings are restored.
- **Exceptions:** the internal "Body of the request is None" guard in `sandbox_api` now consistently raises a bare `Exception` (matching the volume client) instead of mixing `Exception`/`SandboxException` between sync and async.
- **Misc:** async `Filesystem.write` now passes keyword args; the async constructor reuses the cached `envd_api_url` property instead of recomputing the sandbox URL; async pty `resize` gains a `-> None` annotation.
- **Docstrings:** aligned the deprecation marker and `get_metrics`/`write_files`/`kill` wording across sync/async, and fixed a `**seconds**s` typo.

These are alignment/consistency fixes only; the deeper architectural async-vs-sync splits (streaming-vs-polling `watch_dir`, pty `on_data`, command output callbacks) are intentional and left untouched.

### DIFF
--- a/.changeset/python-sync-async-alignment.md
+++ b/.changeset/python-sync-async-alignment.md
@@ -1,0 +1,5 @@
+---
+"@e2b/python-sdk": patch
+---
+
+Align the sync and async Python SDK implementations: consistent parameter ordering (`_create`, `Commands._start`), matching docstrings, keyword arguments in `Filesystem.write`, and a consistent bare `Exception` for the internal "Body of the request is None" guard (matching the volume client).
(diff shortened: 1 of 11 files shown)


huv1k (approved):

.changeset/python-sync-async-alignment.md
```diff
@@ -0,0 +1,7 @@
+---
+"@e2b/python-sdk": patch
```
huv1k: I think this shouldn't be patch

PR #1404, opened 2026-06-09
Add sandbox lifecycle create option

matthewlouisbrockman:
adds the lifecycle for autoresume/autopause and timeout to the cli

### DIFF
--- a/packages/cli/src/commands/sandbox/create.ts
+++ b/packages/cli/src/commands/sandbox/create.ts
@@ -33,13 +33,13 @@ export function createCommand(
     .addOption(configOption)
     .option('-d, --detach', 'create sandbox without connecting terminal to it')
     .option(
-      '--ontimeout <action>',
+      '--lifecycle.ontimeout <action>',
       'action when sandbox timeout is reached: pause or kill',
       parseOnTimeout
     )
     .option(
-      '--autoresume',
-      'enable sandbox auto-resume, requires --ontimeout pause'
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/sandbox/create.ts
```diff
@@ -25,6 +30,12 @@ export function createCommand(
     .addOption(pathOption)
     .addOption(configOption)
     .option('-d, --detach', 'create sandbox without connecting terminal to it')
+    .option(
+      '--lifecycle <json>',
+      'sandbox lifecycle JSON, eg. {"onTimeout":"pause","autoResume":true}',
+      parseLifecycle
+    )
```
mishushakov, line 43: I would flatten this to --ontimeout and --autoresume

mishushakov (changes requested):
would change DX here by flattening the params, that should also help with parsing.
if you want them under lifecycle anyways, you could also:

```
e2b sbx create base --lifecycle.autoresume --lifecycle.ontimeout="pause"
```

packages/cli/src/commands/sandbox/create.ts
matthewlouisbrockman: kk, fixed

mishushakov:
have changed it to prefix with --lifecycle. to avoid collisions once we have network also (--network.)

mishushakov (approved):

PR #1407, opened 2026-06-09
Add Python SDK option to opt out of HTTP/2

sarmientoF:
## Summary
- Keep HTTP/2 as the Python SDK default.
- Add a public `http2` connection option so callers can opt out with `http2=False` for API and envd transports.
- Thread the option through sync/async API clients and sandbox envd transport creation.
- Add tests for the default HTTP/2 behavior and the explicit HTTP/1.1 opt-out.

## Why
For normal use, HTTP/2 remains the right default. The opt-out is for server-side high fan-out workloads where `httpx` uses one HTTP/2 connection per origin and therefore one server `MAX_CONCURRENT_STREAMS` budget. Once a workload has more long-lived concurrent envd RPCs than the peer stream cap, HTTP/2 requests queue behind that single connection instead of spreading across the existing connection pool.

`http2=False` gives those workloads a documented alternative: HTTP/1.1 uses the SDK/httpx connection pool and avoids coupling all concurrent envd RPCs to one HTTP/2 stream budget. This is useful for bash/filesystem/commands/pty/control-plane fan-out and does not change behavior for default users.

## Reproduction / metrics
Using this branch and `AsyncSandbox.commands.run`, I ran 180 concurrent commands against one sandbox. Benchmark script: https://gist.github.com/sarmientoF/0d7db8c8f9cdaa154ab915088a238967

```sh
python /tmp/e2b_cloud_http2_repro.py \
  --env-file /Users/fernando-sarmiento/Desktop/jai/E2B/.env \
  --concurrency 180 \
  --sleep-seconds 4 \
  --request-timeout 45
```

Workload: `sleep 4; echo ok`, 180 concurrent server-streaming envd RPCs.

| mode | ok/errors | elapsed | throughput | p50 | p95 | p99 | max |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| HTTP/2 default | 180/0 | 9.03s | 19.94/s | 4.82s | 9.01s | 9.01s | 9.01s |
| HTTP/1.1 opt-out | 180/0 | 5.20s | 34.63/s | 5.01s | 5.16s | 5.17s | 5.18s |

The HTTP/2 run completes in two waves because concurrency above the stream budget cannot use a second HTTP/2 connection. The HTTP/1.1 opt-out completes in one wave because requests are distributed through the connection pool.

## Tests
- `./.venv/bin/pytest tests/test_api_client_transport.py tests/sync/sandbox_sync/test_config_propagation.py tests/async/sandbox_async/test_config_propagation.py`
- `./.venv/bin/ruff check .`
- `./.venv/bin/ruff format --check .`
- `./.venv/bin/ty check`

### DIFF
--- a/.changeset/python-http1-transports.md
+++ b/.changeset/python-http1-transports.md
@@ -1,0 +1,5 @@
+---
+"@e2b/python-sdk": patch
+---
+
+Add a Python SDK `http2` connection option so high fan-out workloads can opt out of HTTP/2.
--- a/packages/python-sdk/e2b/api/client_async/__init__.py
+++ b/packages/python-sdk/e2b/api/client_async/__init__.py
@@ -13,7 +13,7 @@ logger = logging.getLogger(__name__)
 def get_api_client(config: ConnectionConfig, **kwargs) -> AsyncApiClient:
     return AsyncApiClient(
         config,
-        transport=get_transport(config),
+        transport=get_transport(config, http2=config.http2),
         **kwargs,
     )
 
--- a/packages/python-sdk/e2b/api/client_sync/__init__.py
+++ b/packages/python-sdk/e2b/api/client_sync/__init__.py
@@ -13,7 +13,7 @@ logger = logging.getLogger(__name__)
 def get_api_client(config: ConnectionConfig, **kwargs) -> ApiClient:
     return ApiClient(
         config,
-        transport=get_transport(config),
+        transport=get_transport(config, http2=config.http2),
         **kwargs,
     )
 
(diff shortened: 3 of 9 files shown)


mishushakov:
I will defer to @matthewlouisbrockman, but I don't think we should allow this - it would re-introduce the connection issues we've been trying to avoid by introducing http2.

mishushakov:
We've decided that we will handle this internally - thanks for bringing this into our attention, though!

sarmientoF:
Thank you 🙌

PR #1409, opened 2026-06-09
feat(python-sdk): add logger option for request/debug logging

mishushakov:
Adds a `logger` option (a standard library `logging.Logger`) to `Sandbox.create`/`AsyncSandbox.create` and the static `Sandbox.connect(sandbox_id, ...)`, wired into the API client, the envd client, the volume content client, and the RPC (ConnectRPC) path. The logger is stored on the sandbox and propagates to all of its later operations — including control-plane calls like `kill`/`pause`/`set_timeout`/`get_info` (via `get_api_params`) — so logging keeps working after construction; mirroring the JS SDK, `logger` is a construction-time option and not a public per-request parameter those methods accept from the caller, and nothing is logged unless a logger is supplied. The stdlib `logging.Logger` is used directly as the adapter (no ported JS `Logger` interface), and log levels match JS: requests at `INFO`, successful API and unary RPC responses at `INFO`, streamed RPC messages at `DEBUG`, failed API responses (status >= 400) at `ERROR`. The always-on module-level (`e2b.*`) request logging at the transport layer was removed in favor of this opt-in client-layer logging, and volume content operations continue to accept `logger` per call via `VolumeApiParams` to match the JS Volume API. Includes a changeset and unit tests in `tests/test_logging_option.py`.

## Usage

```python
import logging
from e2b import Sandbox

logging.basicConfig(level=logging.DEBUG)
logger = logging.getLogger("my-app.e2b")

sbx = Sandbox.create(logger=logger)
sbx.commands.run("echo hello")   # RPC logged via `logger`
sbx.set_timeout(60)              # control-plane call also logged via `logger`
```


mishushakov:
@matthewlouisbrockman was the two commits added on purpose? if it's not related to the PR, i'd move that

matthewlouisbrockman:
apparently that's what happens when u click the copilot link next to review at top of github ui, lemme figure out how to undo those lol

packages/python-sdk/e2b/api/__init__.py
```diff
@@ -22,7 +22,45 @@
...
+    to the given ``logging.Logger``. Requests log at ``INFO``, successful
+    responses at ``INFO`` and responses with status >= 400 at ``ERROR``.
+
+    Returns no hooks when ``log`` is ``None`` so that nothing is logged unless a
+    logger was explicitly supplied."""
+    if log is None:
+        return {}
+
+    def on_request(request) -> None:
+        log.info(f"Request {request.method} {request.url}")
```
matthewlouisbrockman, line 37: should we sanitize these? used to be
```
    def handle_request(self, request):
        url = f"{request.url.scheme}://{request.url.host}{request.url.path}"
        logger.info(f"Request: {request.method} {url}")
        response = super().handle_request(request)

        # data = connect.GzipCompressor.decompress(response.read()).decode()
        logger.info(f"Response: {response.status_code} {url}")

        return response

```

now it's the raw url

matthewlouisbrockman (approved):

packages/python-sdk/e2b/api/__init__.py
mishushakov: I will merge this one and then implement the sanitization in both SDKs, as JS also appears to be lacking.

PR #1411, opened 2026-06-09
fix: align behavior and API surface between the JS and Python SDKs

mishushakov:
Aligns several behavioral and API-surface discrepancies between the JS and Python SDKs found during a cross-SDK audit. **Python:** `commands.send_stdin`/`CommandHandle.send_stdin` now accept `bytes` (plus `request_timeout` on the handle), `git.reset` gets a typed `GitResetMode` with JS-matching validation, `sandbox_url` is threaded through `get_api_params` (and the dead `SandboxOpts` key removed), and `from_image` requires both `username` and `password` when credentials are given. **JS:** `getFullInfo` was removed in favor of a single `getInfo` that now includes `sandboxDomain` (matching Python's `get_info`), `fromImage` requires both credentials, `getBuildStatus` defaults `logsOffset` to `0`, `getMetrics`/`kill` short-circuit consistently in debug mode (instance + static), and `requestTimeoutMs: 0` explicitly disables the request timeout. Tests were added on both sides (git-arg validation, stdin bytes, credential validation, timeout-0, connection config) and the CLI's `sandbox info` now uses `getInfo`. See the changeset for the full per-SDK list.

## Usage examples

```ts
// JS: registry credentials now require both fields
Template().fromImage('registry.example.com/img:latest', { username: 'u', password: 'p' })

// JS: getInfo now exposes sandboxDomain (getFullInfo removed)
const info = await Sandbox.getInfo(sandboxId)
console.log(info.sandboxDomain)

// JS: disable the request timeout
await Sandbox.create({ requestTimeoutMs: 0 })
```

```python
# Python: send raw bytes to stdin
sandbox.commands.send_stdin(cmd.pid, b"hello")

# Python: typed git reset mode (validated)
sandbox.git.reset(repo, mode="hard")

# Python: registry credentials require both fields
Template().from_image("registry.example.com/img:latest", username="u", password="p")
```

### DIFF
--- a/packages/js-sdk/src/sandbox/sandboxApi.ts
+++ b/packages/js-sdk/src/sandbox/sandboxApi.ts
@@ -805,6 +805,10 @@ export class SandboxApi {
       signal: config.getSignal(opts?.requestTimeoutMs, opts?.signal),
     })
 
+    if (res.error?.code === 404) {
+      throw new SandboxNotFoundError(`Sandbox ${sandboxId} not found`)
+    }
+
     const err = handleApiError(res)
     if (err) {
       throw err
--- a/packages/js-sdk/src/sandbox/sandboxApi.ts
+++ b/packages/js-sdk/src/sandbox/sandboxApi.ts
@@ -690,10 +690,61 @@ export class SandboxApi {
     sandboxId: string,
     opts?: SandboxApiOpts
   ): Promise<SandboxInfo> {
-    const fullInfo = await this.getFullInfo(sandboxId, opts)
-    delete fullInfo.envdAccessToken
+    const config = new ConnectionConfig(opts)
+    const client = new ApiClient(config)
 
-    return fullInfo
+    const res = await client.api.GET('/sandboxes/{sandboxID}', {
+      params: {
+        path: {
+          sandboxID: sandboxId,
+        },
+      },
+      signal: config.getSignal(opts?.requestTimeoutMs, opts?.signal),
+    })
+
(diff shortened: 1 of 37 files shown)


packages/js-sdk/src/template/index.ts
```diff
@@ -452,6 +452,13 @@ export class TemplateBase
 
     // Set the registry config if provided
     if (credentials) {
+      if (!credentials.username || !credentials.password) {
```
matthewlouisbrockman: probably should do credential check before doing the work above

packages/python-sdk/e2b/template/main.py
```diff
@@ -1008,7 +1008,14 @@ def from_image(
         self._base_template = None
 
         # Set the registry config if provided
-        if username and password:
+        if username or password:
```
matthewlouisbrockman: slight divergence with the js here; js checks `if (credentials) {` this checks `if username or password:` so if username and password are present but falsy, JS will reject but Python will allow

packages/js-sdk/src/sandbox/sandboxApi.ts
```diff
@@ -822,65 +889,6 @@ export class SandboxApi {
     }
   }
 
-  static async getFullInfo(sandboxId: string, opts?: SandboxApiOpts) {
```
matthewlouisbrockman, line 825: maybe a reason to make major instead of minor but not end of world

matthewlouisbrockman (approved):

PR #1413, opened 2026-06-09
refactor: replace rich with stdlib ANSI for template logger

zsxh1990:
## Summary

Replace the `rich` library dependency with stdlib ANSI escape codes in the template build logger (`packages/python-sdk/e2b/template/logger.py`).

This is the first step in a footprint reduction effort — `rich` is currently used in only 1 file, making it a clean target for elimination.

### `packages/python-sdk/e2b/template/logger.py`

| Before | After |
|--------|-------|
| `rich.console.Console` | `print()` + ANSI escape codes |
| `rich.text.Text` | String concatenation |
| `rich.style.Style` | ANSI color constants |

### Visual output

Identical build progress display:
```
⣾ Building 12.3s | 14:30:25 INFO  Starting build...
```

## Technical Rationale

- **Deployment image size**: `rich` pulls in `pygments`, `markdown-it-py`, `mdurl` — eliminating it reduces the dependency tree
- **Security surface**: Fewer transitive dependencies in isolated runtime environments
- **Cold-start latency**: No need to import rich at module load time
- **Stdlib sufficiency**: ANSI escape codes provide identical functionality for this use case

## Testing

```bash
cd packages/python-sdk
python -m pytest tests/ -v
```

The visual output is identical — ANSI codes produce the same colored, animated build progress display.

Signed-off-by: zsxh1990 <zsxh1990@users.noreply.github.com>

### DIFF
--- a/packages/python-sdk/e2b/template/logger.py
+++ b/packages/python-sdk/e2b/template/logger.py
@@ -5,10 +5,6 @@ from dataclasses import dataclass, field
 from datetime import datetime
 from typing import Optional, TypedDict, Callable, Dict, Literal
 
-from rich.console import Console
-from rich.style import Style
-from rich.text import Text
-
 from e2b.template.utils import strip_ansi_escape_codes
 
 """Log entry severity levels."""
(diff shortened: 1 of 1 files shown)


mishushakov:
Thanks for opening the PR, but we're not seeing any major user pain this PR would solve, since is mostly cosmetic.
Closing.

PR #1421, opened 2026-06-11
fix(sdks): transport proxy caching, request timeout, and option handling

mishushakov:
Fixes a batch of connection-handling bugs found in review, with unit tests for each and a changeset for both SDKs.

- **Python**: cached HTTP transports are now keyed on the configured proxy, so clients with different (or no) proxy settings no longer silently reuse the transport built for the first proxy seen.
- **Python**: `request_timeout` now applies to control-plane API requests — the underlying httpx client was previously built with no timeout at all (`request_timeout=0` still disables it).
- **Python**: the server-stream parser no longer stalls or drops the final envelope when the remaining payload is shorter than the 5-byte envelope header; also removed the no-op `@_retry` decorators from the streaming RPC methods to avoid confusion.
- **JS + Python**: an explicit `debug=False` / `debug: false` now overrides `E2B_DEBUG=true` instead of being ignored.
- **JS**: the RPC logger no longer crashes requests with a `TypeError` when a response contains protobuf int64 (`bigint`) fields (e.g. `EntryInfo.size` returned by `files.list()`/`stat()` and `includeEntry` watch events); they are logged as strings.

## Usage examples

```ts
// JS: logging RPCs whose responses carry int64 fields no longer throws
const sbx = await Sandbox.create({ logger: console })
await sbx.files.list('/')

// JS: force-disable debug mode even when E2B_DEBUG=true is set
const sbx2 = await Sandbox.create({ debug: false })
```

```python
# Python: per-call API timeout is now actually enforced
Sandbox.list(request_timeout=10)

# Python: clients with different proxies get their own transports
Sandbox.create(proxy="http://127.0.0.1:8080")
Sandbox.create()  # no longer routed through the proxy above
```


matthewlouisbrockman (approved):

packages/js-sdk/src/connectionConfig.ts
```diff
@@ -206,7 +206,7 @@ export class ConnectionConfig {
 
   constructor(opts?: ConnectionOpts) {
     this.apiKey = opts?.apiKey || ConnectionConfig.apiKey
-    this.debug = opts?.debug || ConnectionConfig.debug
+    this.debug = opts?.debug ?? ConnectionConfig.debug
     this.domain = opts?.domain || ConnectionConfig.domain
     this.accessToken = opts?.accessToken || ConnectionConfig.accessToken
```
djeebus, line 211: should we be using `??` on the rest of these?

mishushakov: this one is specifically because we want to test truthiness

PR #1423, opened 2026-06-11
Deduplicate sync thread-local client setup

matthewlouisbrockman:
## Summary
- remove eager sync envd/RPC client construction from sandbox helpers
- make sync sandbox filesystem, commands, pty, and API clients initialize from the calling thread's transport
- deduplicate `Sandbox` and `Filesystem` envd HTTP client caching so each thread gets one envd client
- cover `Template.build` file uploads by making generated API clients thread-local too

### DIFF
--- a/packages/python-sdk/e2b/api/__init__.py
+++ b/packages/python-sdk/e2b/api/__init__.py
@@ -1,11 +1,14 @@
+import asyncio
 import json
 import logging
 import os
 import re
+import threading
 from dataclasses import dataclass
 from types import TracebackType
-from typing import Optional, Protocol, Union
+from typing import Callable, Dict, Optional, Protocol, Union
 
+import httpx
 from httpx import AsyncBaseTransport, BaseTransport, Limits, Timeout
 
 from e2b.api.client.client import AuthenticatedClient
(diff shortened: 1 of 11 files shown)


packages/python-sdk/e2b/api/__init__.py
```diff
@@ -1,11 +1,14 @@
...
 import os
 import re
+import threading
 from dataclasses import dataclass
 from types import TracebackType
-from typing import Optional, Protocol, Union
+from typing import Callable, Dict, Optional, Protocol, Union
 
+import httpx
 from httpx import AsyncBaseTransport, BaseTransport, Limits, Timeout
```
wj-e2b, line 12: Minor nit, I hate dual imports

packages/python-sdk/e2b/api/__init__.py
```diff
@@ -177,6 +197,53 @@ def __init__(
...
+                transport=self._transport_factory(),
+            )
+            self._thread_local.client = client
+        return client
+
+    def get_async_httpx_client(self) -> httpx.AsyncClient:
+        if self._async_client is not None or self._async_transport_factory is None:
+            return super().get_async_httpx_client()
+
+        loop_id = id(asyncio.get_running_loop())
```
wj-e2b, line 231: ID is not guaranteed unique across the same process since ID's are bound to memory addresses. This is _probably_ fine, but depends on cpython's memory allocator.

wj-e2b (approved):

PR #1430, opened 2026-06-12
fix(js-sdk): raise errors for non-2xx API and envd responses with empty bodies

mishushakov:
## Description

`handleApiError` and `handleEnvdApiError` used the presence of an error body as the failure signal, but openapi-fetch returns `error: undefined` for non-2xx responses with `Content-Length: 0` (and `error: ''` for empty bodies without that header), so such failures were silently treated as success — e.g. `sandbox.isRunning()` returned `true` for a failing `/health` endpoint. Both handlers now gate on `response.ok` and fall back to the status text when no error message is available. Volumes go through the same shared handler, so they inherit the fix. Regression tests cover both handlers and the volume path, including a 500-with-empty-body case that fails against the old code.

## Example

```ts
// Before: a 500 response with Content-Length: 0 was treated as success
await Volume.create('my-volume') // resolved with 'Response data is missing'

// After: the failure surfaces as the proper SDK error
await Volume.create('my-volume') // throws VolumeError('500: Internal Server Error')
```

### DIFF
--- a/.changeset/shiny-pugs-check.md
+++ b/.changeset/shiny-pugs-check.md
@@ -1,0 +1,5 @@
+---
+"e2b": patch
+---
+
+Raise an error for non-2xx API and envd responses with empty bodies (e.g. `Content-Length: 0`) instead of treating them as successful.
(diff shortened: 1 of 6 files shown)


matthewlouisbrockman (approved):
valid fix, when u say 204 do you mean 404 tho? maybe can reword the explanation/comment for clarity before merging

PR #1431, opened 2026-06-12
Fix volume SDK issues: transports, timeouts, empty files, eager stream errors

mishushakov:
Fixes a batch of review findings in the JS and Python volume SDKs, aligning behavior between the two. Python now caches `AsyncVolume` HTTP transports per event loop and proxy (sync per thread) instead of a process-wide singleton, applies the 60s default `request_timeout` to metadata operations that previously ran with httpx timeouts disabled, no longer falls back to `E2B_ACCESS_TOKEN` for volume content auth, and no longer mutates the caller's `headers` dict. JS `Volume.readFile` now returns empty values instead of `undefined` for empty files, and volume content requests get the documented 60s default request timeout. All changes are covered by new mock-based unit tests (no live API needed) plus changesets for both SDKs.

## Usage examples

```python
volume = await AsyncVolume.connect(volume_id)

# Times out after 60s by default (previously could hang indefinitely)
entries = await volume.list("/")
```

```ts
const volume = await Volume.connect(volumeId)

// Returns an empty Blob / ReadableStream for empty files (previously undefined)
const blob = await volume.readFile('empty.txt', { format: 'blob' })

// Times out after the documented 60s by default; pass 0 to disable
await volume.getInfo('file.txt', { requestTimeoutMs: 0 })
```

### DIFF
--- a/packages/python-sdk/e2b/volume/client_async/__init__.py
+++ b/packages/python-sdk/e2b/volume/client_async/__init__.py
@@ -1,9 +1,11 @@
+import asyncio
 import logging
 import os
-from typing import Optional
+from typing import Dict, Optional, Tuple
 
 import httpx
 from httpx import Limits
+from httpx._types import ProxyTypes
 
 from e2b.api.metadata import default_headers
 from e2b.exceptions import AuthenticationException
@@ -18,12 +20,15 @@ limits = Limits(
     keepalive_expiry=int(os.getenv("E2B_KEEPALIVE_EXPIRY", "300")),
 )
 
+TransportKey = Tuple[int, Optional[ProxyTypes]]
+
 
 def get_api_client(config: VolumeConnectionConfig, **kwargs) -> AsyncVolumeApiClient:
     if config.access_token is None:
         raise AuthenticationException(
-            "Access token is required for volume operations. "
-            "Set `E2B_ACCESS_TOKEN` or pass `token` in options.",
+            "Volume token is required for volume content operations. "
+            "Use `AsyncVolume.create`/`AsyncVolume.connect` to obtain it "
+            "or pass `token` in options.",
         )
 
     headers = {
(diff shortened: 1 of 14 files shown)


packages/python-sdk/e2b/volume/client_async/__init__.py
```diff
@@ -58,12 +68,14 @@ def pool(self):
 
 
 def get_transport(config: VolumeConnectionConfig) -> AsyncTransportWithLogger:
-    if AsyncTransportWithLogger.singleton is not None:
-        return AsyncTransportWithLogger.singleton
+    key: TransportKey = (id(asyncio.get_running_loop()), config.proxy)
```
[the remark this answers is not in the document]
mishushakov: it's same defect as the rest of the SDK, would merge as-is and then do a follow up PR that fixes that in both. @matthewlouisbrockman for visibility

packages/python-sdk/e2b/volume/volume_async.py
```diff
@@ -473,28 +473,38 @@ async def read_file(
...
         if format == "stream":
+            httpx_client = api_client.get_async_httpx_client()
+            request = httpx_client.build_request(
+                method="GET",
+                url=f"/volumecontent/{self._volume_id}/file",
+                params=params,
+                timeout=timeout,
+            )
+            # Send the request eagerly so errors are raised here, not on first iteration
+            response = await httpx_client.send(request, stream=True)
```
matthewlouisbrockman: potential footgun by eagerly requesting if we don't consume them all right away - before could iterate thru file reads without opening connections right away, now need to be a bit more careful because a connection is made immediately

matthewlouisbrockman (approved):

packages/python-sdk/e2b/volume/volume_async.py
mishushakov: I'd drop it from this PR and move streaming changes here - it appears to be bit more involved:
https://github.com/e2b-dev/E2B/pull/1433

PR #1435, opened 2026-06-12
Stream template build-context upload from disk instead of buffering in memory

mishushakov:
## Summary

Template builds previously buffered the entire gzipped build-context tar archive in memory before uploading it. This PR spools the archive to a temporary file and streams it from disk during upload — in the JS SDK and both sync and async Python SDKs — so memory usage no longer scales with the size of the build context.

The upload keeps an explicit `Content-Length` header (taken from the spooled file's size), which S3 presigned PUT URLs require — they reject `Transfer-Encoding: chunked` with `501 NotImplemented` (#1243).

## Changes

- **JS** (`packages/js-sdk/src/template/`): `tarFileStream`/`tarFileStreamUpload` are replaced by `tarFileToStream`, which writes the archive to a temp file and returns a self-cleaning read stream plus its `size`. The spooled temp file deletes itself once the stream is closed (consumed, errored, or destroyed) via the stream's `close` event — mirroring the Python SDK's `tar_file_stream`. `buildApi` streams this body with `duplex: 'half'` and an explicit `Content-Length` from `size`; if `fetch` throws before consuming the body, it destroys the stream to trigger the same cleanup. There is no separate cleanup callback, so a cleanup failure can no longer mask the upload result.
- **Python** (`packages/python-sdk/e2b/template/utils.py`, `template_async/build_api.py`, `template_sync/build_api.py`): `tar_file_stream` now writes to a `tempfile.TemporaryFile` instead of `io.BytesIO` and returns the file object positioned at the start; the upload streams from it with an explicit `Content-Length` and closes it (deleting the temp file) when done.
- Tests updated for the new return shapes (JS `tarFileToStream.test.ts`, `uploadFile.test.ts`; Python upload/tar tests), including assertions that the spooled archive is removed on both the consume and destroy paths.

## Usage

No API changes — `Template.build()` / template builds behave the same, just without holding the build context in memory:

```ts
await Template.build(template, { alias: 'my-template' })
```

```python
Template.build(template, alias="my-template")
```

Split out of #1433, which covers streaming for sandbox/volume file uploads and downloads.


matthewlouisbrockman (approved):
- we can build via cli so might want to include that
- maybe test for showing we aren't buffering would be nice
neither blocking

mishushakov:
@matthewlouisbrockman CLI is just calling into Template.build method

mishushakov:
I have not thought about not buffering tests, perhaps it would require to monitor that the heap is not increasing, idk.

PR #1440, opened 2026-06-14
feat: add skipApiKeyValidation flag to bypass API key format check

### LINKED ISSUE #1398: SDKs should allow bypassing API key format validation for custom/self-hosted endpoints
## Summary

After [#1356](https://github.com/e2b-dev/E2B/pull/1356), both the JS and Python SDKs validate API keys client-side and only accept keys matching the `e2b_<hex>` format.

This is helpful for E2B Cloud, but it breaks compatibility with custom or self-hosted endpoints that expose an E2B-compatible API while using a different credential format.

## Context

We are using a custom endpoint via `domain` / `apiUrl`.

That backend accepts a different API key format (for example `sk-...`), and this used to work with older SDK versions because the request was sent to the backend directly.

With the newer SDK behavior introduced in [#1356](https://github.com/e2b-dev/E2B/pull/1356), the SDK rejects the key locally before making any HTTP request.

This affects both:
- the JS SDK
- the Python SDK

## Problem

The SDK currently assumes that any backend reached through `domain` / `apiUrl` must also use the official E2B Cloud API key format.

That assumption is valid for E2B Cloud, but not necessarily for:
- self-hosted deployments
- private compatible backends
- custom gateways/proxies that intentionally use a different credential scheme

As a result, valid credentials for custom deployments are rejected before the server can evaluate them.

## Possible fixes

A few implementation options would solve this cleanly:

1. Automatically skip client-side API key format validation when a custom `domain` or `apiUrl` is provided.
   - This keeps the current behavior for E2B Cloud.
   - It avoids hardcoding E2B Cloud credential assumptions for self-hosted or compatible backends.

2. Add an explicit SDK option to skip validation.
   - JS example: `skipApiKeyValidation: true`
   - Python example: `skip_api_key_validation=True`

3. Add an environment-variable-based escape hatch.
   - Example: `E2B_SKIP_API_KEY_VALIDATION=1`

## Preferred direction

My preference would be one of these:

- automatically skip validation for custom endpoints, or
- add an explicit opt-out parameter for self-hosted / compatible deployments

Either approach would preserve the better UX for E2B Cloud users while restoring compatibility for custom deployments.

## Additional note

This request is not about weakening validation for E2B Cloud.

It is specifically about avoiding hardcoded credential-format assumptions when the SDK is pointed at a non-default backend.

## Offer to help

If this direction is acceptable, I’d be happy to help implement it and open a PR.

If you prefer, I can prepare a PR for:
- skipping validation when `domain` / `apiUrl` is customized, or
- adding an explicit opt-out flag in both JS and Python SDKs.

shashank-poola:
## Problem

Since #1356, both SDKs validate API keys against the `e2b_<hex>` pattern client-side before any request is made. This breaks self-hosted or E2B-compatible backends that issue credentials in different formats (e.g. `sk-...`, JWTs, opaque tokens).

Fixes #1398

## Solution

Add a `skipApiKeyValidation` / `skip_api_key_validation` boolean that bypasses the format check when set. The presence check is still enforced — a key is still required, only the `e2b_<hex>` pattern check is skipped.

**JS SDK**

```ts
  apiKey: 'sk-my-custom-token',
  skipApiKeyValidation: true,
})
```

**Python SDK**

```python
sandbox = Sandbox(
    api_key="sk-my-custom-token",
    skip_api_key_validation=True,
)
```

**Environment variable**

```bash
E2B_SKIP_API_KEY_VALIDATION=true
```

## Difference from PR #1360

PR #1360 adds `apiKeyPrefix` which still validates a hex body after the prefix, fails for JWTs and tokens with non-hex characters. This PR skips the entire format check, covering all non-standard credential formats.

## Changes

- New `skipApiKeyValidation` (JS) / `skip_api_key_validation` (Python) boolean, defaults to `false`
- `E2B_SKIP_API_KEY_VALIDATION` env var support
- `SandboxApiOpts` updated so per-call overrides are type-safe
- `get_api_params` propagates the flag through child `ConnectionConfig` objects
- Tests: default value, env var, case-insensitivity, arg priority, propagation through `sandbox.pause()`, `ApiClient` accept/reject behavior

### DIFF
--- a/.changeset/three-pants-tap.md
+++ b/.changeset/three-pants-tap.md
@@ -1,0 +1,6 @@
+---
+'@e2b/python-sdk': patch
+'e2b': patch
+---
+
+Add skipApiKeyValidation / skip_api_key_validation option to bypass API key format check for self-hosted endpoints
--- a/packages/js-sdk/src/api/index.ts
+++ b/packages/js-sdk/src/api/index.ts
@@ -83,7 +83,7 @@ class ApiClient {
       )
     }
 
-    if (config.apiKey) {
+    if (config.apiKey && !config.skipApiKeyValidation) {
       validateApiKey(config.apiKey)
     }
 
--- a/packages/js-sdk/src/connectionConfig.ts
+++ b/packages/js-sdk/src/connectionConfig.ts
@@ -22,6 +22,17 @@ export interface ConnectionOpts {
    */
   apiKey?: string
   /**
+   * When true, skips client-side API key **format** validation.
+   * The key is still required — only the `e2b_<hex>` pattern check is bypassed.
+   *
+   * Use this when connecting to a self-hosted or E2B-compatible endpoint
+   * that issues credentials in a different format (e.g. `sk-...`, JWTs,
+   * opaque tokens). Has no effect on E2B Cloud where the default format applies.
+   *
+   * @default E2B_SKIP_API_KEY_VALIDATION // environment variable or `false`
+   */
+  skipApiKeyValidation?: boolean
+  /**
    * E2B access token to use for authentication.
    *
    * @default E2B_ACCESS_TOKEN // environment variable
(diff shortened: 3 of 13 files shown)


shashank-poola:
@cla-bot

mishushakov:
Thanks for the PR! Will proceed by updating #1360

PR #1442, opened 2026-06-15
fix(sdks): fix template build bugs and consolidate shell quoting

mishushakov:
Fixes seven template-build correctness bugs across the JS and Python (sync + async) SDKs, plus a small shell-quoting cleanup. Each fix has unit/regression coverage, and real end-to-end builds were run against the API in all three SDK variants.

**Template fixes**
- `getAllFilesInPath` now sorts by full path so the files hash no longer depends on filesystem traversal order (the JS `sort()` was a no-op on glob `Path` objects).
- `waitForPort` anchors the port match so port 80 no longer matches 8080.
- The readycmd helpers (`waitForURL`/`waitForFile`/`waitForProcess`) and the file-op helpers (`remove`/`rename`/`makeDir`/`makeSymlink`) now shell-quote interpolated values/paths.
- `waitForBuildFinish` keeps fetching logs after a terminal status so the tail of the build logs (beyond the API's 100-entries-per-call limit) is no longer dropped.
- COPY instructions now collect one stack trace each, so failed-step traces stay aligned after `copy()` with multiple sources or `copyItems()`.
- JS `LogEntry` strips ANSI escape codes in its constructor, matching the Python SDK.

**Cleanup:** consolidated three duplicate single-quote shell helpers (`shellQuote`, the new `quoteShellArg`, and git's `shellEscape`) into one faithful `shlex.quote` port in `utils.ts` — safe values stay unquoted, keeping generated commands and layer-cache hashes stable.

**Behavior note:** templates with paths/URLs containing spaces or shell metacharacters now build correctly; plain paths are unchanged, so existing layer caches are preserved.


matthewlouisbrockman (approved):
works, maybe shoulda split into multiple prs but too late for that cause approvingthis

PR #1444, opened 2026-06-15
fix(python-sdk): drop stream read timeout to fix flaky TimeoutException

mishushakov:
## Summary

`test_run_with_too_short_timeout_iterating` was flaky because the Python SDK set the streaming HTTP `read` timeout to the command `timeout`, so it raced the server's own `deadline_exceeded` response — when the client read timer won, a raw `httpcore.ReadTimeout` leaked out instead of a `TimeoutException`. This PR drops the read timeout on streams and relies on the server-side `connect-timeout-ms` header to enforce the command timeout, matching the JS SDK which has no per-chunk read timeout. As a safety net it also maps transport-level `httpcore` timeouts (e.g. a unary call exceeding `request_timeout`) to `TimeoutException`, and adds a JS parity test asserting `TimeoutError` when iterating a timed-out background command.

## Usage

```python
cmd = sandbox.commands.run("sleep 10", timeout=2, background=True)
try:
    for _ in cmd:
        pass
except TimeoutException:
    print("command timed out")  # now raised reliably, no raw httpcore.ReadTimeout
```


matthewlouisbrockman (approved):
lgtm - not sure if this is too broad with the types of timeouts its consolidating but fixes immediate issues

mishushakov:
yeah, will park it for now - will double check tomorrow

mishushakov:
root cause of the race condition here:
> By setting the client read timeout = command timeout, both fire at the same instant (t = timeout), and they race. That race is the flakiness — not a quirk to be mapped away.

i will instead tie the read timeout to a keep alive interval and keep the command timeout unchanged

PR #1448, opened 2026-06-16
fix(python-sdk): drop stream read timeout, enforce command timeout server-side

mishushakov:
## Summary

Follow-up to #1444 (the `httpcore` → `TimeoutException` mapping, now merged). That mapping made the flaky timeout deterministic, but the underlying cause remained: the streaming HTTP `read` timeout was set to the command `timeout`, so it raced the server's own `deadline_exceeded` response. This PR removes the read timeout on streaming calls entirely and relies on the server-side `connect-timeout-ms` header to enforce the command timeout — matching the JS SDK, which has no per-chunk read timeout. The race is now structurally impossible rather than mapped over.

## Usage

```python
cmd = sandbox.commands.run("sleep 10", timeout=2, background=True)
try:
    for _ in cmd:
        pass
except TimeoutException:
    print("command timed out")  # server-side deadline_exceeded, no transport race
```

## Tradeoff

A silently dropped connection (no RST) on a command with `timeout=0` (disabled) now has no client-side read backstop and relies on keepalive pings — the same posture as the JS SDK.

### DIFF
--- a/.changeset/tidy-streams-relax.md
+++ b/.changeset/tidy-streams-relax.md
@@ -1,0 +1,5 @@
+---
+"@e2b/python-sdk": patch
+---
+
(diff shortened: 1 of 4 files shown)


matthewlouisbrockman (approved):
lgtm. fair tradeoff.

PR #1453, opened 2026-06-17
Stream volume file uploads and downloads instead of buffering in memory

mishushakov:
Follow-up to #1433. Builds on the shared streaming infrastructure introduced there (`FILE_TIMEOUT_MS`, request-controller/stream-cleanup helpers in `connectionConfig`, `io_utils` chunk iterators, the `runtime` guard) and applies the same streaming model to volumes.

> [!NOTE]
> Based on `mishushakov/stream-write-file-upload` (#1433). Merge that PR first; this PR's diff will then retarget to `main` automatically.

## What changed

- **`Volume.writeFile()` / `Volume.write_file()`** — stream the request body instead of buffering it in memory.
  - JS: `ReadableStream` data is streamed outside the browser (half-duplex); browsers still buffer since they can't stream request bodies.
  - Python: file-like objects are streamed in chunks (async wraps them in an async iterator; sync passes them to httpx directly, text-mode IO is encoded chunk-by-chunk).
- **`Volume.readFile(format="stream")` / `read_file(format="stream")`** — the request timeout now bounds only the initial handshake, not the body read, matching the sandbox `files.read` stream path. A dropped connection during the handshake surfaces the same typed, health-checked error; JS supports `signal` to cancel an in-flight stream and cancels unconsumed bodies on error so the pooled connection is released.

## Usage

JS — stream a file straight to a volume without buffering:
```ts
import { createReadStream } from 'node:fs'
import { Readable } from 'node:stream'

const stream = Readable.toWeb(createReadStream('large-input.bin'))
await volume.writeFile('/data/large-input.bin', stream)

// read back as a stream; the body lives until consumed/cancelled
const out = await volume.readFile('/data/large-input.bin', { format: 'stream' })
for await (const chunk of out) {
  // process chunk
}
```

Python — stream a file-like object:
```python
with open("large-input.bin", "rb") as f:
    volume.write_file("/data/large-input.bin", f)  # streamed, not read() into memory

for chunk in volume.read_file("/data/large-input.bin", format="stream"):
    ...  # process chunk
```

## Testing

- `pnpm run format`, `pnpm run lint`, `pnpm run typecheck` pass.
- Added volume streaming tests (JS `tests/volume/file.test.ts`; Python sync/async `test_file.py` text-stream cases).


packages/python-sdk/e2b/api/__init__.py
```diff
@@ -129,45 +127,30 @@ def __init__(
...
 
         headers = {
             **default_headers,
             **(config.headers or {}),
         }
 
+        # Deprecated: send the access token alongside the API key when one is
+        # available, mirroring the JS SDK. Prefer `api_headers` instead.
+        if config.access_token is not None:
+            headers["Authorization"] = f"Bearer {config.access_token}"
```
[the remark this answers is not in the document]
mishushakov: unrelated

packages/js-sdk/src/volume/index.ts
```diff
@@ -524,12 +529,13 @@ export class Volume {
    */
   async readFile(
     path: string,
-    opts?: VolumeApiOpts & { format: 'stream' }
+    opts?: VolumeApiOpts & { format: 'stream'; streamIdleTimeoutMs?: number }
```
mishushakov: can we put this into VolumeReadOpts ?

packages/js-sdk/src/volume/index.ts
```diff
@@ -524,12 +529,13 @@ export class Volume {
...
   ): Promise<ReadableStream<Uint8Array>>
   async readFile(
     path: string,
     opts?: VolumeApiOpts & {
       format?: 'text' | 'stream' | 'bytes' | 'blob'
+      streamIdleTimeoutMs?: number
```
mishushakov: can we put this into VolumeWriteOpts ?

packages/js-sdk/src/volume/index.ts
mishushakov: Done — extracted a named `VolumeReadOpts` type (in `volume/types.ts`, mirroring the sandbox `FilesystemReadOpts`) that holds `streamIdleTimeoutMs`, and the `readFile` overloads/impl now use `VolumeReadOpts & { format: ... }` instead of the inline literal. Exported from `volume/index.ts` and the package root.

packages/js-sdk/src/volume/index.ts
mishushakov: Same here — both stream-opt additions now come from `VolumeReadOpts`, so the inline `streamIdleTimeoutMs` is gone from the impl signature too. (Read this as the read-opts extraction since it's on the `readFile` impl.)

matthewlouisbrockman (commented):
hmm, don't think there's timeouts on this is there?

packages/js-sdk/tests/sandbox/files/read.test.ts
```diff
@@ -77,3 +77,50 @@ sandboxTest('read empty file in all formats', async ({ sandbox }) => {
...
+  expect(blob.size).toBe(0)
+
+  const stream = await sandbox.files.read(filename, { format: 'stream' })
+  expect(stream).toBeInstanceOf(ReadableStream)
+  const chunks: Uint8Array[] = []
+  for await (const chunk of stream as unknown as AsyncIterable<Uint8Array>) {
+    chunks.push(chunk)
+  }
+  expect(Buffer.concat(chunks).length).toBe(0)
+})
```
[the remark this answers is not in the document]
mishushakov: False positive — there's no duplication here. Each test name (`read file as stream`, `read non-existing file as stream`, `read empty file in all formats`) is defined exactly once; the `read empty file in all formats` case is just a long single test spanning lines 65–126, not a repeat of an earlier block.

packages/python-sdk/e2b/connection_config.py
```diff
@@ -10,6 +10,11 @@
 
 REQUEST_TIMEOUT: float = 60.0  # 60 seconds
 
+# Timeout for volume file transfers, which stream large bodies and so must not
+# inherit the short REQUEST_TIMEOUT. (Sandbox filesystem streaming instead
+# bounds each chunk by the request timeout and leaves the total to the server.)
+FILE_TIMEOUT: float = 3600.0  # 1 hour
```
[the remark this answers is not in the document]
mishushakov: Fixed in 037b73a4f — removed the unused `FILE_TIMEOUT` from `e2b.connection_config`. All consumers import it from `e2b.volume.connection_config` (which keeps its own copy), so the top-level one was dead code; moved the explanatory comment over to the volume constant that's actually used.

matthewlouisbrockman (approved):

PR #1457, opened 2026-06-17
fix(python): avoid quadratic stdout/stderr accumulation in command handles

davidzeng-pplx:
## Problem

`AsyncCommandHandle` / `CommandHandle` accumulate streamed output with `self._stdout += out` per chunk. Because `self._stdout` is an instance attribute (`STORE_ATTR`), CPython's in-place string-concatenation optimization — which only applies to local `STORE_FAST` targets — doesn't apply, so each append re-copies the entire buffer. For commands that emit large volumes of output this becomes O(n²) in total bytes, and in async contexts it stalls the event loop for hundreds of ms per chunk near the tail.

## Fix

Buffer decoded chunks in a `list[str]` and `"".join()` them on read. This restores linear-time accumulation and keeps streaming responsive, with no change to the resulting `stdout`/`stderr` values or the public API.

## Notes

- Applies the same change to both the sync and async command handles.
- Pure internal change; the incremental UTF-8 decoding behavior is preserved.
- Changeset included (`@e2b/python-sdk`, patch).


davidzeng-pplx:
@cla-bot check

mishushakov:
can you provide a reproduction where we can see the issue happening? thanks!

davidzeng-pplx:
@mishushakov  Sure! Since the slowdown is purely in the SDK's client-side accumulation (not the sandbox), here's a self-contained reproduction isolating the exact pattern from command_handle.py — no sandbox needed.

The handles do self._stdout += chunk per chunk. Because _stdout is an instance attribute (STORE_ATTR), CPython's in-place string-concat optimization — which only applies to local STORE_FAST targets — doesn't kick in, so each append re-copies the whole buffer → O(n²) in total bytes.

python
import time

class Buggy:   # current SDK
    def __init__(self): self._stdout = ""
    def feed(self, c): self._stdout += c

class Fixed:   # this PR
    def __init__(self): self._chunks = []
    def feed(self, c): self._chunks.append(c)

def run(cls, n, size=1024):
    h, chunk = cls(), "x" * size
    t = time.perf_counter()
    for _ in range(n): h.feed(chunk)
    return time.perf_counter() - t

for n in (2000, 4000, 8000, 16000):
    print(n, f"buggy={run(Buggy,n):.3f}s  fixed={run(Fixed,n):.4f}s")
Results on CPython 3.11 (chunks → buggy → fixed):

2,000 (1 MiB): 0.98s → 0.001s

4,000 (3 MiB): 3.33s → 0.003s

8,000 (7 MiB): 15.41s → 0.011s

16,000 (15 MiB): 72.27s → 0.014s

Each doubling ~quadruples the buggy time (4× = 2², classic O(n²)), while list+join stays flat in ms. A command streaming ~15 MiB of stdout blocks for over a minute today (and stalls the event loop in async), vs ~14 ms with this PR.

matthewlouisbrockman (approved):
good catch

matthewlouisbrockman:
Alright so may need to merge this into a branch and then merge that branch in to pass the tests but double checking if we have another way / what's going on here

matthewlouisbrockman:
merged to https://github.com/e2b-dev/E2B/pull/1472

PR #1471, opened 2026-06-22
feat(sdk): filesystem-only auto-pause via lifecycle.onTimeout object form

bchalios:
## Filesystem-only auto-pause (`onTimeout` object form)

Adds an object form to the sandbox **lifecycle** `onTimeout` (`on_timeout` in Python) that controls the snapshot kind taken when a sandbox auto-pauses on timeout, via `keepMemory` (`keep_memory`).

`onTimeout` now accepts either the existing bare action (`'pause'` / `'kill'`) or the object form `{ action, keepMemory }`. When `keepMemory` is `false` (with `action: 'pause'`), a timeout auto-pause takes a **filesystem-only** snapshot (no memory) instead of a full memory one, so the sandbox cold-boots (reboots) from disk on resume — losing running processes and open connections. Defaults to `true` (full memory snapshot), so existing callers are unaffected. **The bare string form is unchanged.**

It's the create-time / auto-pause counterpart to the explicit `pause(keepMemory=false)` from #1465: same `keepMemory` naming, mapped onto the `autoPauseMemory` create field.

### Type safety
The object form is a **discriminated union** on `action`: `keepMemory` is only valid with `action: 'pause'`. Pairing it with `action: 'kill'` is a **compile-time type error** (TS) / static error (`ty`), and is additionally rejected at runtime (`InvalidArgumentError` / `InvalidArgumentException`) for untyped callers.

### Behavior & validation
- `keepMemory` only applies to a `pause` action.
- **Incompatible with auto-resume** — auto-resume wakes a paused sandbox on inbound traffic by restoring its memory snapshot in place; a filesystem-only snapshot has no memory to restore (resuming cold-boots it), so it must be resumed explicitly via `connect()`. Combining `keepMemory: false` with `autoResume` is rejected client-side.

### Usage
```ts
// JS/TS — filesystem-only auto-pause on timeout
const sbx = await Sandbox.create({
  lifecycle: { onTimeout: { action: 'pause', keepMemory: false } },
})

// bare string form still works (full memory snapshot)
const sbx2 = await Sandbox.create({ lifecycle: { onTimeout: 'pause' } })
```
```python
# Python
sbx = Sandbox.create(
    lifecycle={"on_timeout": {"action": "pause", "keep_memory": False}}
)
```

### Changes
- `spec/openapi.yml`: `autoPauseMemory` on the create body (+ regenerated JS/Python clients).
- JS `SandboxOnTimeout` discriminated union (`'pause' | 'kill' | { action: 'pause'; keepMemory? } | { action: 'kill' }`) and the Python `SandboxOnTimeoutPause` / `SandboxOnTimeoutKill` TypedDicts, wired through `createSandbox` / `_create_sandbox` (sync + async) to `autoPauseMemory`, with the client-side guards.
- Tests: payload serialization + validation (offline, incl. the `action: 'kill'` type/runtime guard) and live cold-boot e2e in both SDKs; changeset (`e2b` + `@e2b/python-sdk`, minor).

### Backend dependency
The live e2e tests exercise the real auto-pause→cold-boot path and require the infra-side `autoPauseMemory` support (e2b-dev/infra#3055), now merged and deployed.

### DIFF
--- a/.changeset/auto-pause-filesystem-only.md
+++ b/.changeset/auto-pause-filesystem-only.md
@@ -3,25 +3,31 @@
 "@e2b/python-sdk": minor
 ---
 
-Add a `keepMemory` (`keep_memory` in Python) option to the sandbox `lifecycle`
-config, controlling the snapshot kind taken when a sandbox auto-pauses on
-timeout.
+Add an object form to the sandbox `lifecycle.onTimeout` (`on_timeout` in Python)
+that controls the snapshot kind taken when a sandbox auto-pauses on timeout, via
+`keepMemory` (`keep_memory`).
 
-When `keepMemory` is `false`, a timeout auto-pause drops the in-memory state and
-persists only the filesystem (a filesystem-only snapshot); resuming such a
-sandbox cold-boots (reboots) it from disk, losing running processes and open
-connections. Defaults to `true` (full memory snapshot). It only applies when
-`onTimeout`/`on_timeout` is `pause`, and cannot be combined with auto-resume (a
-filesystem-only snapshot must be resumed explicitly).
+`onTimeout` now accepts either the existing bare action (`'pause'` / `'kill'`) or
+the object form. The object form is a discriminated union on `action`:
+`keepMemory` is only accepted alongside `action: 'pause'` — pairing it with
+`action: 'kill'` is a compile-time type error (and is rejected at runtime for
+untyped callers). When `keepMemory` is `false`, a timeout auto-pause drops the
+in-memory state and persists only the filesystem (a filesystem-only snapshot);
+resuming such a sandbox cold-boots (reboots) it from disk, losing running
+processes and open connections. Defaults to `true` (full memory snapshot). It
+cannot be combined with auto-resume (a filesystem-only snapshot must be resumed
+explicitly). The bare string form is unchanged.
 
 ```python
 # Python
-sbx = Sandbox.create(lifecycle={"on_timeout": "pause", "keep_memory": False})
+sbx = Sandbox.create(
+    lifecycle={"on_timeout": {"action": "pause", "keep_memory": False}}
+)
 ```
 
 ```ts
 // JS/TS
 const sbx = await Sandbox.create({
-  lifecycle: { onTimeout: 'pause', keepMemory: false },
+  lifecycle: { onTimeout: { action: 'pause', keepMemory: false } },
 })
 ```
--- a/.changeset/auto-pause-filesystem-only.md
+++ b/.changeset/auto-pause-filesystem-only.md
@@ -15,8 +15,10 @@ untyped callers). When `keepMemory` is `false`, a timeout auto-pause drops the
 in-memory state and persists only the filesystem (a filesystem-only snapshot);
 resuming such a sandbox cold-boots (reboots) it from disk, losing running
 processes and open connections. Defaults to `true` (full memory snapshot). It
-cannot be combined with auto-resume (a filesystem-only snapshot must be resumed
-explicitly). The bare string form is unchanged.
+cannot be combined with auto-resume: auto-resume wakes a paused sandbox on
+inbound traffic by restoring its memory snapshot in place, and a filesystem-only
+snapshot has no memory to restore (resuming cold-boots it), so it must be resumed
+explicitly. The bare string form is unchanged.
 
 ```python
 # Python
--- a/.changeset/auto-pause-filesystem-only.md
+++ b/.changeset/auto-pause-filesystem-only.md
@@ -16,7 +16,7 @@ filesystem-only snapshot must be resumed explicitly).
 
 ```python
 # Python
-sbx = Sandbox(lifecycle={"on_timeout": "pause", "keep_memory": False})
+sbx = Sandbox.create(lifecycle={"on_timeout": "pause", "keep_memory": False})
 ```
 
 ```ts
--- a/.changeset/auto-pause-filesystem-only.md
+++ b/.changeset/auto-pause-filesystem-only.md
@@ -1,0 +1,27 @@
+---
+"e2b": minor
+"@e2b/python-sdk": minor
+---
+
+Add a `keepMemory` (`keep_memory` in Python) option to the sandbox `lifecycle`
+config, controlling the snapshot kind taken when a sandbox auto-pauses on
+timeout.
+
+When `keepMemory` is `false`, a timeout auto-pause drops the in-memory state and
+persists only the filesystem (a filesystem-only snapshot); resuming such a
+sandbox cold-boots (reboots) it from disk, losing running processes and open
+connections. Defaults to `true` (full memory snapshot). It only applies when
+`onTimeout`/`on_timeout` is `pause`, and cannot be combined with auto-resume (a
+filesystem-only snapshot must be resumed explicitly).
+
+```python
+# Python
+sbx = Sandbox(lifecycle={"on_timeout": "pause", "keep_memory": False})
+```
+
+```ts
+// JS/TS
+const sbx = await Sandbox.create({
+  lifecycle: { onTimeout: 'pause', keepMemory: false },
+})
+```
--- a/packages/js-sdk/src/sandbox/sandboxApi.ts
+++ b/packages/js-sdk/src/sandbox/sandboxApi.ts
@@ -248,7 +248,9 @@ export type SandboxLifecycle = {
   /**
    * Auto-resume enabled flag.
    * @default false
-   * Can be `true` only when the resolved timeout action is `pause`.
+   * Can be `true` only when `onTimeout` is `pause`. Not supported when
+   * `keepMemory` is `false` (a filesystem-only snapshot must be resumed
+   * explicitly via `connect()`).
    */
   autoResume?: boolean
 }
@@ -1161,8 +1163,6 @@ export class SandboxApi {
       allow_internet_access: opts?.allowInternetAccess ?? true,
       network: buildNetworkBody(opts?.network),
       autoPause: action === 'pause',
-      // Only relevant to a timeout auto-pause; omit it otherwise so the body
-      // matches the field's contract (keepMemory is forced true when not pause).
       autoPauseMemory: action === 'pause' ? keepMemory : undefined,
       autoResume: { enabled: autoResume },
     }
--- a/packages/js-sdk/src/sandbox/sandboxApi.ts
+++ b/packages/js-sdk/src/sandbox/sandboxApi.ts
@@ -210,6 +210,19 @@ export type SandboxLifecycle = {
    * Can be `true` only when `onTimeout` is `pause`.
    */
   autoResume?: boolean
+
+  /**
+   * Whether a timeout auto-pause keeps a full memory snapshot.
+   *
+   * When `false`, the auto-pause drops the in-memory state and persists only the
+   * filesystem (a filesystem-only snapshot); resuming such a sandbox cold-boots
+   * (reboots) it from disk, losing running processes and open connections. Only
+   * relevant when `onTimeout` is `pause`, and cannot be combined with
+   * `autoResume` (a filesystem-only snapshot must be resumed explicitly).
+   *
+   * @default true
+   */
+  keepMemory?: boolean
 }
 
 export type SandboxInfoLifecycle = {
@@ -1081,10 +1094,23 @@ export class SandboxApi {
     const client = new ApiClient(config)
     const onTimeout = opts?.lifecycle?.onTimeout ?? 'kill'
     const autoResume = opts?.lifecycle?.autoResume ?? false
+    const keepMemory = opts?.lifecycle?.keepMemory ?? true
 
     if (autoResume && onTimeout !== 'pause') {
       throw new InvalidArgumentError(
         "autoResume can only be true when the resolved onTimeout is 'pause'."
+      )
+    }
+
+    if (!keepMemory && onTimeout !== 'pause') {
+      throw new InvalidArgumentError(
+        "lifecycle.keepMemory=false only applies when onTimeout is 'pause'."
+      )
+    }
+
+    if (!keepMemory && autoResume) {
+      throw new InvalidArgumentError(
+        'lifecycle.keepMemory=false (filesystem-only auto-pause) cannot be combined with autoResume: a filesystem-only snapshot cannot be auto-resumed by traffic and must be resumed explicitly.'
       )
     }
 
@@ -1098,6 +1124,9 @@ export class SandboxApi {
       allow_internet_access: opts?.allowInternetAccess ?? true,
       network: buildNetworkBody(opts?.network),
       autoPause: onTimeout === 'pause',
+      // Only relevant to a timeout auto-pause; omit it otherwise so the body
+      // matches the field's contract (keepMemory is forced true when not pause).
+      autoPauseMemory: onTimeout === 'pause' ? keepMemory : undefined,
       autoResume: { enabled: autoResume },
     }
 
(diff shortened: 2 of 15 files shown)


bchalios:
Addressed the two actionable points from the Cursor summary (pushed):

1. **`autoPauseMemory` sent on every create** — now only sent when `onTimeout`/`on_timeout` is `pause`; omitted otherwise (JS sends `undefined`, Python leaves it `UNSET`). `keepMemory` is forced `true` when not pausing anyway, so this is behavior-preserving and matches the field's contract.
2. **Missing test for the `keepMemory=false` + non-`pause` guard** — added a dedicated offline test in all three SDKs: `keepMemory=false requires onTimeout pause` (JS) and `test_filesystem_only_auto_pause_requires_pause` (Python sync + async). Both guards (requires-pause and incompatible-with-auto-resume) now have offline coverage.

The third point (live cold-boot e2e depends on the backend `autoPauseMemory` support) is the known dependency on e2b-dev/infra#3055 — those tests stay red in CI until it deploys; the offline payload/validation tests pass independently.

JS typecheck/lint and Python ruff all pass.

.changeset/auto-pause-filesystem-only.md
```diff
@@ -0,0 +1,27 @@
...
+persists only the filesystem (a filesystem-only snapshot); resuming such a
+sandbox cold-boots (reboots) it from disk, losing running processes and open
+connections. Defaults to `true` (full memory snapshot). It only applies when
+`onTimeout`/`on_timeout` is `pause`, and cannot be combined with auto-resume (a
+filesystem-only snapshot must be resumed explicitly).
+
+```python
+# Python
+sbx = Sandbox(lifecycle={"on_timeout": "pause", "keep_memory": False})
+```
```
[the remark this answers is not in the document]
bchalios: Good catch — fixed in 6fb0818bb. `Sandbox(lifecycle=...)` hits the deprecated `__init__` (which takes only internal `SandboxOpts` fields, no `lifecycle`); the option is on `Sandbox.create()`. Updated both the changeset and the PR description to `Sandbox.create(lifecycle={...})`, matching the JS example and the live e2e tests.

packages/python-sdk/tests/async/sandbox_async/test_create.py
```diff
@@ -58,6 +59,76 @@ def test_create_payload_deserializes_auto_resume_enabled():
...
+async def test_filesystem_only_auto_pause_requires_pause():
+    # keep_memory only governs a timeout auto-pause, so keep_memory=False without
+    # on_timeout="pause" is rejected client-side.
+    with pytest.raises(InvalidArgumentException):
+        await AsyncSandbox.create(
+            timeout=3,
+            lifecycle={"on_timeout": "kill", "keep_memory": False},
+        )
+
+
```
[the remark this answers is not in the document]
bchalios: Fixed — added `@pytest.mark.skip_debug()` to the four Python validation tests and `test.skipIf(isDebug)` to the two JS ones, matching the e2e tests in the same files. Confirmed the cause: `Sandbox.create` returns a debug sandbox before `_create_sandbox`/`createSandbox` (where the lifecycle validation lives), so under `E2B_DEBUG` the validation never runs. (The deeper fix — moving client-side validation ahead of the debug short-circuit — is a wider refactor that also touches the pre-existing auto_resume guard, so out of scope here.)

packages/python-sdk/e2b/sandbox_async/sandbox_api.py
```diff
@@ -204,16 +204,32 @@ async def _create_sandbox(
...
+            )
+
         network_body = build_network_config(network)
         body = NewSandbox(
             template_id=template,
             auto_pause=on_timeout == "pause",
+            # Only relevant to a timeout auto-pause; leave it unset otherwise so the
+            # body matches the field's contract (keep_memory is forced True when
+            # on_timeout is not "pause").
+            auto_pause_memory=keep_memory if on_timeout == "pause" else UNSET,
```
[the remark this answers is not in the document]
bchalios: Fixed in the latest commit — a missing or explicit `None` `keep_memory` now defaults to `True` (full memory) in both sync and async, matching the JS `?? true`. `.get("keep_memory", True)` only used the default for an absent key, so an explicit `None` read as `False` (tripping validation / sending `autoPauseMemory: null`). JS already handled it correctly via `?? true`, so no JS change needed.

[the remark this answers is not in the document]
bchalios: Added a regression test (`test_keep_memory_none_defaults_to_full_memory`, sync + async): a timeout auto-pause with `keep_memory=None` resumes the **same** sandbox in place with an **unchanged boot id** — proving None defaults to a full memory snapshot, not filesystem-only (which would cold-boot). Passes live.

packages/python-sdk/tests/sync/sandbox_sync/test_create.py
```diff
@@ -59,6 +61,105 @@ def test_create_payload_deserializes_auto_resume_enabled():
     assert body.auto_resume.to_dict() == {"enabled": False}
 
 
+def test_create_payload_serializes_auto_pause_memory():
```
mishushakov: unnecessary

packages/python-sdk/tests/sync/sandbox_sync/test_create.py
```diff
@@ -59,6 +61,105 @@ def test_create_payload_deserializes_auto_resume_enabled():
...
+    sandbox = sandbox_factory(
+        timeout=3,
+        lifecycle={"on_timeout": {"action": "pause", "keep_memory": False}},
+    )
+
+    marker = "auto-pause-fs-only"
+    sandbox.commands.run(f"echo {marker} > /home/user/auto-pause-marker.txt")
+    boot_before = sandbox.commands.run(
+        "cat /proc/sys/kernel/random/boot_id"
+    ).stdout.strip()
```
mishushakov: we have files.write and files.read in the SDK

packages/python-sdk/tests/sync/sandbox_sync/test_create.py
```diff
@@ -59,6 +61,105 @@ def test_create_payload_deserializes_auto_resume_enabled():
...
+    assert not sandbox.is_running()
+
+    # A filesystem-only snapshot cannot auto-resume on traffic; connect resumes
+    # it by cold-booting.
+    resumed = sandbox.connect()
+    assert resumed.is_running()
+
+    persisted = resumed.commands.run(
+        "cat /home/user/auto-pause-marker.txt"
+    ).stdout.strip()
```
mishushakov: same as above

packages/python-sdk/tests/sync/sandbox_sync/test_create.py
```diff
@@ -59,6 +61,105 @@ def test_create_payload_deserializes_auto_resume_enabled():
...
+    assert persisted == marker
+
+    boot_after = resumed.commands.run(
+        "cat /proc/sys/kernel/random/boot_id"
+    ).stdout.strip()
```
mishushakov: same

packages/python-sdk/tests/async/sandbox_async/test_create.py
mishushakov, line 1: check my sync tests comments

packages/js-sdk/src/sandbox/sandboxApi.ts
```diff
@@ -1079,12 +1120,34 @@ export class SandboxApi {
...
+    const keepMemory =
+      typeof onTimeout !== 'string' && 'keepMemory' in onTimeout
+        ? (onTimeout.keepMemory ?? true)
+        : true
     const autoResume = opts?.lifecycle?.autoResume ?? false
 
-    if (autoResume && onTimeout !== 'pause') {
+    if (hasKeepMemory && action !== 'pause') {
+      throw new InvalidArgumentError(
+        "onTimeout.keepMemory is not allowed when action is 'kill'."
```
mishushakov: onTimeout.keepMemory is only allowed when action is 'pause'.

packages/js-sdk/src/sandbox/sandboxApi.ts
```diff
@@ -1079,12 +1120,34 @@ export class SandboxApi {
...
+      )
+    }
+
+    if (autoResume && action !== 'pause') {
+      throw new InvalidArgumentError(
+        "autoResume can only be true when the resolved onTimeout action is 'pause'."
```
mishushakov: autoResume can only be true when onTimeout action is 'pause'.

packages/js-sdk/src/sandbox/sandboxApi.ts
```diff
@@ -1079,12 +1120,34 @@ export class SandboxApi {
...
+      )
+    }
+
+    if (!keepMemory && autoResume) {
       throw new InvalidArgumentError(
-        "autoResume can only be true when the resolved onTimeout is 'pause'."
+        'onTimeout keepMemory=false (filesystem-only auto-pause) cannot be combined with autoResume: a filesystem-only snapshot cannot be auto-resumed by traffic and must be resumed explicitly.'
```
mishushakov: autoResume: 'true' is not a valid value when keepMemory: 'false' - a filesystem-only snapshot cannot be auto-resumed by traffic and must be resumed explicitly using Sandbox.connect

packages/js-sdk/src/sandbox/sandboxApi.ts
```diff
@@ -1097,7 +1160,10 @@ export class SandboxApi {
       secure: opts?.secure ?? true,
       allow_internet_access: opts?.allowInternetAccess ?? true,
       network: buildNetworkBody(opts?.network),
-      autoPause: onTimeout === 'pause',
+      autoPause: action === 'pause',
+      // Only relevant to a timeout auto-pause; omit it otherwise so the body
+      // matches the field's contract (keepMemory is forced true when not pause).
```
mishushakov: unnecessary comment

packages/js-sdk/tests/sandbox/lifecyclePayload.test.ts
```diff
@@ -31,6 +66,51 @@ test.skipIf(isDebug)(
...
+    })
+
+    try {
+      const marker = 'auto-pause-fs-only'
+      await sandbox.commands.run(
+        `echo ${marker} > /home/user/auto-pause-marker.txt`
+      )
+      const bootBefore = (
+        await sandbox.commands.run('cat /proc/sys/kernel/random/boot_id')
+      ).stdout.trim()
```
mishushakov: use sandbox.files for consistency

packages/js-sdk/tests/sandbox/lifecyclePayload.test.ts
```diff
@@ -31,6 +66,51 @@ test.skipIf(isDebug)(
...
+
+      await wait(5_000)
+
+      assert.equal((await sandbox.getInfo()).state, 'paused')
+      assert.isFalse(await sandbox.isRunning())
```
mishushakov: actually this might be unnecessary - it only calls sandbox /healthcheck endpoint

packages/js-sdk/tests/sandbox/lifecyclePayload.test.ts
```diff
@@ -31,6 +66,51 @@ test.skipIf(isDebug)(
...
+      assert.isTrue(await sandbox.isRunning())
+
+      const persisted = (
+        await sandbox.commands.run('cat /home/user/auto-pause-marker.txt')
+      ).stdout.trim()
+      assert.equal(persisted, marker)
+
+      const bootAfter = (
+        await sandbox.commands.run('cat /proc/sys/kernel/random/boot_id')
+      ).stdout.trim()
```
mishushakov: same as above

packages/python-sdk/e2b/sandbox/sandbox_api.py
```diff
@@ -239,21 +239,66 @@ class SandboxNetworkInfo(TypedDict, total=False):
...
+
+
 class SandboxLifecycle(TypedDict):
     """
     Sandbox lifecycle configuration; defines post-timeout behavior and auto-resume settings.
     Defaults to `on_timeout="kill"` and `auto_resume=False`.
     """
 
-    on_timeout: Literal["pause", "kill"]
+    on_timeout: Union[Literal["pause", "kill"], SandboxOnTimeout]
```
mishushakov: can't you just move Literal["pause", "kill"] to SandboxOnTimeout Union?

packages/python-sdk/e2b/sandbox/sandbox_api.py
```diff
@@ -239,21 +239,66 @@ class SandboxNetworkInfo(TypedDict, total=False):
...
     """
-    What should happen to the sandbox when timeout is reached. `"kill"` means the sandbox will be terminated, while `"pause"` means the sandbox will be paused and can be resumed later. Defaults to `"kill"`.
+    What should happen to the sandbox when timeout is reached. `"kill"` terminates
+    the sandbox; `"pause"` pauses it for later resume. Accepts either the bare
+    action or an object `{"action": "pause", "keep_memory": ...}` /
+    `{"action": "kill"}` to also control the pause snapshot kind. Defaults to
+    `"kill"`.
     """
 
     auto_resume: NotRequired[bool]
```
mishushakov, line 302: add comment here is that only supported with pause with keep_memory false

packages/python-sdk/e2b/sandbox/sandbox_api.py
```diff
@@ -239,21 +239,66 @@ class SandboxNetworkInfo(TypedDict, total=False):
...
     """
     Whether activity should cause the sandbox to resume when paused. Defaults to `False`.
-    Can be `True` only when `on_timeout` is `pause`.
+    Can be `True` only when the resolved timeout action is `pause`.
```
mishushakov: I'd avoid the "resolved timeout action" wording - it's vague, `onTimeout` was much clearer

packages/js-sdk/src/sandbox/sandboxApi.ts
```diff
@@ -197,17 +197,58 @@ export type SandboxNetworkUpdate = {
...
-  onTimeout: 'pause' | 'kill'
+  onTimeout: SandboxOnTimeout
 
   /**
    * Auto-resume enabled flag.
    * @default false
-   * Can be `true` only when `onTimeout` is `pause`.
+   * Can be `true` only when the resolved timeout action is `pause`.
    */
   autoResume?: boolean
```
mishushakov, line 255: would add a note here that it does not support when keepMemory is false

packages/python-sdk/e2b/sandbox_async/sandbox_api.py
```diff
@@ -202,18 +202,50 @@ async def _create_sandbox(
...
+                "a filesystem-only snapshot cannot be auto-resumed by traffic and must be resumed explicitly."
+            )
+
         network_body = build_network_config(network)
         body = NewSandbox(
             template_id=template,
             auto_pause=on_timeout == "pause",
+            # Only relevant to a timeout auto-pause; leave it unset otherwise so the
+            # body matches the field's contract (keep_memory is forced True when
+            # on_timeout is not "pause").
```
mishushakov: self-explanatory

packages/python-sdk/e2b/sandbox_sync/sandbox_api.py
mishushakov, line 1: same applies as in async

mishushakov (changes requested):
let's do final round

packages/python-sdk/tests/sync/sandbox_sync/test_create.py
```diff
@@ -59,6 +61,87 @@ def test_create_payload_deserializes_auto_resume_enabled():
...
+        timeout=60,
+        lifecycle={"on_timeout": {"action": "pause", "keep_memory": None}},
+    )
+    boot_before = sbx.files.read("/proc/sys/kernel/random/boot_id").strip()
+
+    sbx.set_timeout(0)  # force the timeout auto-pause now
+    for _ in range(150):
+        if not sbx.is_running():
+            break
+        sleep(0.2)
```
mishushakov, line 116: this could potentially halt CI for `150*0.2*n`

mishushakov (approved):

PR #1476, opened 2026-06-23
Add test file

matthewlouisbrockman:
just testing what's up with some failing tests


matthewlouisbrockman:
well, it only ran on ones that were changed it turns out so that didn't help

PR #1479, opened 2026-06-24
feat(templates): template list pagination via GET /v2/templates

huv1k:
## Summary

Adds support for listing templates with pagination via the new `GET /v2/templates` endpoint (added to `spec/openapi.yml`; clients regenerated). The JS and Python SDKs gain a `Template.list()` paginator that yields `TemplateInfo` and is built on the shared paginator base from #1491 (`Paginator` in JS, `PaginatorBase` in Python), mirroring `Sandbox.list()`. The CLI's `e2b template list` pages through results behind a `-l, --limit` option (default 1000, 100 per page, `0` for no limit) with a "Showing first N…" hint — aligned with `e2b sandbox list`. Pagination uses the `limit`/`nextToken` query params and the `X-Next-Token` response header. The API key is team-scoped, so listing takes no team identifier.

## What's included

- **Spec/codegen:** `GET /v2/templates` (`limit` + `nextToken`, `Template[]`), regenerated JS + Python clients.
- **JS SDK (`e2b`):** `Template.list()` → `TemplatePaginator` (`hasNext` / `nextItems()`) returning `TemplateInfo`. Lives in `src/template/templateApi.ts` and depends only on shared infra (`Paginator`, `ConnectionOpts`).
- **Python SDK (`@e2b/python-sdk`):** `Template.list()` / `AsyncTemplate.list()` → `TemplatePaginator` / `AsyncTemplatePaginator` returning `TemplateInfo`, living in the template package (`e2b/template/template_api.py`, `e2b/template_{sync,async}/paginator.py`) and built on the shared `PaginatorBase`.
- **CLI (`@e2b/cli`):** `e2b template list` pagination matching `sandbox list`; the `delete`/`publish` `--select` flows consume the same helper.
- **Tests:** integration tests (no mocks) for JS, CLI, and Python sync + async — they build a template and assert it shows up via `Template.list()` against the live API.
- **Changeset:** minor bump for `e2b`, `@e2b/python-sdk`, `@e2b/cli`.

## Usage

CLI:
```bash
e2b template list                 # up to 1000 templates (100 per page)
e2b template list --limit 20      # cap at 20; prints "Showing first 20… Use --limit to change."
e2b template list --limit 0 -f json   # no cap, JSON output
```

JS SDK:
```ts
import { Template } from 'e2b'

const paginator = Template.list({ limit: 100 })
while (paginator.hasNext) {
  const templates = await paginator.nextItems()
  console.log(templates)
}
```

Python SDK:
```python
from e2b import Template  # or AsyncTemplate

paginator = Template.list(limit=100)
while paginator.has_next:
    for template in paginator.next_items():
        print(template.template_id)
```

## Notes

- `TemplatePaginator.nextItems()` returns `[]` when exhausted (the sandbox/snapshot paginators still throw — to be unified separately).
- Resolves EN-603.


packages/cli/src/commands/template/list.ts
```diff
@@ -111,17 +125,62 @@ function renderTable(templates: e2b.components['schemas']['Template'][]) {
...
-    params: {
-      query: { teamID },
-    },
-  })
+  limit?: number
+}): Promise<ListSandboxTemplatesResult> {
+  let pageLimit = limit
+  if (!limit || limit > PAGE_LIMIT) {
+    pageLimit = PAGE_LIMIT
+  }
```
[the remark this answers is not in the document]
huv1k: This should be fixed

packages/js-sdk/src/template/templateApi.ts
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * Options for listing templates.
+ */
+export interface TemplateListOpts extends Omit<SandboxApiOpts, 'signal'> {
+  /**
+   * Identifier of the team whose templates should be listed. Defaults to the
+   * team the API key belongs to.
+   */
+  teamId?: string
```
mishushakov: that makes no sense, API key is scoped per team, I cannot check templates for a different team with same API key

packages/js-sdk/src/template/templateApi.ts
```diff
@@ -0,0 +1,185 @@
+import { ApiClient, components, handleApiError } from '../api'
+import { ConnectionConfig } from '../connectionConfig'
+import { BasePaginator, SandboxApiOpts } from '../sandbox/sandboxApi'
```
mishushakov: importing anything from Sandbox here is no-bueno, if it's shared it should go to shared utils.ts

packages/js-sdk/src/template/templateApi.ts
```diff
@@ -0,0 +1,185 @@
...
+ * @example
+ * ```ts
+ * const paginator = Template.list()
+ * while (paginator.hasNext) {
+ *   const templates = await paginator.nextItems()
+ *   console.log(templates)
+ * }
+ * ```
+ */
+export class TemplatePaginator extends BasePaginator<TemplateInfo> {
```
mishushakov: BasePaginator might not be re-usable here - its constructor opts are SandboxApiOpts
maybe you can make BasePaginator constructor more generic?

packages/js-sdk/src/template/templateApi.ts
```diff
@@ -0,0 +1,185 @@
...
+  constructor(opts?: TemplateListOpts) {
+    super(opts, opts?.limit, opts?.nextToken)
+
+    this.teamId = opts?.teamId
+  }
+
+  async nextItems(opts?: SandboxApiOpts): Promise<TemplateInfo[]> {
+    if (!this.hasNext) {
+      throw new Error('No more items to fetch')
+    }
```
mishushakov: please check in other places in SDK - but not having next items should not raise any error - instead remove empty []

if that's the case elsewhere - add some comment that we'll be changing this

packages/js-sdk/src/template/templateApi.ts
```diff
@@ -0,0 +1,185 @@
...
+        updatedAt: new Date(template.updatedAt),
+        lastSpawnedAt: template.lastSpawnedAt
+          ? new Date(template.lastSpawnedAt)
+          : null,
+        spawnCount: template.spawnCount,
+        buildCount: template.buildCount,
+        envdVersion: template.envdVersion,
+        createdBy: template.createdBy ?? null,
+        buildStatus: template.buildStatus,
+      })
```
mishushakov: I think you can just spread most of these

packages/python-sdk/e2b/sandbox/sandbox_api.py
mishushakov, line 1: none of the changes belong to sandbox_api - it should be on template SDK

packages/python-sdk/e2b/sandbox/sandbox_api.py
```diff
@@ -597,3 +667,15 @@ def __init__(
     ):
         super().__init__(limit=limit, next_token=next_token, **opts)
         self.query = query
+
+
+class TemplatePaginatorBase(PaginatorBase):
```
mishushakov: move PaginatorBase to utils if it's reused

packages/python-sdk/e2b/sandbox_async/paginator.py
mishushakov, line 1: again, none of the changes belong here, they belong to template SDK

packages/python-sdk/e2b/sandbox_async/paginator.py
```diff
@@ -140,3 +143,56 @@ async def next_items(self, **opts: Unpack[ApiParams]) -> List[SnapshotInfo]:
...
+        Call this method only if `has_next` is `True`, otherwise it will raise an exception.
+
+        :param opts: Per-call connection options (e.g. `api_key`, `domain`,
+            `headers`, `request_timeout`). When provided, this call uses these
+            options instead of the ones the paginator was constructed with.
+
+        :returns: List of templates
+        """
+        if not self.has_next:
+            raise Exception("No more items to fetch")
```
mishushakov: same as in JS

packages/python-sdk/e2b/sandbox_sync/paginator.py
mishushakov, line 1: everything I said above applies here

packages/python-sdk/e2b/template_async/main.py
```diff
@@ -10,6 +10,7 @@
 from e2b.template.main import TemplateBase, TemplateClass
 from e2b.template.types import BuildInfo, InstructionType, TemplateTag, TemplateTagInfo
 from e2b.template.utils import normalize_build_arguments, read_dockerignore
+from e2b.sandbox_async.paginator import AsyncTemplatePaginator
```
mishushakov: importing from sandbox here is no-bueno

packages/python-sdk/e2b/__init__.py
```diff
@@ -91,16 +91,25 @@
...
+    TemplateInfo,
 )
 from .sandbox_async.commands.command_handle import AsyncCommandHandle
 from .sandbox_async.filesystem.watch_handle import AsyncWatchHandle
 from .sandbox_async.main import AsyncSandbox
-from .sandbox_async.paginator import AsyncSandboxPaginator, AsyncSnapshotPaginator
+from .sandbox_async.paginator import (
+    AsyncSandboxPaginator,
+    AsyncSnapshotPaginator,
+    AsyncTemplatePaginator,
```
mishushakov: no bueno

packages/js-sdk/src/template/templateApi.ts
```diff
@@ -0,0 +1,185 @@
...
+      },
+      signal: config.getSignal(opts?.requestTimeoutMs, opts?.signal),
+    })
+
+    const err = handleApiError(res)
+    if (err) {
+      throw err
+    }
+
+    this.updatePagination(res.response)
```
mishushakov: is there a particular reason we have this method in the JS SDK but not in Python?

mishushakov (changes requested):
move template method to template and shared classes to utils

packages/cli/src/commands/template/list.ts
```diff
@@ -4,23 +4,32 @@ import * as e2b from 'e2b'
...
 import { teamOption } from '../../options'
-import { handleE2BRequestError } from '../../utils/errors'
+
+const DEFAULT_LIMIT = 1000
+const PAGE_LIMIT = 100
 
 export const listCommand = new commander.Command('list')
   .description('list sandbox templates')
   .alias('ls')
   .addOption(teamOption)
```
mishushakov: this option should be prob deprecated

mishushakov:
also, no tests for the Python SDKs?

packages/js-sdk/tests/template/list.test.ts
mishushakov, line 1: please make it integration test instead, not mock

packages/cli/tests/commands/template/list.test.ts
mishushakov, line 1: same here - no mock pls

packages/cli/src/commands/template/list.ts
```diff
@@ -29,6 +38,11 @@ export const listCommand = new commander.Command('list')
 
       if (format === 'pretty') {
         renderTable(templates)
+        if (hasMore) {
+          console.log(
+            `Showing first ${limit} templates. Use --limit to change.`
+          )
```
mishushakov: no need for this, our users can count

packages/cli/src/commands/template/list.ts
```diff
@@ -111,17 +125,72 @@ function renderTable(templates: e2b.components['schemas']['Template'][]) {
...
+    templates: limit ? templates.slice(0, limit) : templates,
+    // We can't change the page size during iteration, so we may have to check
+    // if we have more templates than the limit.
+    hasMore: paginator.hasNext || (limit ? templates.length > limit : false),
+  }
+}
+
+// Adapt the SDK's TemplateInfo back to the raw API schema shape the rest of the
+// CLI (table rendering, selection prompts, `--format json`) is built around.
+function toTemplateSchema(
```
mishushakov: is it actually an issue?

packages/cli/src/commands/template/list.ts
```diff
@@ -4,23 +4,32 @@ import * as e2b from 'e2b'
 
 import { listAliases } from '../../utils/format'
 import { sortTemplatesAliases } from 'src/utils/templateSort'
-import { client, ensureAPIKey, resolveTeamId } from 'src/api'
+import { ensureAPIKey, resolveTeamId } from 'src/api'
 import { teamOption } from '../../options'
-import { handleE2BRequestError } from '../../utils/errors'
+
+const DEFAULT_LIMIT = 1000
```
mishushakov: I think it's fine for now to just return the whole list, auto-paginate should be cheap op.
the reason is that ppl pipe the CLI output into other tools and we don't have a good way to paginate in the CLI - perhaps next task to figure out?

mishushakov (commented):
more CLI feedback

packages/cli/src/commands/template/list.ts
huv1k: Yes, we are calling SDK and it has different output then calling raw endpoint.

packages/python-sdk/e2b/template/template_api.py
mishushakov, line 1: this should be here
packages/python-sdk/e2b/template/types.py

packages/js-sdk/src/template/templateApi.ts
```diff
@@ -0,0 +1,173 @@
...
+ * @example
+ * ```ts
+ * const paginator = Template.list()
+ * while (paginator.hasNext) {
+ *   const templates = await paginator.nextItems()
+ *   console.log(templates)
+ * }
+ * ```
+ */
+export class TemplatePaginator extends Paginator<TemplateInfo, ConnectionOpts> {
```
mishushakov: this should be in template/paginator.ts

packages/js-sdk/src/template/templateApi.ts
```diff
@@ -0,0 +1,173 @@
+import { ApiClient, components, handleApiError } from '../api'
+import { ConnectionConfig, ConnectionOpts } from '../connectionConfig'
+import { TemplateError } from '../errors'
+import { Paginator } from '../paginator'
+
+/**
+ * Information about a sandbox template.
+ */
+export interface TemplateInfo {
```
mishushakov: this should be in template/types.ts

packages/cli/src/commands/template/list.ts
```diff
@@ -111,17 +120,66 @@ function renderTable(templates: e2b.components['schemas']['Template'][]) {
...
+    hasMore: paginator.hasNext || (limit ? templates.length > limit : false),
+  }
+}
+
+// Adapt the SDK's TemplateInfo back to the raw API schema shape the rest of the
+// CLI (table rendering, selection prompts, `--format json`) is built around.
+function toTemplateSchema(
+  template: e2b.TemplateInfo
+): e2b.components['schemas']['Template'] {
+  return {
```
mishushakov, line 158: you can spread all of them and only modify the date ones

packages/cli/src/commands/template/list.ts
```diff
@@ -4,31 +4,40 @@ import * as e2b from 'e2b'
...
-      })
+      const { templates, hasMore } = await listSandboxTemplates({ limit })
 
       for (const template of templates) {
         sortTemplatesAliases(template.aliases)
       }
 
       if (format === 'pretty') {
         renderTable(templates)
+        if (hasMore) {
```
mishushakov: as said, would remove this block, this is unnecessary info

packages/cli/src/commands/template/list.ts
```diff
@@ -4,31 +4,40 @@ import * as e2b from 'e2b'
 
 import { listAliases } from '../../utils/format'
 import { sortTemplatesAliases } from 'src/utils/templateSort'
-import { client, ensureAPIKey, resolveTeamId } from 'src/api'
-import { teamOption } from '../../options'
-import { handleE2BRequestError } from '../../utils/errors'
+import { ensureAPIKey } from 'src/api'
+
+const DEFAULT_LIMIT = 1000
```
mishushakov: remove, should be no limit

PR #1487, opened 2026-06-26
fix(cli): make migrated Python build scripts runnable

### LINKED ISSUE #1477: [Bug]: `e2b template migrate` Python output is not runnable as instructed (relative import vs `python build_dev.py`)
### Sandbox ID or Build ID

### Environment

- Python: 3.12.13
- e2b (Python SDK): 2.26.0

### Timestamp of the issue

2026-06-24 03:54 UTC

### Frequency

Happens every time

### Expected behavior

After running `e2b template migrate -l python-async`, the command prints instructions to run the generated build scripts:

```
🔨 To get started with your template:
   pip install e2b (install e2b dependency)
   python build_dev.py (run development build)
   python build_prod.py (run production build)
```

Following these instructions verbatim (`python build_dev.py`) should successfully invoke the build (or at least reach the build call without an import error).

### Actual behavior

`python build_dev.py` exits immediately with:

```
Traceback (most recent call last):
  File ".../e2b-template/build_dev.py", line 3, in <module>
    from .template import template
ImportError: attempted relative import with no known parent package
```

The generated scripts cannot be run as instructed. Same applies to `python build_prod.py`.

### Issue reproduction

1. In a directory with a valid `e2b.toml` (containing `team_id`, `template_id`, `dockerfile`) and `e2b.Dockerfile`, run:
   ```
   ```
2. The command completes successfully and prints the instructions above.
3. Run the instruction verbatim:
   ```
   python build_dev.py
   ```
4. Observe the `ImportError: attempted relative import with no known parent package`.

### Additional context

**Root cause**

The generated `build_dev.py` / `build_prod.py` use a relative import:

```python
from .template import template
```

…but they are emitted as standalone files (no `__init__.py` generated), and the migrate command instructs single-file execution (`python build_dev.py`). Relative imports require the script to be part of a package and executed as one (e.g. `python -m pkg.build_dev`), so the printed instruction is guaranteed to fail.

**Why the tests don't catch it**

The committed fixtures match the generated output verbatim, so the unit tests pass even though the generated files don't actually run:

- Template source (Handlebars): [`packages/cli/src/templates/python-build-async.hbs`](https://github.com/e2b-dev/e2b/blob/main/packages/cli/src/templates/python-build-async.hbs) — emits `from .template import template`
- User-facing instruction: [`packages/cli/src/commands/template/migrate.ts`](https://github.com/e2b-dev/e2b/blob/main/packages/cli/src/commands/template/migrate.ts) — prints `python build_dev.py`
- Fixture (treated as expected output): [`packages/cli/tests/commands/template/fixtures/minimal-dockerfile/expected/python-async/build_dev.py`](https://github.com/e2b-dev/e2b/blob/main/packages/cli/tests/commands/template/fixtures/minimal-dockerfile/expected/python-async/build_dev.py)

The same pattern is in `python-build-sync.hbs` (sync variant likely affected too).

**Workaround**

Manually change `from .template import template` to `from template import template` in each generated build script, then run `python build_dev.py` from the migrate output directory (Python adds the cwd to `sys.path`).

**Possible fixes (deferring to maintainers for direction)**

- **(A)** Change the Handlebars templates to `from template import template`. Single-file execution then works as instructed, and `python -m <pkg>.build_dev` still works as long as the script is run from a directory on `sys.path`. Simplest fix.
- **(B)** Keep the relative import and switch the layout / instruction to package-style execution: emit an `__init__.py` and print `python -m <pkg>.build_dev`. Requires a python-safe package name (no hyphens), which adds complexity if the migration directory name contains hyphens.

Happy to send a PR for either approach once you confirm the direction.

snvtac:
## Summary

Fixes #1477.

The Python build scripts emitted by `e2b template migrate --language python-sync` and `python-async` are documented as directly runnable with `python build_dev.py` / `python build_prod.py`, but they used a package-relative import (`from .template import template`) that fails outside package execution.

This changes the generated Python build scripts to import the sibling generated `template.py` directly, updates the migration fixtures, and adds an assertion so the migration test covers the intended standalone import behavior.

## Changes

- Use `from template import template` in the Python sync and async build templates.
- Update all Python sync/async migrate expected fixtures.
- Add migrate-test assertions that Python outputs do not contain the relative import.
- Add a patch changeset for `@e2b/cli`.

## Test plan

- [x] `pnpm install --frozen-lockfile`
- [x] `pnpm --filter @e2b/cli build`
- [x] `pnpm --filter @e2b/cli test -- tests/commands/template/migrate.test.ts` (24 passed)
- [x] `pnpm --filter @e2b/cli test -- tests/commands/template/init.test.ts` (14 passed)
- [x] `pnpm --filter @e2b/cli lint`
- [x] Python smoke test importing generated `build_dev.py` / `build_prod.py` for python-sync and python-async
- [x] `git diff --check`
- [ ] `pnpm --filter @e2b/cli test` fails in this environment because `tests/commands/template/create.test.ts` requires `E2B_API_KEY`; other CLI test files reported 10 passed / 2 skipped before that environment-gated failure.

### DIFF
--- a/.changeset/fix-python-migrate-build-import.md
+++ b/.changeset/fix-python-migrate-build-import.md
@@ -1,0 +1,5 @@
+---
+"@e2b/cli": patch
+---
+
+Fix Python template migration build scripts so the generated `build_dev.py` and `build_prod.py` files can be run directly with `python build_dev.py` as instructed.
--- a/packages/cli/src/templates/python-build-async.hbs
+++ b/packages/cli/src/templates/python-build-async.hbs
@@ -1,6 +1,6 @@
 import asyncio
 from e2b import AsyncTemplate, default_build_logger
-from .template import template
+from template import template
 
 
 async def main():
--- a/packages/cli/src/templates/python-build-sync.hbs
+++ b/packages/cli/src/templates/python-build-sync.hbs
@@ -1,5 +1,5 @@
 from e2b import Template, default_build_logger
-from .template import template
+from template import template
 
 
 if __name__ == "__main__":
--- a/packages/cli/tests/commands/template/fixtures/complex-python/expected/python-async/build_dev.py
+++ b/packages/cli/tests/commands/template/fixtures/complex-python/expected/python-async/build_dev.py
@@ -1,6 +1,6 @@
 import asyncio
 from e2b import AsyncTemplate, default_build_logger
-from .template import template
+from template import template
 
 
 async def main():
--- a/packages/cli/tests/commands/template/fixtures/complex-python/expected/python-async/build_prod.py
+++ b/packages/cli/tests/commands/template/fixtures/complex-python/expected/python-async/build_prod.py
@@ -1,6 +1,6 @@
 import asyncio
 from e2b import AsyncTemplate, default_build_logger
-from .template import template
+from template import template
 
 
 async def main():
--- a/packages/cli/tests/commands/template/fixtures/complex-python/expected/python-sync/build_dev.py
+++ b/packages/cli/tests/commands/template/fixtures/complex-python/expected/python-sync/build_dev.py
@@ -1,5 +1,5 @@
 from e2b import Template, default_build_logger
-from .template import template
+from template import template
 
 
 if __name__ == "__main__":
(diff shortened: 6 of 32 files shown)


mishushakov (approved):

mishushakov:
please sign the CLA and we would proceed to merge this

mishushakov:
@snvtac reminder to sign the CLA

mishushakov:
CLA has not been signed, we resolved the issue ourselves in https://github.com/e2b-dev/E2B/pull/1505

PR #1491, opened 2026-06-26
Introduce generic paginator base class for JS and Python SDKs

huv1k:
Extracts the cursor-based pagination state machine into a reusable base class — `Paginator` in the JS SDK's `utils`, `PaginatorBase` in `e2b/utils.py` — that owns `hasNext`/`nextToken` and the `x-next-token` header handling, and migrates the sandbox and snapshot paginators onto it. Each concrete paginator now just implements `nextItems`/`next_items` to fetch its own page, so future list endpoints (templates, builds, etc.) can add pagination by subclassing without reimplementing the bookkeeping. Applied equivalently to the JS SDK and both Python sync and async implementations, with unit tests covering the shared base. There are no public API changes — `Sandbox.list()` / `listSnapshots()` and the existing paginator types behave identically.

## Usage (unchanged)

```ts
const paginator = Sandbox.list()
while (paginator.hasNext) {
  const sandboxes = await paginator.nextItems()
  console.log(sandboxes)
}
```

```python
paginator = Sandbox.list()
while paginator.has_next:
    sandboxes = paginator.next_items()
    print(sandboxes)
```


packages/js-sdk/src/index.ts
```diff
@@ -127,6 +127,8 @@ import { Sandbox } from './sandbox'
 
 export default Sandbox
 
+export { Paginator } from './utils'
```
mishushakov: don't need to export here

packages/python-sdk/e2b/utils.py
```diff
@@ -0,0 +1,46 @@
...
+    paginator implements `next_items` to do the actual fetching for its
+    endpoint, so any model can expose pagination by subclassing this without
+    reimplementing the bookkeeping.
+    """
+
+    def __init__(
+        self,
+        limit: Optional[int] = None,
+        next_token: Optional[str] = None,
+        **opts: Unpack[ApiParams],
```
mishushakov: you can use generics here

packages/js-sdk/src/utils.ts
```diff
@@ -67,6 +67,82 @@ export function timeoutToSeconds(timeout: number): number {
...
+ * @example
+ * ```ts
+ * const paginator = Sandbox.list()
+ * while (paginator.hasNext) {
+ *   const items = await paginator.nextItems()
+ *   console.log(items)
+ * }
+ * ```
+ */
+export abstract class Paginator<T, O = unknown> {
```
mishushakov: I think O should have constraint extends ConnectionConfig

packages/js-sdk/src/index.ts
huv1k: Done — removed the package-root export in a7491c0. `Paginator` is still `export`ed from `./utils` for internal use, so the exported `SandboxPaginator`/`SnapshotPaginator` subclasses still type-resolve.

packages/python-sdk/e2b/utils.py
huv1k: Done — `PaginatorBase` is now `Generic[T, OptsT]`: `T` for the item type and `OptsT` (bound to `ApiParams`) for the connection options, so `**opts` is generic too. The model bases bind them as `PaginatorBase[SandboxInfo, ApiParams]` / `PaginatorBase[SnapshotInfo, ApiParams]`.

packages/js-sdk/src/utils.ts
huv1k: Done — constrained to `O extends ConnectionOpts = ConnectionOpts` (via a type-only import to avoid a runtime import cycle). Note I used `ConnectionOpts` rather than `ConnectionConfig`: `O` is the per-call *options* type (e.g. `SandboxApiOpts`), not the `ConnectionConfig` instance built from them. Let me know if you meant the config object instead.

mishushakov:
@huv1k am thinking maybe moving the paginator to connection config instead of utils? since it depends on it anyways for types

huv1k:
> @huv1k am thinking maybe moving the paginator to connection config instead of utils? since it depends on it anyways for types

I think this is a weird place. I would personally move it to a separate file 😅

mishushakov:
separate file works too!

mishushakov (approved):

PR #1492, opened 2026-06-26
ci: trigger docs SDK reference sync after SDK release

mishushakov:
Adds a `trigger-docs-sync` job to the release workflow that dispatches an `sdk-release` repository event to `e2b-dev/docs`, so the docs SDK reference regenerates automatically on release. It mints a scoped GitHub App token for the docs repo and dispatches with `sdk=all` / `version=latest`. The job runs only when `publish` succeeds and at least one of the JS or Python SDKs was released, so a CLI-only release (or skipped CLI) won't trigger it.


ValentaTomas:

PR #1493, opened 2026-06-26
fix(cli): resolve template alias for migrated build scripts

### LINKED ISSUE #1478: [Bug]: `e2b template migrate` writes `template_id` into the `name` slot of `build_prod.py`, causing 409 alias collision on rebuild
### Sandbox ID or Build ID

### Environment

- Python: 3.12.13
- e2b (Python SDK): 2.26.0

### Timestamp of the issue

2026-06-24 04:16 UTC

### Frequency

Happens every time (when migrating an existing template)

### Expected behavior

After `e2b template migrate -l python-async` is applied to an existing template's `e2b.toml`, the generated `build_prod.py` should be runnable and rebuild the existing template in place (cache-hit when the Dockerfile content is unchanged).

### Actual behavior

`python build_prod.py` fails with:

```
e2b.exceptions.BuildException: 409: Alias 'abc1234567890xyz' is already used
```

(`abc1234567890xyz` above is a placeholder; the actual `template_id` from our environment has been redacted.)

The generated script writes the source template's `template_id` (internal opaque ID, e.g. `abc1234567890xyz`) into the second positional argument of `AsyncTemplate.build`. Per the SDK, that argument is the **template name** (human-readable alias), not the `template_id`. Passing the existing ID as a new alias collides with itself.

The `dev` variant happens to succeed because the generated `name` is `<template_id>-dev`, which is a fresh unused alias — so the bug only surfaces when the user actually tries to push to prod.

### Issue reproduction

1. Have an existing E2B template — for example, alias `my-template`, template_id `abc1234567890xyz`.
2. In a directory with a complete v1 `e2b.toml`:
   ```toml
   team_id = "..."
   template_id = "abc1234567890xyz"
   dockerfile = "e2b.Dockerfile"
   ```
   and the corresponding `e2b.Dockerfile`, run:
   ```
   ```
3. Migration succeeds; the generated `build_prod.py` reads:
   ```python
   await AsyncTemplate.build(
       template,
       "abc1234567890xyz",   # the template_id
       on_build_logs=default_build_logger(),
   )
   ```
4. Run `python build_prod.py` (after working around the relative-import issue from #1477).
5. Observe: `e2b.exceptions.BuildException: 409: Alias 'abc1234567890xyz' is already used`.

### Additional context

**Why `template_id` is the wrong value here**

`AsyncTemplate.build` docstring:

```
:param name: Template name in 'name' or 'name:tag' format
:param alias: (Deprecated) Alias name for the template. Use name instead.
```

`name` is the human-readable alias (e.g. `my-template`), not the internal `template_id`. Passing the existing `template_id` as `name` makes the API try to create a new template with that string as a new alias, which already exists → 409.

**The fix that actually rebuilds the existing template**

Replacing the argument with the existing alias works as expected:

```python
await AsyncTemplate.build(
    template,
    "my-template",   # the existing alias
    cpu_count=2,
    memory_mb=1024,
    on_build_logs=default_build_logger(),
)
```

Verified locally end-to-end:
- `python build_prod.py` succeeded; the build log shows `Template created with ID: <existing-template_id>` (i.e. it routed to the existing template, not a new one), and every layer hit cache (`CACHED [base] FROM e2bdev/base`, etc.) because the Dockerfile content is unchanged.
- After the rebuild, `Sandbox.create("my-template", ...)` spawns successfully and executes commands — confirming the rebuilt template is intact.

**Where in the codebase**

Likely the migration code substitutes `template_id` from the parsed `e2b.toml` into the `name` slot of the generated build script (e.g. `packages/cli/src/templates/python-build-async.hbs` consuming a `template_id` field from the migration context built in `packages/cli/src/commands/template/migrate.ts`). The same pattern would affect `python-build-sync.hbs` and `typescript-build.hbs`.

**Possible fix (deferring to maintainers for direction)**

The migration should write the human-readable **alias** into the generated `name`, not the opaque `template_id`. Sources for the alias could be:

- Resolve via the templates API using the `template_id` (since `e2b template list` shows the alias is stored server-side).
- Accept an `--alias` flag on `e2b template migrate` for users whose `e2b.toml` doesn't carry the alias inline.
- Prompt interactively when ambiguous.

Happy to send a PR once you confirm the direction.

### Related

- #1477 — a separate `template migrate` bug where generated Python scripts use a relative import (`from .template import template`) but the printed instructions tell users to run them as single files (`python build_dev.py`), producing `ImportError: attempted relative import with no known parent package`.

syf2211:
## Summary

When `e2b template migrate` runs against an existing `e2b.toml` that only has `template_id` (no `template_name`), the generated `build_prod.py` / `build.prod.ts` previously wrote the opaque template ID into the `Template.build` name argument. That causes a 409 alias collision when rebuilding the existing template.

This change resolves the human-readable alias before generating build scripts.

Fixes #1478

## Motivation

`Template.build` expects a template alias/name, not the internal `template_id`. Migrating legacy configs that only stored the ID therefore produced scripts that could not rebuild the template in place.

Related: #1477 (separate import fix in #1487).

## Changes

- Add `resolveTemplateBuildName()` to resolve build script names in this order:
  1. `template_name` from `e2b.toml`
  2. `--alias` CLI flag
  3. `GET /templates/{templateID}` (first sorted name/alias)
  4. Fallback to `template_id` with a targeted warning for opaque IDs
- Wire resolution into `template migrate`
- Add unit tests for resolution paths
- Add changeset for `@e2b/cli`

## Tests

- `pnpm typecheck` — pass
- `pnpm lint` — pass
- `pnpm test tests/commands/template/resolveBuildName.test.ts` — 6/6 pass
- `pnpm test tests/commands/template/migrate.test.ts` — 24/24 pass

## Notes

- When unauthenticated and no `--alias` is provided, behavior falls back to the previous `template_id` value (with a warning for opaque-looking IDs). Logged-in users get the correct alias automatically.
- Does not change generated file contents for configs that already include `template_name`.

### DIFF
--- a/.changeset/fix-migrate-template-alias.md
+++ b/.changeset/fix-migrate-template-alias.md
@@ -1,0 +1,7 @@
+---
+"@e2b/cli": patch
+---
+
+fix(cli): resolve template alias for migrated build scripts when only template_id is set
+
+`e2b template migrate` now looks up the human-readable template alias from the API (or accepts `--alias`) instead of writing the opaque `template_id` into generated `build_prod.py`, which caused 409 alias collisions when rebuilding existing templates.
--- a/packages/cli/src/commands/template/migrate.ts
+++ b/packages/cli/src/commands/template/migrate.ts
@@ -14,6 +14,7 @@ import {
   Language,
   languageDisplay,
 } from './generators'
+import { resolveTemplateBuildName } from './resolveBuildName'
 
 /**
  * Migrate Dockerfile to a specific target language using SDK
@@ -22,7 +23,8 @@ async function migrateToLanguage(
   root: string,
   config: E2BConfig,
   dockerfileContent: string,
-  language: Language
+  language: Language,
+  opts?: { alias?: string }
 ): Promise<void> {
   // Initialize template with file context
   const template = Template({
@@ -66,7 +68,9 @@ async function migrateToLanguage(
     parsedTemplate = baseTemplate.setReadyCmd(config.ready_cmd)
   }
 
-  const name = config.template_name || config.template_id
+  const name = await resolveTemplateBuildName(config, {
+    alias: opts?.alias,
+  })
   if (!name) {
     throw new Error('Template name or ID is required')
   }
@@ -94,6 +98,10 @@ export const migrateCommand = new commander.Command('migrate')
   )
   .addOption(configOption)
   .option(
+    '--alias <name>',
+    'template alias to use in generated build scripts when e2b.toml has no template_name'
+  )
+  .option(
     '-l, --language <language>',
     `specify target language: ${Object.values(Language).join(', ')}`,
     (value) => {
@@ -114,6 +122,7 @@ export const migrateCommand = new commander.Command('migrate')
       config?: string
       path?: string
       language?: Language
+      alias?: string
     }) => {
       let success = false
       try {
@@ -173,7 +182,9 @@ export const migrateCommand = new commander.Command('migrate')
         }
 
         // Perform migration
-        await migrateToLanguage(root, config, dockerfileContent, language)
+        await migrateToLanguage(root, config, dockerfileContent, language, {
+          alias: opts.alias,
+        })
 
         // Rename old files to .old extensions
         const oldFilesRenamed: { oldPath: string; newPath: string }[] = []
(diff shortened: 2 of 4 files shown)


mishushakov:
closing in favor of https://github.com/e2b-dev/E2B/pull/1494

PR #1495, opened 2026-06-27
fix(sdks): robust http2 transport reset retries

jashwanth-reddy-g:


mishushakov:
Hey, thanks for trying to contribute.
We will be implementing the retry mechanism ourselves in an upcoming version of E2B SDK.

PR #1499, opened 2026-06-29
kill sandbox on MCP gateway startup failure to prevent orphaned resources

### LINKED ISSUE #1498: MCP gateway startup failure leaves orphaned sandbox running
Body:

When Sandbox.create() is called with an mcp config and the MCP gateway fails to start (non-zero exit code or exception from commands.run), the already-created sandbox is never killed. This leaves orphaned sandboxes consuming resources until they time out.

Affected: Python SDK (async + sync) and JS/TS SDK.

Steps to reproduce:

Call Sandbox.create(mcp={"invalid_server": {}}) with a config that causes mcp-gateway to fail
Observe the exception is raised
The sandbox remains running (visible via Sandbox.list())
Expected: The sandbox should be killed before the error is propagated.

AdaAibaby:
fix https://github.com/e2b-dev/E2B/issues/1498

#### Problem
When Sandbox.create() is called with an mcp configuration, the sandbox is created first, then the MCP gateway is started inside it. If the gateway startup fails (non-zero exit code or exception), the error is raised but the sandbox is never cleaned up, leaving an orphaned sandbox running and consuming resources until it times out.
This issue impacts all three SDK implementations:
Python async: AsyncSandbox.create
Python sync: Sandbox.create
JS/TS: Sandbox.create
#### Fix
Wrap MCP gateway startup logic with try/catch:
On any gateway startup failure, run best-effort sandbox.kill() before re-raising the original error.
Catch errors from kill() separately to guarantee the original failure error is always propagated to the caller.
#### Code Changes
main.py: Add async cleanup with await sandbox.kill()
main.py: Add sync cleanup with sandbox.kill()
index.ts: Add cleanup logic await sandbox.kill().catch(() => {})
test_mcp_gateway_cleanup.py: 14 unit tests (Python async + sync)
mcpGatewayCleanup.test.ts: 7 unit tests (TS)
#### Test Coverage
Total 21 new unit tests, all scenarios verified:
image

### DIFF
--- a/packages/js-sdk/src/sandbox/index.ts
+++ b/packages/js-sdk/src/sandbox/index.ts
@@ -282,17 +282,17 @@ export class Sandbox extends SandboxApi {
     const { template, sandboxOpts } =
       typeof templateOrOpts === 'string'
         ? {
-            template: templateOrOpts,
-            sandboxOpts: opts,
-          }
+          template: templateOrOpts,
+          sandboxOpts: opts,
+        }
         : {
-            template:
-              templateOrOpts?.template ??
-              (templateOrOpts?.mcp
-                ? this.defaultMcpTemplate
-                : this.defaultTemplate),
-            sandboxOpts: templateOrOpts,
-          }
+          template:
+            templateOrOpts?.template ??
+            (templateOrOpts?.mcp
+              ? this.defaultMcpTemplate
+              : this.defaultTemplate),
+          sandboxOpts: templateOrOpts,
+        }
 
     const config = new ConnectionConfig(sandboxOpts)
     if (config.debug) {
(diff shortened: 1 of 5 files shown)


mishushakov:
Hey there, thanks for the contribution!
Could you please update the tests so they are integration tests instead of mock?

hxaxd:
Hi @AdaAibaby, I noticed this PR has been waiting on integration-test coverage since June 29.

I'd be happy to take over the remaining work, replace the mocked tests with integration tests, and open a superseding PR while preserving attribution for your original implementation.

@mishushakov, would that approach work for you if the original author isn't planning to continue?

mishushakov:
@hxaxd yes, but we require individual CLAs, so your contribution cannot be based on this author's work, it has to be implemented from scratch.

hxaxd:
@mishushakov Thanks for confirming. I opened #1547 with an independent implementation and real integration tests.

mishushakov:
Closing this one in favor of https://github.com/e2b-dev/E2B/pull/1547

PR #1500, opened 2026-06-29
fix(python-sdk): correct Sandbox.list() docstring (also lists paused)

tristone13th:
Sandbox.list() currently documents itself as "List all running sandboxes", but the server (api/internal/handlers/sandboxes_list.go) defaults to returning both `running` and `paused` sandboxes when no state filter is provided:

    // If no state is provided we want to return both running and paused sandboxes
    states := make([]api.SandboxState, 0)
    if params.State == nil {
        states = append(states, api.Running, api.Paused)
    } else {
        states = append(states, *params.State...)
    }

The docstring change brings the SDK documentation in line with the actual backend behavior, and points users at SandboxQuery(state=[...]) for narrowing the result set. Applied to both sync and async variants for consistency.

No behavior change.

### DIFF
--- a/.changeset/sandbox-list-docstring.md
+++ b/.changeset/sandbox-list-docstring.md
@@ -1,0 +1,6 @@
+---
+"@e2b/python-sdk": patch
+"e2b": patch
+---
+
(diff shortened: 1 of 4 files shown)


tristone13th:
@cla-bot check

mishushakov:
I think the description is not very correct still, it doesn't actually return the list, but it returns a Paginator that will then return a page on next_items()

tristone13th:
> I think the description is not very correct still, it doesn't actually return the list, but it returns a Paginator that will then return a page on next_items()

You are right, those descrptions have been updated.

mishushakov (approved):

PR #1502, opened 2026-06-29
fix(sdk): support sending stdin through the PTY command handle

mishushakov:
## Summary

Calling `send_stdin`/`sendStdin` on the handle returned by `pty.create()`/`pty.connect()` previously raised _"Sending stdin is not supported for this command handle"_ because the PTY path never wired up the handle's stdin callback — unlike regular command handles. This wires it up across the JS and Python (sync + async) SDKs, makes PTY input methods also accept `str` (encoded as UTF-8), and in JS adds `pty.sendStdin()` as the canonical method (mirroring `commands.sendStdin()`) with `pty.sendInput()` kept as a deprecated alias.

## Usage

```python
terminal = sandbox.pty.create(size=PtySize(rows=24, cols=80))
terminal.send_stdin(b"whoami\n")   # now works on the handle
terminal.send_stdin("exit\n")       # str is accepted too
terminal.wait(on_pty=lambda data: print(data.decode(), end=""))
```

```ts
const terminal = await sandbox.pty.create({ cols: 80, rows: 24, onData: () => {} })
await terminal.sendStdin('whoami\n')        // on the handle
await sandbox.pty.sendStdin(terminal.pid, 'exit\n')  // new canonical module method
```

## Tests

Added handle-level and `str`-input tests for sync/async Python and JS (plus a JS test for the deprecated `sendInput` alias); all pass against a live sandbox.


ValentaTomas:
My reluctance here is that we don't feed the input to the process stdin, so I didn't want to name it stdin. Also the output is not called stdout (the same case), so the symmetry is already broken.

Might make sense to have both for compatibility, but does not feel very clean then.

mishushakov:
@ValentaTomas I would settle on one name, in that case as far as I understand, sendInput would be the correct one?

PR #1503, opened 2026-06-29
fix(python-sdk): percent-encode git credentials in clone URL

Osamaali313:
## Problem

`with_credentials` (`packages/python-sdk/e2b/sandbox/_git/auth.py`) embeds the username and password into the Git URL **raw**:

```python
netloc = f"{username}:{password}@{parsed.netloc}"
```

When the username/token contains URL-significant characters (`@`, `/`, `:`, space, …) the URL is malformed. A password `p@ss` produces `https://user:p@ss@github.com/org/repo.git`, which git parses with host `ss@github.com` and rejects:

```
fatal: unable to access '.../': URL rejected: Bad hostname
```

The JS SDK builds the URL via the `URL` username/password setters, which **do** percent-encode — so the two SDKs disagree and the Python one breaks on real-world tokens.

## Reproduction

`with_credentials("https://github.com/o/r.git", "user", "p@ss")` → `https://user:p@ss@github.com/o/r.git` (before) vs `https://user:p%40ss@github.com/o/r.git` (after, matching JS).

Reachable via `Sandbox.git.clone(url, username=..., password=...)` and `git.dangerously_authenticate(...)`.

## Fix

Percent-encode both credentials with `urllib.parse.quote(..., safe='')`. `strip_credentials` still round-trips the encoded URL back to the clean one.

## Tests

Adds `packages/python-sdk/tests/shared/git/test_auth.py` (encoding of special chars, strip round-trip, no-credentials passthrough). Changeset included.

### DIFF
--- a/.changeset/git-credentials-url-encode.md
+++ b/.changeset/git-credentials-url-encode.md
@@ -1,0 +1,5 @@
+---
+"@e2b/python-sdk": patch
+---
+
+Percent-encode the username and password when embedding HTTP(S) Git credentials into a clone URL. Previously they were inserted raw, so a username or token containing URL-significant characters (`@`, `/`, `:`, spaces, etc.) produced a malformed URL that git could not use. This now matches the JS SDK, which already encodes the credentials.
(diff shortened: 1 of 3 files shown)


Osamaali313:
@cla-bot check

mishushakov:
Hey there, thanks for your contribution, but we're not currently considering any changes to our git integration.

PR #1506, opened 2026-06-30
feat(js-sdk): expose E2B_ENVD_CONNECTIONS env var for file REST API connection limit

AdaAibaby:
The envd REST file API (files.read/files.write) had its undici connection\npool hardcoded to 10 connections with no way to override it, while the\nRPC transport (commands) allowed 200 connections configurable via\nE2B_ENVD_RPC_CONNECTIONS. Under high concurrency this 20x gap caused\nsocket write contention and crashes on the file API path.\n\n- Add getEnvdConnectionLimit() reading E2B_ENVD_CONNECTIONS (default 10)\n- Pass connectionLimit into createEnvdFetch() instead of relying on the\n  hardcoded fallback\n- Add tests for the new env var (valid + malformed values)"

/cc @mishushakov please review.

### DIFF
--- a/.changeset/envd-connection-limit-env-var.md
+++ b/.changeset/envd-connection-limit-env-var.md
@@ -1,0 +1,5 @@
+---
+"e2b": patch
+---
+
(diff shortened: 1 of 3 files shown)


mishushakov:
@matthewlouisbrockman feel free to take a look

mishushakov:
Thanks for the contribution, we're closing this for now as we will be changing transports in an upcoming changes and also will start using http2 down the road to optimize connections

PR #1507, opened 2026-06-30
fix(python-sdk): percent-encode git credentials in with_credentials

anxkhn:
### What

`Sandbox.git.clone` and `Sandbox.git.pull` embed HTTP(S) credentials by passing them
through `with_credentials` (`packages/python-sdk/e2b/sandbox/_git/auth.py`), which built
the URL netloc by raw string interpolation:

```python
netloc = f"{username}:{password}@{parsed.netloc}"
```

A username, password, or token that contains URL-reserved characters (`@ : / # ?`) is
spliced straight into the netloc, so the resulting URL is corrupted and git auth fails.
Common GitHub PATs are alphanumeric and unaffected, but GitLab/Bitbucket app passwords
and many tokens do contain reserved characters.

The JavaScript SDK does not have this problem: `withCredentials`
(`packages/js-sdk/src/sandbox/git/utils.ts`) assigns `parsed.username` / `parsed.password`
through the WHATWG `URL` setters, which percent-encode automatically. So the two SDKs
disagree on identical input.

| password     | Python (before)                 | JS / Python (after)                   |
|--------------|---------------------------------|---------------------------------------|
| `p@ss/w:rd`  | `user:p@ss/w:rd@github.com/...` | `user:p%40ss%2Fw%3Ard@github.com/...` |
| `x#y?z`      | `user:x#y?z@github.com/...`     | `user:x%23y%3Fz@github.com/...`       |
| `ghp_AbC123` | `user:ghp_AbC123@...` (ok)      | `user:ghp_AbC123@...` (unchanged)     |

### Fix

Percent-encode both parts with `quote(safe='')` before building the netloc, matching the
JS SDK byte-for-byte. Alphanumeric tokens are unchanged.

```python
netloc = f"{quote(username, safe='')}:{quote(password, safe='')}@{parsed.netloc}"
```

This is the only behavior change. Sync and async share this single helper, so both clone
and pull are fixed by the one-line edit.

### Usage example (now works)

```python
from e2b import Sandbox

sbx = Sandbox()
# token with reserved characters no longer corrupts the URL
sbx.git.clone(
    "https://github.com/org/private.git",
    username="user",
    password="p@ss/w:rd",
)
```

### Tests

- `packages/python-sdk/tests/shared/git/test_auth.py` (new, 5 offline cases): reserved
  chars encode, alnum token unchanged, no-cred passthrough, partial creds raise,
  non-http(s) raises. Runs without a sandbox (imports `e2b.sandbox._git`).
- `packages/js-sdk/tests/sandbox/git/validation.test.ts` (+1 case): locks the JS encoded
  output so the two SDKs stay in parity.

Changeset: `@e2b/python-sdk` patch.

### DIFF
--- a/.changeset/python-git-credentials-encode.md
+++ b/.changeset/python-git-credentials-encode.md
@@ -1,0 +1,12 @@
+---
+"@e2b/python-sdk": patch
+---
+
+fix(python-sdk): percent-encode git credentials so reserved characters are preserved
+
+`Sandbox.git.clone`/`push` embed credentials into HTTP(S) URLs with
+`with_credentials`, which previously interpolated the username and password
+straight into the URL. A username or token containing URL-reserved characters
+(`@ : / # ?`) produced a corrupted URL and broke git auth. The credentials are
+now percent-encoded, matching the JavaScript SDK, which already encodes via the
+WHATWG `URL` setters. Alphanumeric tokens are unaffected.
--- a/packages/js-sdk/tests/sandbox/git/validation.test.ts
+++ b/packages/js-sdk/tests/sandbox/git/validation.test.ts
@@ -3,6 +3,7 @@ import { test, expect } from 'vitest'
 import { Git } from '../../../src/sandbox/git'
 import type { Commands } from '../../../src/sandbox/commands'
 import { InvalidArgumentError } from '../../../src/errors'
+import { withCredentials } from '../../../src/sandbox/git/utils'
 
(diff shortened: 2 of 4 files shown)


anxkhn:
@cla-bot check

anxkhn:
Pushed a revision addressing the latest review/CI feedback.

mishushakov:
We're holding off contributions to the Git integration for a bit. Thanks!

PR #1517, opened 2026-06-30
feat(cli): track CLI usage via ConnectionConfig integration header

mishushakov:
Tags all CLI traffic to the API with an `e2b-cli/<version>` identifier in the `User-Agent` header so CLI usage and version distribution can be tracked server-side, with no SDK changes and no new dependencies.

The first commit tagged the shared `ConnectionConfig`, but the auth, sandbox-lifecycle, and `template build` commands build their own connection from per-call options and were still sending untagged requests (flagged by Cursor Bugbot + Codex). This adds a `withCliIntegration` helper (and a `CLI_INTEGRATION` constant) in `packages/cli/src/api.ts` and threads it through every remaining call site: `auth login`/`configure`, `sandbox create`/`connect`/`resume`/`kill`/`pause`/`list`/`info`/`metrics`/`exec`, and `template build`.

`integration` is accepted at runtime by every `ConnectionConfig` the SDK builds internally, but isn't part of the narrower option types the `Sandbox.*`/`Template.*` static methods declare — the helper centralizes that single cast so the call sites stay clean. Direct `new ConnectionConfig(...)` sites (auth) pass `integration` inline since the constructor's type already declares it.

## Usage

```console
$ e2b sandbox create my-template
$ e2b sandbox list
$ e2b auth login
$ e2b template build
# Every request now sends:  User-Agent: e2b-js-sdk/<sdk-version> e2b-cli/<cli-version>
```

Server-side, filter API logs by `User-Agent` containing `e2b-cli/` to measure CLI vs SDK traffic and version distribution, broken down by team/user via the existing auth headers.

## Tests

- `api_integration_header.test.ts` covers `CLI_INTEGRATION`, the module-level config, and the `withCliIntegration` helper.
- `create_lifecycle.test.ts` now asserts `Sandbox.create` is called with the `integration` field, locking in the threading as a regression guard.

### DIFF
--- a/.changeset/cli-integration-user-agent.md
+++ b/.changeset/cli-integration-user-agent.md
@@ -2,4 +2,4 @@
 "@e2b/cli": patch
 ---
 
-Identify CLI traffic to the API by setting the `integration` field on the CLI's `ConnectionConfig`. Every request the CLI makes now carries `e2b-cli/<version>` in the `User-Agent` header, so CLI usage and version distribution can be tracked server-side. No user-facing behavior changes.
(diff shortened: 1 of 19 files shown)


mishushakov:
waiting until the backend team enables integration header on the backend

mishushakov:
this will be minimal shape using global config

PR #1525, opened 2026-07-03
feat(cli): attribute CLI traffic via the e2b-cli integration tag

mishushakov:
Stacked on #1524.

Sets the CLI integration string introduced there: `ConnectionConfig.setIntegration('e2b-cli/<version>')` runs at the top of `src/api.ts`, before the shared connection config is constructed at import time, so every CLI request carries the tag in the `User-Agent` header:

```
User-Agent: e2b-js-sdk/2.31.0 e2b-cli/2.13.0
```

No dependency changes needed: the CLI typechecks and bundles the **workspace** js-sdk source (`packages/cli/tsconfig.json` maps `e2b` → `../js-sdk/src`, and esbuild follows the same mapping when bundling via `noExternal`), so the setter is available as soon as #1524 merges — no published `e2b` release required.

### ⚠️ Blocked on a backend User-Agent parsing fix

The API currently parses the SDK version from the `User-Agent` by taking **everything after** `e2b-js-sdk/`, so any integration tag appended after the SDK token breaks template builds (verified against production):

```
❌ Template build failed.
Error: 400: Error when parsing user agent: parsing JS SDK version: invalid current version format: v2.31.0 e2b-cli/2.13.0
```

This affects every integration using #1524's setter, not just the CLI. The parser needs to be fixed on the backend (stop at the first whitespace after the SDK token) before this — or any integration tagging — ships. Prepending the tag instead (`e2b-cli/2.13.0 e2b-js-sdk/2.31.0`) passes against today's backend, if a client-side ordering workaround is preferred in #1524.

### Merge order

- [ ] backend User-Agent parser handles extra product tokens after `e2b-js-sdk/<version>`
- [ ] #1524 merges
- [ ] this merges

### Verified

- New `packages/cli/tests/attribution.test.ts` covers the import-time shared config and configs constructed later at runtime (auth commands, Sandbox calls).
- Full CLI suite (96 tests) passes locally.
- Ran the built binary against a local HTTP server and observed `User-Agent: e2b-js-sdk/2.31.0 e2b-cli/2.13.0`, confirming the call survives esbuild bundling (`sideEffects: false` makes a bare side-effect import unsafe, hence placement in `api.ts`) and that the bundle carries the workspace SDK regardless of the npm `e2b` version installed.


mishushakov:
will reopen a fresh one, since this one drifted too far

PR #1526, opened 2026-07-03
test: skip live volume tests unless ENABLE_VOLUME_TESTS is set

mishushakov:
Live volume tests create real volumes against the API; this gates them behind an `ENABLE_VOLUME_TESTS` env var so they skip by default. In the JS SDK, the `volumeTest` fixture is chained with `.skipIf(process.env.ENABLE_VOLUME_TESTS === undefined)`, skipping all of `tests/volume/file.test.ts`. In the Python SDK, the `volume` and `async_volume` fixtures call `pytest.skip` when the env var is unset, gating `tests/{sync/volume_sync,async/volume_async}/test_file.py`. Mocked and unit volume tests (msw-based `volume.test.ts`, `test_volume.py`, `test_volume_content.py`, `test_volume_client.py`, `test_volume_connection_config.py`) still run unconditionally. To run the live tests: `ENABLE_VOLUME_TESTS=1 pnpm run test` or `ENABLE_VOLUME_TESTS=1 poetry run pytest`.

### DIFF
--- a/packages/js-sdk/tests/setup.ts
+++ b/packages/js-sdk/tests/setup.ts
@@ -115,27 +115,29 @@ export const buildTemplateTest = base.extend<BuildTemplateFixture>({
   ],
 })
 
-export const volumeTest = base.extend<VolumeFixture>({
-  volume: [
-    // eslint-disable-next-line no-empty-pattern
-    async ({}, use) => {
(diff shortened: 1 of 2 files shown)


matthewlouisbrockman (approved):
lgtm

PR #1541, opened 2026-07-10
fix(python-sdk): strip uppercase AS stage alias from base image in from_dockerfile

anxkhn:
## What

`Template().from_dockerfile(...)` only stripped a lowercase `" as "` alias from the
`FROM` line, so `FROM python:3.11 AS builder` produced the base image
`"python:3.11 AS builder"` instead of `"python:3.11"`.

`dockerfile-parse` returns the `FROM` argument verbatim. The Python SDK's guard was
case-insensitive, but its split only matched lowercase `as`. The JavaScript SDK
already handles both forms correctly.

## The fix

Split the stage alias case-insensitively and accept any whitespace around `AS`.
This handles uppercase, lowercase, and mixed-case aliases while preserving the
existing behavior for Dockerfiles without an alias.

## Usage example

```python
from e2b import Template, TemplateBase

template = Template().from_dockerfile(
    "FROM python:3.11 AS builder\nRUN echo hi"
)

print(template._template._base_image)
# Before: python:3.11 AS builder
# After:  python:3.11

print(TemplateBase.to_dockerfile(template).splitlines()[0])
# Before: FROM python:3.11 AS builder
# After:  FROM python:3.11
```

## Tests

Added offline regression coverage to the sync and async Python suites for both
uppercase and lowercase aliases.

```text
.venv/bin/python -m pytest \
  tests/sync/template_sync/methods/test_from_dockerfile.py \
  tests/async/template_async/methods/test_from_dockerfile.py -q
# 16 passed
```

## Changeset

Added a patch changeset for `@e2b/python-sdk`.

### DIFF
--- a/.changeset/python-from-dockerfile-uppercase-as.md
+++ b/.changeset/python-from-dockerfile-uppercase-as.md
@@ -1,0 +1,11 @@
+---
+"@e2b/python-sdk": patch
+---
+
+fix(python-sdk): strip uppercase AS stage alias from base image in from_dockerfile
+
+`from_dockerfile` now removes the stage alias from a `FROM <image> AS <name>`
+line case-insensitively, matching the JavaScript SDK. The Docker `AS` keyword is
+case-insensitive, but the parser only matched a lowercase `as`, so a canonical
+`FROM python:3.11 AS builder` leaked the alias and produced the base image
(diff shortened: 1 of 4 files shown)


mishushakov:
We don't support multi-stage builds so no need for this

PR #1543, opened 2026-07-13
fix(templates): upgrade base Dockerfile Node from 20.9.0 to 20.20.2

dongjiang1989:
## Summary

The `templates/base/e2b.Dockerfile` pinned Node **20.9.0**, which is below the minimum version required by the JS SDK `engines` field (`>=20.18.1`). Users building sandbox templates with the SDK installed could hit version mismatches or fail engine checks at install time.

### Changes

- **Node 20.9.0 → 20.20.2** — latest 20.x release, satisfies the SDK engines constraint
- **GPG keys updated** — removed 5 expired signing keys and added the 2 new keys used to sign Node v20.20.2. The key list was sourced from the last `docker-node/20/bookworm/Dockerfile` before the Node 20 EOL directory was removed from the official repo.
- **Added `gpg --fingerprint` verification** — matches the current official `docker-node` pattern, preventing silent key import failures (previously, a key failing to import would only log a warning and proceed; now the fingerprint is printed and the build fails fast if any key is missing)
- **Yarn 1.22.19 → 1.22.22**

### Why keep Node 20.x instead of jumping to 22?

The SDK `engines` field is `>=20.18.1 <21 || >=22`, so Node 20.x is still a supported runtime. Staying on 20.x minimises behavioural risk for existing sandbox templates — users who pinned the base template image are not forced into a major-version migration. Node 20.x will continue receiving security patches through its maintenance LTS window.

## Test plan

- [x] Build `templates/base/e2b.Dockerfile` with Docker and verify `node --version` prints `v20.20.2`
- [x] Verify `yarn --version` prints `1.22.22`
- [x] Confirm GPG verification passes during image build (no keyserver errors)
- [x] Run `npm install e2b@latest` inside a container from the new image and confirm no engine warnings

### DIFF
--- a/templates/base/e2b.Dockerfile
+++ b/templates/base/e2b.Dockerfile
@@ -9,7 +9,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y \
 
 RUN groupadd -r node && useradd -r -g node -s /bin/bash -m node
 
-ENV NODE_VERSION=20.9.0
+ENV NODE_VERSION=20.20.2
 
 RUN ARCH= && dpkgArch="$(dpkg --print-architecture)" \
   && case "${dpkgArch##*-}" in \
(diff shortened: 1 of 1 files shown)


mishushakov:
We will manage it ourselves, thanks.

PR #1546, opened 2026-07-13
feat(sdk): add httpsPorts network config for HTTPS sandbox backends

mishushakov:
Adds `httpsPorts` (JS) / `https_ports` (Python, sync and async) to the sandbox network config (SDK-235) so ports whose in-sandbox service serves TLS itself get their public URLs proxied to the sandbox over HTTPS. This is not TLS passthrough — traffic is still terminated at the E2B proxy and re-encrypted on the hop to the sandbox, and the backend certificate is not verified, so self-signed certificates work. The configured ports are also returned in the sandbox info network config. The OpenAPI spec field already landed on `main` via the infra spec sync (e2b-dev/infra#3280, now deployed), so this PR is the SDK wiring, live tests (all passing), and a changeset.

### Usage

```ts
const sandbox = await Sandbox.create({ network: { httpsPorts: [8443] } })
// https://<host> now proxies to the HTTPS server listening on 8443 in the sandbox
const url = `https://${sandbox.getHost(8443)}`
```

```python
sandbox = Sandbox.create(network={"https_ports": [8443]})
url = f"https://{sandbox.get_host(8443)}"
```

### DIFF
--- a/.changeset/https-backend-ports.md
+++ b/.changeset/https-backend-ports.md
@@ -1,0 +1,6 @@
+---
+"e2b": minor
+"@e2b/python-sdk": minor
+---
+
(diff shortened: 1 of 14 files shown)


matthewlouisbrockman (approved):
looks good

PR #1547, opened 2026-07-13
fix(sdk): clean up sandbox when MCP gateway startup fails

### LINKED ISSUE #1498: MCP gateway startup failure leaves orphaned sandbox running

hxaxd:
## Problem

Fixes #1498.

`Sandbox.create` allocates a remote sandbox before starting `mcp-gateway`. If gateway startup returns a non-zero exit code or throws an exception, creation fails before the sandbox object is returned. As a result, the caller has no sandbox ID to clean up, and the orphaned sandbox continues consuming resources until it times out.

This state transition exists in synchronous Python, asynchronous Python, and JavaScript/TypeScript.

## Changes

- Add a rollback boundary around MCP gateway startup in all three SDK implementations.
- Attempt to kill the newly allocated sandbox when gateway startup fails.
- Keep cleanup best-effort so a failed or cancelled `kill` does not replace the original gateway error.
- Add real integration coverage for synchronous Python, asynchronous Python, and TypeScript.
- Add a patch changeset for `e2b` and `@e2b/python-sdk`.

## Usage Behavior

No API changes are required. Existing code continues to receive the original startup error, but a failed creation will normally no longer leave a sandbox behind.

## Validation

Integration tests have been added and verified successfully.

## Notes

PR #1499 is a friendly competing implementation for the same issue. The maintainer requested that its mocked tests be replaced with real integration tests, but the PR received no further updates for several days. The maintainer subsequently confirmed that an alternative contribution could be submitted as long as it was implemented independently from scratch, so this PR was prepared only after that confirmation.

### DIFF
--- a/.changeset/quiet-sandboxes-clean.md
+++ b/.changeset/quiet-sandboxes-clean.md
@@ -1,0 +1,6 @@
+---
+"e2b": patch
+"@e2b/python-sdk": patch
+---
+
+Kill newly created sandboxes when MCP gateway startup fails.
--- a/packages/js-sdk/src/sandbox/index.ts
+++ b/packages/js-sdk/src/sandbox/index.ts
@@ -320,17 +320,22 @@ export class Sandbox extends SandboxApi {
 
     if (sandboxOpts?.mcp) {
       sandbox.mcpToken = crypto.randomUUID()
-      const res = await sandbox.commands.run(
-        `mcp-gateway --config ${shellQuote(JSON.stringify(sandboxOpts.mcp))}`,
-        {
-          user: 'root',
-          envs: {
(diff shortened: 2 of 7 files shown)


hxaxd:
@cla-bot check

mishushakov (approved):

PR #1554, opened 2026-07-15
feat(sdk): add sandbox fork to JS and Python SDKs

mishushakov:
## Summary

Adds SDK support for the new `POST /sandboxes/{sandboxID}/fork` endpoint (e2b-dev/infra#3202): checkpoint a running sandbox in place (briefly paused, snapshotted with full memory state, and resumed — its ID and expiration stay untouched) and boot `count` new sandboxes from that snapshot.

- **spec**: adds `SandboxForkRequest` / `SandboxForkResult` schemas and the `/sandboxes/{sandboxID}/fork` path (mirroring the infra spec); JS and Python API clients regenerated via `make codegen`.
- **js-sdk**: `sandbox.fork(opts)` instance method and `Sandbox.fork(sandboxId, opts)` static method. Returns `Promise<Array<Sandbox | Error>>` — one entry per requested fork, each either a connected `Sandbox` instance or an `Error` describing why that fork failed to start (`Promise.allSettled`-style, matching the per-fork results of the API). Per-fork error codes go through the same code→class mapping as other API errors (extracted from `handleApiError` into `apiErrorFromCode`), so e.g. a per-fork 429 (sandbox limit) surfaces as `RateLimitError`. `SandboxForkOpts` extends the full `ConnectionOpts` (like `SandboxConnectOpts`), so `proxy`, `logger`, `apiUrl`, etc. work with fork-by-ID. `timeoutMs` defaults to 5 minutes like `create`/`connect`; `count` defaults to 1 and is validated client-side (`InvalidArgumentError` for `count < 1`); a whole-request 404 maps to `SandboxNotFoundError` (the source sandbox is the missing resource — same semantics as `pause`/`connect`/`setTimeout`), carrying the API error message when present; per-fork 404 error codes map to generic `NotFoundError` (the missing resource is fork-internal, e.g. the snapshot).
- **python-sdk**: `sandbox.fork(timeout=..., count=...)` / `Sandbox.fork(sandbox_id, ...)` and the `AsyncSandbox` equivalents (same `@class_method_variant` instance/static pattern as `connect`/`pause`), returning `List[Union[Sandbox, Exception]]`. Per-fork errors map through the shared `api_exception_from_code` (extracted from `handle_api_exception`). `timeout` is in seconds per Python SDK convention; an explicit `timeout=0` is preserved. Whole-request 404 raises `SandboxNotFoundException`; per-fork 404 codes map to generic `NotFoundException`.
- **changesets**: minor bumps for `e2b` and `@e2b/python-sdk`.

## Usage

JS:

```ts

const [fork1, fork2] = await sandbox.fork({ count: 2, timeoutMs: 60_000 })
if (fork1 instanceof Sandbox) {
  await fork1.commands.run('echo "hello from fork"')
}

// or by ID
const forks = await Sandbox.fork(sandbox.sandboxId, { count: 2 })
```

Python (sync / async):

```python
sandbox = Sandbox.create()

fork1, fork2 = sandbox.fork(count=2, timeout=60)
if isinstance(fork1, Sandbox):
    fork1.commands.run('echo "hello from fork"')

# or by ID
forks = Sandbox.fork(sandbox.sandbox_id, count=2)
```

```python
sandbox = await AsyncSandbox.create()
fork1, fork2 = await sandbox.fork(count=2)
```

## Notes

- The JS option is named `timeoutMs` (milliseconds) to match `SandboxOpts.timeoutMs` / `SandboxConnectOpts.timeoutMs`; the API receives seconds via `timeoutToSeconds` as elsewhere.
- Failed forks are returned as error **values** in the array rather than rejected promises, so a partial failure doesn't throw away the successful forks and there are no unhandled-rejection hazards. A per-fork error message includes the API error code only when the API returned one.

## Test plan

- [x] `pnpm run format`, `pnpm run lint`, `pnpm run typecheck` pass at the repo root (`ty` diagnostics identical to baseline)
- [x] Offline tests pass: `count < 1` → `InvalidArgumentError` / `InvalidArgumentException` in JS, Python sync, and Python async; `handleApiError` suite passes after the `apiErrorFromCode` extraction (plus a behavior-parity check of the Python `handle_api_exception` refactor)
- [ ] Integration tests (single fork with FS state inheritance + independence, multi-fork with unique IDs, fork-by-ID, fork of killed sandbox → `SandboxNotFoundError`) are written but currently fail against prod with 404 because the fork endpoint (e2b-dev/infra#3202) is not deployed yet — they should pass once it lands.


packages/js-sdk/src/sandbox/index.ts
```diff
@@ -383,6 +384,55 @@ export class Sandbox extends SandboxApi {
...
+      sandboxId,
+      opts?.timeoutMs ?? this.defaultSandboxTimeoutMs,
+      opts?.count ?? 1,
+      opts
+    )
+
+    return results.map((result) =>
+      result instanceof SandboxError
+        ? result
+        : (new this({ ...result, ...config }) as InstanceType<S>)
```
mishushakov, line 433: why do we pass ...config here ?

packages/js-sdk/src/sandbox/index.ts
```diff
@@ -413,6 +463,40 @@ export class Sandbox extends SandboxApi {
...
+   * const sandbox = await Sandbox.create()
+   *
+   * const [fork1, fork2] = await sandbox.fork({ count: 2 })
+   * if (fork1 instanceof Sandbox) {
+   *   await fork1.commands.run('echo "hello from fork"')
+   * }
+   * ```
+   */
+  async fork(opts?: SandboxForkOpts): Promise<Array<this | SandboxError>> {
+    const cls = this.constructor as typeof Sandbox
```
mishushakov, line 496: why this.constructor instead of just this.fork?

packages/js-sdk/src/sandbox/sandboxApi.ts
```diff
@@ -1203,6 +1243,62 @@ export class SandboxApi {
...
+      body: {
+        timeout: timeoutToSeconds(timeoutMs),
+        count,
+      },
+      signal: config.getSignal(opts?.requestTimeoutMs, opts?.signal),
+    })
+
+    if (res.error?.code === 404) {
+      throw new SandboxNotFoundError(`Sandbox ${sandboxId} not found`)
+    }
```
mishushakov: I think it should be a generic NotFoundError, not a Sandbox one?

packages/js-sdk/src/sandbox/sandboxApi.ts
```diff
@@ -1203,6 +1243,62 @@ export class SandboxApi {
...
+    const err = handleApiError(res)
+    if (err) {
+      throw err
+    }
+
+    return (res.data ?? []).map(
+      (result: components['schemas']['SandboxForkResult']) => {
+        if (result.error || !result.sandbox) {
+          return new SandboxError(
+            `${result.error?.code ?? 500}: ${
```
mishushakov: we shouldn't assume 500 when we don't know it's 500

packages/js-sdk/tests/sandbox/fork.test.ts
```diff
@@ -0,0 +1,139 @@
...
+  vi.restoreAllMocks()
+})
+
+test('fork with count lower than 1 fails', async () => {
+  await expect(
+    Sandbox.fork('sbx-test', { count: 0, apiKey: TEST_API_KEY })
+  ).rejects.toThrowError(InvalidArgumentError)
+})
+
+test('fork maps per-fork results to sandbox instances and errors', async () => {
```
mishushakov: we can skip mock tests for now

mishushakov (commented):
reviewed

packages/js-sdk/src/sandbox/index.ts
mishushakov: It threads the caller's connection options (apiKey, domain, headers, proxy, logger, ...) into the new instances — the `Sandbox` constructor builds its `ConnectionConfig` from these opts, so without it the forks would fall back to env vars only. Same pattern as `create` and `connect` (`new this({ ...sandboxInfo, ...config })`).

packages/js-sdk/src/sandbox/index.ts
mishushakov: `this.fork` here is this very instance method — calling it would recurse infinitely. We need the *static* `fork`, and `this.constructor` (rather than hardcoding `Sandbox`) resolves it on the runtime class, so a subclass like code-interpreter's `Sandbox` gets forks constructed as instances of the subclass.

packages/js-sdk/src/sandbox/sandboxApi.ts
mishushakov: Changed to generic `NotFoundError` in a0ca9ad (and `NotFoundException` in Python), now surfacing the API error message when present. One heads-up: both generic classes carry a `@deprecated` tag ("use FileNotFoundError or SandboxNotFoundError instead") — you may want to lift that if this becomes the intended class for ambiguous 404s.

packages/js-sdk/src/sandbox/sandboxApi.ts
mishushakov: Fixed in a0ca9ad — the code is only included when the API actually returned an error object; otherwise the message is just "Failed to start forked sandbox" with no fabricated code. Same in Python.

packages/js-sdk/tests/sandbox/fork.test.ts
mishushakov: Removed in a0ca9ad. Kept the `count < 1` validation test since it exercises real client-side validation without mocks.

packages/python-sdk/e2b/sandbox_sync/sandbox_api.py
```diff
@@ -403,6 +405,81 @@ def _cls_connect(
...
+    @classmethod
+    def _cls_fork(
+        cls,
+        sandbox_id: str,
+        timeout: Optional[int] = None,
+        count: Optional[int] = None,
+        logger: Optional[logging.Logger] = None,
+        **opts: Unpack[ApiParams],
+    ) -> List[Union[SandboxCreateResponse, SandboxException]]:
+        timeout = timeout or SandboxBase.default_sandbox_timeout
```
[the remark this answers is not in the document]
mishushakov: Fixed in a0ca9ad — both sync and async fork paths now use an explicit `is None` check so `timeout=0` is preserved.

packages/js-sdk/src/sandbox/sandboxApi.ts
```diff
@@ -302,6 +303,45 @@ export interface SandboxPauseOpts extends SandboxApiOpts {
   keepMemory?: boolean
 }
 
+/**
+ * Options for forking a sandbox.
+ */
+export interface SandboxForkOpts extends SandboxApiOpts {
```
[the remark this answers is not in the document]
mishushakov: Fixed in a0ca9ad — `SandboxForkOpts` now extends `ConnectionOpts`, matching `SandboxConnectOpts`.

packages/js-sdk/src/sandbox/sandboxApi.ts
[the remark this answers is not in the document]
mishushakov: Fixed in 7d7e7ff — the whole-request 404 is now detected from `res.response.status` instead of the parsed error body, with the body message still used when present.

packages/js-sdk/src/api/index.ts
```diff
@@ -23,6 +23,35 @@ export function validateApiKey(apiKey: string): void {
...
+    )
+  }
+
+  if (code === 429) {
+    const message = 'Rate limit exceeded, please try again later'
+    return new RateLimitError(content ? `${message} - ${content}` : message)
+  }
+
+  return new errorClass(`${code}: ${content}`, stackTrace)
+}
```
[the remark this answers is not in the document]
mishushakov: Fixed in 7d7e7ff — per-fork 404 codes now map to `NotFoundError` at the fork call site. Deliberately not added to the shared `apiErrorFromCode`: 404 semantics are call-site-specific in this SDK (kill returns false, getInfo throws SandboxNotFoundError), and a global branch would hijack 404s from callers passing custom classes like `BuildError`/`VolumeError`.

packages/python-sdk/e2b/api/__init__.py
```diff
@@ -80,6 +80,32 @@ class SandboxCreateResponse:
...
+
+    if status_code == 429:
+        text = f"{status_code}: Rate limit exceeded, please try again later."
+        if message:
+            text += f" - {message}"
+        return RateLimitException(text)
+
+    return default_exception_class(f"{status_code}: {message}").with_traceback(
+        stack_trace
+    )
```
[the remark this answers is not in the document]
mishushakov: Fixed in 7d7e7ff — same as the JS side: per-fork 404 maps to `NotFoundException` at the fork call site; `api_exception_from_code` intentionally leaves 404 to callers.

packages/js-sdk/tests/sandbox/fork.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+)
+
+sandboxTest.skipIf(isDebug)('fork a killed sandbox fails', async () => {
+  const sandbox = await Sandbox.create()
+  await sandbox.kill()
+
+  await expect(sandbox.fork()).rejects.toThrowError(NotFoundError)
+})
```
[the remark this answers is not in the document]
mishushakov: Fixed in 7d7e7ff — switched to plain `test.skipIf(isDebug)`.

packages/js-sdk/src/sandbox/sandboxApi.ts
```diff
@@ -1209,6 +1251,72 @@ export class SandboxApi {
...
+      body: {
+        timeout: timeoutToSeconds(timeoutMs),
+        count,
+      },
+      signal: config.getSignal(opts?.requestTimeoutMs, opts?.signal),
+    })
+
+    // check the status, not the parsed error body — openapi-fetch leaves
+    // `error` unset for non-2xx responses with an empty body
+    if (res.response.status === 404) {
```
matthewlouisbrockman, line 1282: We have sandboxNotFound error right?

packages/python-sdk/e2b/sandbox_sync/sandbox_api.py
```diff
@@ -403,6 +410,91 @@ def _cls_connect(
...
+            body=SandboxForkRequest(timeout=timeout, count=count),
+        )
+
+        if res.status_code == 404:
+            message = (
+                res.parsed.message
+                if isinstance(res.parsed, Error)
+                else f"Sandbox {sandbox_id} not found"
+            )
+            raise NotFoundException(message)
```
matthewlouisbrockman: we have sandboxNotFound i think instead of just not found

matthewlouisbrockman (approved):
probably main thing we need to fix next is the footgun where when snapshotting times out it kills the sandbox; is edge case but should address this sooner than later

packages/js-sdk/src/sandbox/sandboxApi.ts
mishushakov: Final resolution after discussion (572474f): whole-request 404 is back to `SandboxNotFoundError` / `SandboxNotFoundException` — it means the source sandbox wasn't found, same semantics as pause/connect/setTimeout/getInfo, and it subclasses `NotFoundError` so generic catches still work. Per-fork 404s stay generic `NotFoundError` / `NotFoundException` since they refer to a resource needed to start that fork, not the source sandbox.

PR #1560, opened 2026-07-20
chore(deps): remove dead pnpm overrides

mishushakov:
Follow-up to #1559: audits the `pnpm.overrides` list in the root package.json and removes four entries whose targets are no longer in the dependency graph at all — `@next/eslint-plugin-next>glob` (the parent package is gone), `yaml@2.x`, `@tootallnate/once`, and `flatted`. The lockfile change is header-only plus reverting one no-longer-rewritten optional peer range; no resolved package versions change. Verified that `pnpm audit` is clean and that every remaining resolution stays at or above its patched floor even with all overrides removed, though the remaining entries are kept because no parent's declared range excludes the vulnerable versions — the overrides are the only enforcement of those floors. Lint and typecheck pass.


jakubno (approved):

mishushakov:
Superseded by #1561, which contains the same override cleanup plus the CLI release changeset.

PR #1580, opened 2026-07-23
feat(cli): rename --team flag to --project and add E2B_PROJECT_ID env var

mishushakov:
Renames the `--team` flag to `-t, --project` on `template list`, `template publish`, `template unpublish`, and `template delete`. `--team` keeps working as a hidden alias that prints a deprecation warning to stderr. The project ID can now also be set via the new `E2B_PROJECT_ID` environment variable, with `E2B_TEAM_ID` still supported as a fallback. Resolution precedence: `--project` > `--team` > `E2B_PROJECT_ID` > `E2B_TEAM_ID` > `~/.e2b/config.json`.

## Usage

```sh
e2b template list --project <project-id>   # new flag (also -t)
e2b template list --team <project-id>      # still works, warns: "The --team flag is deprecated, use --project instead."

E2B_PROJECT_ID=<project-id> e2b template list   # new env var
E2B_TEAM_ID=<project-id> e2b template list      # still supported
```

### DIFF
--- a/packages/cli/src/options.ts
+++ b/packages/cli/src/options.ts
@@ -37,7 +37,30 @@ export const selectMultipleOption = new commander.Option(
   'select sandbox template from interactive list'
 )
 
-export const teamOption = new commander.Option(
-  '-t, --team <team-id>',
-  'specify the team ID that the operation will be associated with. You can find team ID in the team settings in the E2B dashboard (https://e2b.dev/dashboard?tab=team).'
+export const projectOption = new commander.Option(
+  '-t, --project <project-id>',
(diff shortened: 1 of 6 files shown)


packages/cli/src/options.ts
```diff
@@ -37,7 +37,30 @@ export const selectMultipleOption = new commander.Option(
...
+
+/**
+ * Read the project ID from parsed command options, honoring the deprecated
+ * --team flag (with a warning) when --project is not set.
+ */
+export function projectIdFromOptions(opts: {
+  project?: string
+  team?: string
+}): string | undefined {
+  if (opts.team && !opts.project) {
```
Kraci: Shouldn't we show the console error always if `opts.team` is provided? To always inform this is deprecated

Kraci (approved):

PR #1581, opened 2026-07-23
fix(js-sdk): avoid createRequire shim that crashes Cloudflare Workers

### LINKED ISSUE #1579: [Bug]: e2b 2.35.1 ESM bundle crashes in Cloudflare Workers at createRequire(import.meta.url)
### Sandbox ID or Build ID

None — module evaluation fails before an E2B API request is made or a sandbox ID is allocated.

### Environment

- `e2b@2.35.1`
- Cloudflare Workers, module syntax
- `nodejs_compat` enabled
- Wrangler `4.113.0`
- Regression from `e2b@2.32.0`, which works in the same Worker and with the same application code

### Timestamp of the issue

2026-07-23 08:00 UTC

### Frequency

Happens every time

### Expected behavior

Importing `e2b` in a Cloudflare Worker should succeed, allowing `Sandbox.create()` to make the sandbox API request.

### Actual behavior

The ESM module throws during evaluation, before `Sandbox.create()` makes any request:

```text
The argument 'path' The argument must be a file URL object, a file URL string, or an absolute path string.. Received 'undefined'
```

The published `dist/index.mjs` in 2.35.1 contains this eager initializer:

```js
import { createRequire } from "node:module";
var __require = (() => createRequire(import.meta.url))();
```

Wrangler preserves it in the flattened production Worker bundle. In workerd, `import.meta.url` is `undefined` in that bundled context, so `createRequire` throws immediately.

### Issue reproduction

Install `e2b@2.35.1` in a Worker with `nodejs_compat` and use a lazy import so the failure is isolated to module evaluation:

```ts
export default {
  async fetch() {
    const { Sandbox } = await import('e2b')
    return Response.json({ loaded: typeof Sandbox.create === 'function' })
  },
}
```

Build/deploy with Wrangler and request the Worker route. The import throws the error above. Pinning the identical project to `e2b@2.32.0` makes the import and sandbox creation work again.

A production-bundle check also makes the regression visible without an E2B API key:

```sh
wrangler deploy --dry-run --outdir dist-worker
rg 'createRequire\(import\.meta\.url\)' dist-worker
```

### Additional context

The JS SDK changed its build from `tsup` to `tsdown` between 2.32.0 and 2.35.x, and the published artifact now includes the Rolldown-style `createRequire(import.meta.url)` runtime helper.

This exact generated-code incompatibility has affected workerd elsewhere: vitejs/vite#20983 was fixed by reverting the dependency change that emitted `createRequire(import.meta.url)` (vitejs/vite#20988).

Suggested fixes:

1. keep the main sandbox client entry Worker/edge-safe and avoid eagerly initializing Node `require` at module scope;
2. separate Node-only template/build functionality behind a conditional export or subpath if needed;
3. add a CI smoke test that bundles the package with Wrangler and evaluates the resulting Worker under workerd.

Related older E2B reports show that edge runtimes are intended to be supported, but they cover different failures: #315 (`process.env`) and #531 (redirect handling). I could not find an existing E2B issue for this `createRequire` / 2.35.1 regression.

mishushakov:
Fixes #1579

## Problem

Since the switch from tsup to tsdown, `e2b` ≥2.33 crashes at import time in Cloudflare Workers (`nodejs_compat`):

```text
The argument 'path' The argument must be a file URL object, a file URL string, or an absolute path string.. Received 'undefined'
```

Two bare `require()` calls in `src/utils.ts` (the `sha256` Node fallback and `dynamicRequire`) made Rolldown emit this eager helper at module scope in `dist/index.mjs`:

```js
import { createRequire } from "node:module";
var __require = /* #__PURE__ */ (() => createRequire(import.meta.url))();
```

In workerd's bundled context `import.meta.url` is `undefined`, so `createRequire` throws during module evaluation — before `Sandbox.create()` can make any request. tsup/esbuild emitted a lazy shim that only threw when called, which is why 2.32.0 worked.

## Fix

- `dynamicRequire` now loads Node built-ins via `process.getBuiltinModule` (Node ≥20.16, Bun, Deno 2 — the package already requires Node ≥20.18.1), so no `require` reference exists in ESM source and no shim is emitted.
- The `sha256` WebCrypto fallback goes through `dynamicRequire` instead of a bare `require('node:crypto')`.
- New build step `scripts/check-bundle-edge-compat.mjs` fails `pnpm build` (and therefore `prepublishOnly`) if a `require` shim (`node:module` import, `createRequire(`, or `__require`) ever reappears in `dist/index.mjs`.
- New unit tests cover `dynamicRequire` and both `sha256` paths.

## Usage example

The reproduction from the issue now works — a Worker with `nodejs_compat`:

```ts
export default {
  async fetch() {
    const { Sandbox } = await import('e2b')
    return Response.json({ loaded: typeof Sandbox.create === 'function' })
  },
}
```

## Verification

- `wrangler deploy --dry-run` of a Worker using the packed tarball: `rg 'createRequire\(import\.meta\.url\)'` finds nothing in the production bundle (the reporter's check).
- `wrangler dev` (real workerd): `curl` returns `{"loaded":true}` — module evaluation and `Sandbox.create` reference succeed.
- Node ESM (`dist/index.mjs`) and CJS (`dist/index.js`) both still load.
- `vitest run` over `tests/api`, `tests/template`, `tests/utils.test.ts`, `tests/connectionConfig.test.ts`, `tests/sandbox/signature.test.ts`: 232 passed, 3 skipped.
- CLI (which bundles the workspace SDK from source) still typechecks and builds.

No Python SDK changes needed — this is a JS bundler artifact issue.

### DIFF
--- a/.changeset/quiet-panthers-repair.md
+++ b/.changeset/quiet-panthers-repair.md
@@ -1,0 +1,5 @@
+---
+'e2b': patch
+---
+
+Fix ESM bundle crashing at import time in Cloudflare Workers (and other edge runtimes) with `The argument 'path' must be a file URL object, a file URL string, or an absolute path string. Received 'undefined'`. Bare `require` calls in the SDK source made the bundler emit an eager `createRequire(import.meta.url)` shim at module scope, which throws in workerd where `import.meta.url` is undefined. Node.js built-ins are now loaded via `process.getBuiltinModule` instead, and the build fails if a `require` shim ever reappears in the ESM bundle.
--- a/packages/js-sdk/package.json
+++ b/packages/js-sdk/package.json
@@ -24,7 +24,7 @@
   "types": "dist/index.d.ts",
   "scripts": {
     "prepublishOnly": "pnpm build",
-    "build": "tsc --noEmit && tsdown",
+    "build": "tsc --noEmit && tsdown && node scripts/check-bundle-edge-compat.mjs",
     "dev": "tsdown --watch",
     "example": "tsx example.mts",
     "test": "vitest run",
--- a/packages/js-sdk/scripts/check-bundle-edge-compat.mjs
+++ b/packages/js-sdk/scripts/check-bundle-edge-compat.mjs
@@ -1,0 +1,32 @@
+// Fails the build if the ESM bundle contains a module-scope `require` shim.
+//
+// Bundlers turn bare `require` references in ESM source into a
+// `createRequire(import.meta.url)` helper that is evaluated eagerly when the
+// module loads. In edge runtimes (e.g. Cloudflare Workers) `import.meta.url`
+// is undefined in bundled code, so importing the SDK crashes before any API
+// call is made. See https://github.com/e2b-dev/E2B/issues/1579
+//
+// Instead of `require`, load Node.js built-ins with `dynamicRequire` from
+// src/utils.ts (backed by `process.getBuiltinModule`).
+import { readFileSync } from 'node:fs'
+
+const bundlePath = new URL('../dist/index.mjs', import.meta.url)
+const bundle = readFileSync(bundlePath, 'utf8')
+
+const forbidden = [
+  /from\s*['"]node:module['"]/,
+  /\bcreateRequire\s*\(/,
+  /\b__require\b/,
+]
+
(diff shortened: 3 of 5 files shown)


mishushakov:
I decided we will just replace node:crypto with WebCrypto

Issue #1587, opened 2026-07-23
[Bug]: reattached stdout stream stalls after auto-resume (regression in python-sdk 2.29.1)


nseniak:
## Summary

After a sandbox auto-pauses and resumes, reattaching to an existing
long-lived background process via `commands.connect(pid)` produces a
stdout stream that intermittently delivers its first response and then
goes **silent** — the `on_stdout` callback stops firing for subsequent
responses even though the process is still running and still writing.

This is a **regression introduced in python-sdk 2.29.1** and present in
every release since (through 2.34.0). The likely change is `cb061d2`
("Fix several command and PTY streaming issues"), which reworked
incremental stdout/stderr decoding and stream-connection teardown.

## Bisect (strong evidence)

Identical workload, same account/templates; only the SDK version changed.
The workload is a real stdio server (`@modelcontextprotocol/server-everything`
over `npx`) driven through repeated pause → resume → `commands.connect(pid)`
reattach cycles. "Stall rate" = fraction of reattach cycles where the
reconnected stdout stream failed to deliver a complete response, over 6
cycles:

| Version | Stall rate |
|---------|------------|
| 2.20.2  | 1/6 |
| 2.21.0  | 1/6 |
| 2.27.0  | 1/6 |
| 2.29.0  | 1/6  ← **last good** |
| 2.29.1  | 6/6  ← **first bad** |
| 2.30.0  | 6/6 |
| 2.34.0  | 6/6 |

The transition is sharp and clean: every version up to and including
2.29.0 stalls ~1/6; every version from 2.29.1 on stalls ~6/6. The only
command/stream change in 2.29.1 is `cb061d2`.

## Reproduction conditions

The stall reproduces reliably only through a **realistic streaming
workload** — a long-lived process reattached after resume that answers
**several** requests with **sizeable, multi-chunk** stdout responses (in
our case JSON-RPC `tools/list` / `resources/list` / `prompts/list` from an
stdio MCP server). The pattern observed: the first response after reattach
arrives, later ones never do.

A trivial synthetic process (one small line per second, or a single large
one-shot write) does **not** isolate the regression — those either stream
fine on both versions, or hit an unrelated single-write truncation that is
version-independent. So the reliable repro is the real stdio-server flow
above; we can share the exact harness on request.

## Expected vs actual

- **Expected (≤ 2.29.0):** after reattach, `on_stdout` keeps firing for
  every subsequent response; a full multi-request refresh completes.
[...]

PR #1590, opened 2026-07-23
fix(python-sdk): cancel HTTP/2 server streams closed before the server ends them

### LINKED ISSUE #1352: commands.connect to a resumed sandbox stops delivering stdout after a large response

mishushakov:
## Problem

The Python SDK's connect client (`e2b_connect`) talks to envd over a single shared HTTP/2 connection (one `httpcore` pool per event loop, all sandboxes multiplexed). When a command/PTY/watch stream is closed before the server finished it — `disconnect()`, an error, stream-setup failure, or the consumer just stopping — httpcore only stops reading the response; it never sends `RST_STREAM`. The server side therefore keeps the stream attached and keeps writing into a flow-control window nobody drains.

On envd, such a half-open stream becomes a stuck output consumer, and one stuck consumer blocks envd's process output fan-out for **every** consumer of that process (see e2b-dev/infra#3362). That is the mechanism behind #1352 and part of the picture in #1587.

## Fix

`call_server_stream` / `acall_server_stream` now send `RST_STREAM` (CANCEL) whenever the stream ends without the server having completed it, so the proxy/envd tear the stream down and detach the consumer. Normal completions send nothing extra. The reset is best-effort and no-ops on HTTP/1.1 (where closing the response already closes the connection).

## Note on #1587

This fixes the client-side half of the problem (abandoned streams manufacturing dead server-side consumers). The stall reported in #1587 — stdout going silent after **auto-pause/resume** — is primarily caused by consumers captured in the pause snapshot whose restored connections block envd's fan-out after resume; no client-side change can reach those. That is fixed in envd by e2b-dev/infra#3362 and needs a deploy + envd rollout to take effect. Full root-cause analysis in the issue thread.

JS SDK needs no equivalent change: connect-es aborts streams via `AbortController`, which already produces a stream reset/connection teardown.

## Usage example

```python
handle = await sandbox.commands.connect(pid, on_stdout=print)
# Stop receiving events; the H2 stream is now actually cancelled server-side
# instead of lingering half-open on the shared envd connection.
await handle.disconnect()
```

## Testing

- New unit tests (`tests/e2b_connect/test_stream_reset.py`) run the connect client against an in-process mock HTTP/2 transport and assert on emitted frames: early close sends exactly one `RST_STREAM(CANCEL)` for the stream; normal completion sends none (sync + async). They fail without the fix.
- Live: full `commands`, `pty`, and `files watch` suites pass (sync + async, 74 tests); `disconnect()` followed by 2MB of process output to a second consumer works.
- `format`, `lint`, `typecheck` clean.

Fixes #1352

### DIFF
--- a/.changeset/light-poems-refuse.md
+++ b/.changeset/light-poems-refuse.md
@@ -1,0 +1,5 @@
+---
+"e2b": patch
+---
+
+Cancel HTTP/2 server streams (`RST_STREAM`) when a command/PTY/watch stream is closed before the server finished it (e.g. `disconnect()`), instead of leaving the stream half-open on the shared envd connection. Abandoned half-open streams stay attached server-side and can block envd's process output fan-out for every other consumer of the same process (#1352, #1587).
--- a/packages/python-sdk/e2b_connect/client.py
+++ b/packages/python-sdk/e2b_connect/client.py
@@ -5,6 +5,8 @@ import logging
 import struct
 import typing
 
+import h2.errors
+
 from httpcore import (
     ConnectionPool,
     AsyncConnectionPool,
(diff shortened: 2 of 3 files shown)


mishushakov:
aiming to superseed this with
https://github.com/e2b-dev/E2B/pull/1558

PR #1594, opened 2026-07-23
feat(js-sdk): run the template test suite on Bun and Deno

mishushakov:
## What

Extends the Bun (#1584) and Deno (#1585) vitest runs with the `template` project, and fixes the two real runtime bugs the suite surfaced. Full matrix now passes: Node 131 ✓, Bun 522 ✓, Deno 523 ✓ (remainder skipped).

```jsonc
// packages/js-sdk/package.json
"test:bun":  "bunx --bun vitest run --project unit --project connectionConfig --project template",
"test:deno": "deno run -A npm:vitest run --project unit --project connectionConfig --project template",
```

## Bug 1 — Bun: build-error stack traces pointed at the wrong frame

Bun's JavaScriptCore performs proper tail calls in strict mode — and its bytecode generator also treats `const r = f(); return r` as a tail call — so the frame of any method that returns a call result directly disappears from `new Error().stack`. Every builder method implemented as `return this.runInNewStackTraceContext(() => this.runCmd(...))` (`remove`, `rename`, `makeDir`, `makeSymlink`, `pipInstall`, `npmInstall`, `bunInstall`, `aptInstall`, `addMcpServer`, `gitClone`, `betaDevContainerPrebuild`, `betaSetDevContainerStart`) lost its own frame, and the fixed-depth walk landed one frame past the user's call site — under Bun, a failed build step's error pointed at test-runner internals instead of the user's `.remove(...)` line.

Fixed by capturing `getCallerFrame(STACK_TRACE_DEPTH - 1)` in the public method's own body (frames are still live there) and passing it down via `runInStackTraceOverrideContext` — the pattern `copy()`/`copyItems()` already used, whose traces were correct under Bun all along. `runInNewStackTraceContext` and the enable/disable toggle are deleted.

```ts
// now traces to the user's line on Node, Bun, and Deno alike
Template().fromBaseImage().remove('nonexistent/path')
```

## Bug 2 — Deno: template uploads used chunked transfer encoding

Deno's `fetch` ignores an explicit `Content-Length` header on stream bodies and falls back to `Transfer-Encoding: chunked` — exactly the failure #1243 fixed for Node, since S3-compatible presigned PUT URLs reject chunked uploads with 501.

`uploadFile` now sends the spooled archive as a file-backed `Blob` (`fs.openAsBlob`, lazily read, so memory stays bounded); every runtime derives `Content-Length` from the blob itself. One wrinkle: Bun infers a MIME type from the file extension (unstrippable — it survives `slice()` and re-wrapping) and would send a `Content-Type` the presigned signature doesn't cover (observed 403 Forbidden against production storage). Typed blobs therefore fall back to `blob.stream()` + explicit `Content-Length`, which Bun honors. The choice is feature-detected via `blob.type`, not runtime-sniffed. `tests/template/uploadFile.test.ts` now also asserts no `Content-Type` header is sent.

As part of this, `tarFileStream` became `spoolTarArchive`, returning `{ path, size, cleanup }` with caller-owned cleanup instead of a self-deleting read stream.

## Python SDK parity

Intentionally none: CPython never elides frames, so `_run_in_new_stack_trace_context` produces correct traces, and `upload_file` already sends a sized file body via httpx.

## Testing

- `vitest run --project template` — Node 131 passed / 3 skipped
- `pnpm test:bun` — 522 passed / 35 skipped (real template builds + uploads against prod)
- `pnpm test:deno` — 523 passed / 34 skipped (real template builds + uploads against prod)
- `pnpm build`, `lint`, `typecheck`, `prettier --check` clean


mishushakov:
Split into #1595 (Deno: template suite + uploadFile Content-Length fix) and the Bun PR above (template suite + stack-trace fix) — the two changes are independent against main.

mishushakov:
Bun counterpart: #1596

PR #1607, opened 2026-07-24
fix(js-sdk): share lazy fetcher loading and drop the new Function import trick

mishushakov:
Extracts the duplicated api/envd fetcher-loading logic into shared `createRuntimeFetch` + `buildDispatchedFetch` helpers in `undici.ts`, fixing two behaviors along the way: a failed fetcher build is no longer cached forever (the next request retries, with a guarded compare-and-clear so a stale awaiter can't clobber a newer in-flight build — covered by a regression test reproducing the microtask interleaving), and the no-undici fallback now late-binds `globalThis.fetch` so fetch replacements installed after the first request (msw, instrumentation) are picked up.

`loadUndici` now uses the shared `dynamicImport` helper, whose import is kept opaque to downstream bundlers via `webpackIgnore`/`@vite-ignore` annotations instead of the `new Function('return import(...)')` trick — so environments that disallow code generation from strings (CSP, `--disallow-code-generation-from-strings`) now load undici normally instead of silently degrading to the global fetch. The now-internal `toUndiciRequestInput`/`UndiciRequestInit` are no longer exported. No user-facing API changes; verified with the unit suites (lint/typecheck clean) plus real API integration tests through the new dispatcher path.

### Test-suite fallout from the import fix

Dropping the `new Function` trick exposed a hidden test dependency: that trick throws under vitest's vm evaluation, so every vitest run had silently fallen back to the msw-patchable global fetch. With module loading un-broken, the Node test runs dispatched through real undici, bypassing msw's `globalThis.fetch` patch — mocked requests escaped to the real API (real 404s in the tags suite, 3-minute hangs in the abortSignal suites waiting for msw's `request:start`, and 28 real template builds per run from the stacktrace suite).

Fixed centrally: `tests/globalFetchFallback.setup.ts`, registered via `setupFiles` for the unit and template projects, mocks `buildDispatchedFetch` to run the SDK's real undici-unavailable fallback (late-bound `globalThis.fetch`), so msw suites need no per-file mock and future msw suites are covered automatically. Suites that inject their own `loadUndici` (the api/envd transport tests) keep it, so the dispatcher wiring itself stays covered. Verified under Node, Bun, and Cloudflare workerd.

Closes SDK-290


matthewlouisbrockman (approved):
nice improvement

PR #1608, opened 2026-07-24
fix(python-sdk): enable TCP keepalive on API and envd HTTP transports

wj-e2b:
Add keepalive socket options (SO_KEEPALIVE plus TCP_KEEPIDLE / macOS TCP_KEEPALIVE with a 60s initial delay where supported) and keepalive httpx transport subclasses in e2b.api, and wire them into the sync and async API, envd, and proxy transports. On Windows and platforms without a keepalive-idle constant, only SO_KEEPALIVE is set.

Custom network backends force the socket options at TCP connect time, since httpcore 1.0.x drops socket_options when constructing proxy connections. Because an explicit transport disables httpx trust_env proxy handling, a shared _build_env_proxy_mounts helper rebuilds the HTTP_PROXY/HTTPS_PROXY/NO_PROXY mounts for ApiClient. Also drop a dead verify parameter from _create_transport.

### DIFF
--- a/.changeset/warm-sockets-keep-python.md
+++ b/.changeset/warm-sockets-keep-python.md
@@ -1,0 +1,5 @@
+---
+'@e2b/python-sdk': patch
+---
+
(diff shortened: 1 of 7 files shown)


mishushakov:
@wj-e2b keep in mind, we are in process of switching out the transport entirely
https://github.com/e2b-dev/E2B/pull/1601

PR #1610, opened 2026-07-25
fix(js-sdk): detect Request inputs by shape, not instanceof, in the Node fetch bridge

himself65:
## Symptom

In an app embedding the JS SDK, **every** control-plane call (`Sandbox.create`, `Sandbox.list`, …) crashed with:

```
TypeError: Failed to parse URL from [object Request]
    at fetch (…/undici/index.js:155:10)
    at wrapped (…/e2b/src/api/http2.ts:90)
    at …/e2b/src/api/inflight.ts:73
```

## Root cause

`toUndiciRequestInput` (and the in-flight limiter's signal extraction) decide Request-ness with `input instanceof Request`. That brand check misses any Request minted by a **different** `Request` class — and replaced/duplicated Request classes are a real thing in Node processes: runtimes and tools swap `globalThis.Request` exactly like they swap `globalThis.fetch` (test environments, instrumentation, server shims). `src/undici.ts` already late-binds the global fetch for this very reason ("msw, instrumentation may replace `globalThis.fetch`") — this PR extends the same reasoning to `Request`.

The concrete case we debugged: the host app served HTTP with `@hono/node-server`, which installs a lightweight `class Request extends GlobalRequest` shim onto `globalThis` at import — and the dependency tree contained **multiple copies** of that shim (different versions via different packages), each with its own class. A Request built by one copy's class fails `instanceof` against the other copy's, so the SDK passed it **verbatim** into the npm-`undici` `fetch`, whose own brand check also can't recognize it → it coerces the object to a URL string → `new URL('[object Request]')` throws. Sandbox creation and listing died on every call; which copy "wins" the global is import-order-dependent, so the failure appears and disappears with unrelated dependency changes.

## Fix

Treat anything carrying `url` + `method` as a Request (`isRequestLike` in `src/utils.ts`) and destructure it into the plain `(url, init)` pair undici accepts — the existing destructuring already produces exactly that. Two call sites:

- `toUndiciRequestInput` (`src/undici.ts`) — the crash above;
- `limitConcurrency` (`src/api/inflight.ts`) — a foreign-class Request's **abort signal** was silently ignored while the request waited for a slot.

No behavior change for native Requests (the `instanceof` fast path still short-circuits), strings, or `URL` objects (no `method`, so the shape check can't match them).

## Tests

Both new tests reproduce the real-world event faithfully with **sibling subclasses** of the native Request (instances of one are fully functional Requests but fail `instanceof` against the other) plus `vi.stubGlobal('Request', …)`:

- `tests/undici.test.ts` — a foreign-class Request must reach the (injected fake) undici fetch destructured as `(url, init)`, not verbatim;
- `tests/api/inflight.test.ts` — an already-aborted foreign-class Request must reject with `AbortError` without consuming a slot.

Red/green verified: with `src/` reverted, exactly these two tests fail; with the fix, `vitest run tests/undici.test.ts tests/api/inflight.test.ts` = 11/11, and `pnpm build` (tsc + tsdown) + `pnpm lint` are clean. Changeset included (`e2b`: patch).

### DIFF
--- a/.changeset/metal-pandas-brake.md
+++ b/.changeset/metal-pandas-brake.md
@@ -1,0 +1,5 @@
+---
+'e2b': patch
+---
+
+Detect `Request` inputs by shape instead of `instanceof` in the Node fetch bridge. A `Request` minted by a different `Request` class — runtimes and tools replace `globalThis.Request` the same way they replace `globalThis.fetch` (test environments, instrumentation, server shims such as `@hono/node-server`, or two coexisting copies of one shim) — previously failed the brand check, was passed to undici's `fetch` verbatim, and crashed every API call with `Failed to parse URL from [object Request]`. Such Requests are now destructured into a plain `(url, init)` pair like any other, and their abort signal is honored while queued by the in-flight limiter.
--- a/packages/js-sdk/src/api/inflight.ts
+++ b/packages/js-sdk/src/api/inflight.ts
@@ -1,3 +1,5 @@
+import { isRequestLike } from '../utils'
+
 /**
  * Simple FIFO semaphore used to cap the number of in-flight requests sent
  * through a fetch dispatcher.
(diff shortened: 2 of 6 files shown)


himself65:
@cla-bot check

mishushakov:
Hey, thanks for the contribution!
We have superseded this PR and fixing it ourselves: https://github.com/e2b-dev/E2B/pull/1618

PR #1612, opened 2026-07-26
SDK: fromFedoraImage/fromAlpineImage/fromArchImage helpers

tomassrnka:
Adds the non-Debian base-image helpers to both SDKs, mirroring `fromUbuntuImage`/`fromDebianImage`/etc.: `fromFedoraImage`, `fromAlpineImage`, `fromArchImage` (`from_*_image` in Python). Customer-facing half of infra #3381 — the engine already boots these families, but they were reachable only through the generic `fromImage()`.

```ts
Template().fromFedoraImage() // fedora:44
Template().fromAlpineImage() // alpine:3.24
Template().fromArchImage() // archlinux:latest
```

Per review the client-side helper tests were removed — they only asserted the generated `FROM` string. Build coverage lives in the orchestrator's distro build tests.

Docs: e2b-dev/docs#313.


.changeset/multi-distro-base-image-helpers.md
```diff
@@ -0,0 +1,5 @@
+---
+'e2b': minor
```
mishushakov, line 2: changeset should be both JS SDK and Python SDK

packages/js-sdk/src/template/index.ts
```diff
@@ -426,6 +426,18 @@ export class TemplateBase
...
+    return this.fromImage(`fedora:${variant}`)
+  }
+
+  fromAlpineImage(variant: string = 'latest'): TemplateBuilder {
+    return this.fromImage(`alpine:${variant}`)
+  }
+
+  fromArchImage(variant: string = 'latest'): TemplateBuilder {
+    return this.fromImage(`archlinux:${variant}`)
+  }
```
mishushakov: I think maybe we should specify a known current working version with E2B, not latest.
my worry is that the builds will no longer be reproducible: latest moves to something else, build breaks

packages/js-sdk/tests/template/methods/fromDistroImages.test.ts
mishushakov, line 1: the tests here are practically useless, we should be verifying the build on the backend instead.
I will mock the client build tests when I have more time.

packages/python-sdk/e2b/template/main.py
```diff
@@ -846,6 +846,51 @@ def from_ubuntu_image(self, variant: str = "latest") -> TemplateBuilder:
...
+        :param variant: Arch Linux image variant (default: 'latest')
+
+        :return: `TemplateBuilder` class
+
+        Example
+        ```python
+        Template().from_arch_image('base-devel')
+        ```
+        """
+        return self.from_image(f"archlinux:{variant}")
```
mishushakov: same as I said about JS

packages/python-sdk/tests/async/template_async/methods/test_from_distro_images.py
mishushakov, line 1: same as above

packages/python-sdk/tests/sync/template_sync/methods/test_from_distro_images.py
mishushakov, line 1: same as above

mishushakov (changes requested):
would pin the versions for reproducible builds and remove the tests

.changeset/multi-distro-base-image-helpers.md
tomassrnka: Done in 819a8937f — changeset is now `'e2b': minor` + `'@e2b/python-sdk': minor`.

packages/js-sdk/src/template/index.ts
tomassrnka: Agreed, done in 819a8937f: `fromFedoraImage` defaults to `fedora:42` and `fromAlpineImage` to `alpine:3.22`. Both verified building and booting on real KVM before pinning — `Fedora Linux 42` and `Alpine Linux v3.22`, HTTPS 200 in-sandbox.

`fromArchImage` I left on `latest` deliberately: Arch is a rolling release, and template provisioning runs `pacman -Syu` during the build, so whatever tag you pin gets upgraded to current packages anyway — a pin would not change the built result. Happy to pin an Arch `base-YYYYMMDD.N` snapshot for uniformity if you prefer, it just will not change anything; one word and I will.

Worth flagging separately: every existing helper defaults to a moving tag (`debian:stable`, `ubuntu:latest`, `python:3`, `node:lts`, `oven/bun:latest`), so these three are now the only pinned ones. Whether the repo wants pinning across the board feels like a broader convention call than this PR should settle — flagging, not proposing.

packages/js-sdk/tests/template/methods/fromDistroImages.test.ts
tomassrnka: Agreed — removed in 819a8937f (JS + both Python sync/async files). They only asserted the generated `FROM` string, which is a tautology over a one-line wrapper. Leaving the real coverage to your backend/mocked build tests.

packages/python-sdk/e2b/template/main.py
tomassrnka: Same treatment in 819a8937f: `from_fedora_image` -> `fedora:42`, `from_alpine_image` -> `alpine:3.22`, `from_arch_image` stays `latest` with the rolling-release rationale in the docstring. See the JS thread for the reasoning and the Arch offer.

packages/python-sdk/tests/async/template_async/methods/test_from_distro_images.py
tomassrnka: Removed in 819a8937f.

packages/python-sdk/tests/sync/template_sync/methods/test_from_distro_images.py
tomassrnka: Removed in 819a8937f.

packages/python-sdk/e2b/template/main.py
```diff
@@ -846,6 +846,51 @@ def from_ubuntu_image(self, variant: str = "latest") -> TemplateBuilder:
         """
         return self.from_image(f"ubuntu:{variant}")
 
+    def from_fedora_image(self, variant: str = "latest") -> TemplateBuilder:
```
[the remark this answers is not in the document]
tomassrnka: Addressed in 819a8937f — the changeset now covers both `e2b` and `@e2b/python-sdk`, so the Python helpers ship through the normal release workflow.

packages/js-sdk/src/template/index.ts
```diff
@@ -426,6 +426,18 @@ export class TemplateBase
...
 
+  fromFedoraImage(variant: string = 'latest'): TemplateBuilder {
+    return this.fromImage(`fedora:${variant}`)
+  }
+
+  fromAlpineImage(variant: string = 'latest'): TemplateBuilder {
+    return this.fromImage(`alpine:${variant}`)
+  }
+
+  fromArchImage(variant: string = 'latest'): TemplateBuilder {
```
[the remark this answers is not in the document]
tomassrnka: Addressed in 819a8937f — changeset now bumps both `e2b` and `@e2b/python-sdk` (it previously listed only `e2b`, so the Python helpers would not have been released).

tomassrnka:
### Heads-up: `Staging / Python SDK Tests (ubuntu-22.04)` is failing on pre-existing flake, not this change

I re-ran it three times. Each run fails **exactly one** test, a **different** one each time, all in `test_network.py`, all with `CommandExitException: Command exited with code 35` (curl `35` = SSL connect error):

| run | test |
|---|---|
| 1 | `sync/sandbox_sync/test_network.py::test_allow_takes_precedence_over_deny` |
| 2 | `sync/sandbox_sync/test_network.py::test_deny_specific_ip` |
| 3 | `async/sandbox_async/test_network.py::test_update_network_clears_existing_rules` |

Always `1 failed, 806 passed, 58 skipped`.

Why it is not this PR:
- **`Production / Python SDK Tests` passes on both ubuntu-22.04 and windows-latest, and `Staging / windows-latest` passes.** Only Staging/ubuntu fails.
- This branch touches `packages/js-sdk/src/template/{index,types}.ts`, `packages/python-sdk/e2b/template/main.py` (three helper defaults + docstrings), the changeset, and deletes three dockerfile-string test files. It does not touch `test_network.py` or anything in the sandbox networking path.
- The check was green at `819a8937f`; the only change since was a commit-message rewrite to drop `Co-Authored-By` trailers for the CLA — the tree hash is byte-identical.

Likely a transient TLS/egress issue against staging rather than a test bug, though `test_network.py` has needed flake mitigation before (#1456, "add wait for status for public traffic network tests"). Flagging in case it is worth a look independently of this PR — it will cost roughly one random failure per run to anything gated on that check.

mishushakov (approved):

packages/js-sdk/src/template/index.ts
```diff
@@ -426,6 +426,20 @@ export class TemplateBase
     return this.fromImage(`ubuntu:${variant}`)
   }
 
+  fromFedoraImage(variant: string = '42'): TemplateBuilder {
```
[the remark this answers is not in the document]
tomassrnka: Correct, and it was still live on the Python side at merge. The JS default had already moved to `fedora:44` in e2e5011; `from_fedora_image` was left on `42`.

Fixed in #1625: Python now defaults to `fedora:44` (and `alpine:3.24`). Both tags are the ones the orchestrator distro build tests cover on infra `main` (`tests/integration/internal/tests/api/templates/distro_build_test.go`), so the defaults track what CI actually builds and boots.

packages/python-sdk/e2b/template/main.py
```diff
@@ -846,6 +846,55 @@ def from_ubuntu_image(self, variant: str = "latest") -> TemplateBuilder:
...
+        :param variant: Alpine image variant (default: '3.22')
+
+        :return: `TemplateBuilder` class
+
+        Example
+        ```python
+        Template().from_alpine_image('3.22')
+        ```
+        """
+        return self.from_image(f"alpine:{variant}")
```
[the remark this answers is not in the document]
tomassrnka: Accurate on all three points. Fixed in #1625:

- `from_fedora_image` `42` -> `44`, `from_alpine_image` `3.22` -> `3.24`, so both SDKs resolve to the same base image when the variant is omitted.
- `TemplateFromImage` in `packages/js-sdk/src/template/types.ts` no longer documents the old defaults (`@param` and `@example` both updated).
- Python docstrings updated to match.

`fromArchImage`/`from_arch_image` stay on `latest` in both SDKs, with the rolling-release rationale in the doc comment.

packages/js-sdk/src/template/index.ts
```diff
@@ -426,6 +426,20 @@ export class TemplateBase
...
+    return this.fromImage(`fedora:${variant}`)
+  }
+
+  fromAlpineImage(variant: string = '3.22'): TemplateBuilder {
+    return this.fromImage(`alpine:${variant}`)
```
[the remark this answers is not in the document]
tomassrnka: Split verdict.

The tests are not coming back: their removal was requested in review — "the tests here are practically useless, we should be verifying the build on the backend instead. I will mock the client build tests when I have more time." They asserted only the generated `FROM` string, which is a tautology over a one-line wrapper around `fromImage()`. Real coverage is the orchestrator distro build tests on infra `main` (`tests/integration/internal/tests/api/templates/distro_build_test.go`), which build and boot `ubuntu:26.04`, `fedora:44`, `archlinux:base` and `alpine:3.24`, plus a rejection case for an unsupported distro.

The description was stale, though, and that half is valid — it still claimed those tests were committed after they had been removed. Body corrected.

PR #194, opened 2023-10-12
SDK download/upload files

jakubno:

PR #1626, opened 2026-07-30
feat(python-sdk): adopt pyqwest 0.8's proxy object, multipart fix, and request loggers

mishushakov:
## What

Tracked in [SDK-302](https://linear.app/e2b/issue/SDK-302/python-sdk-adopt-pyqwest-08-proxy-object-multipart-fix-request-loggers) (part of the [SDK-268](https://linear.app/e2b/issue/SDK-268) stack). Stacked on #1623. Adopts the three features merged into [pyqwest](https://github.com/curioswitch/pyqwest) after 0.7.0 and drops the workarounds the SDK carried while they were missing:

| pyqwest PR | What it adds | What it lets us delete here |
| --- | --- | --- |
| [#194](https://github.com/curioswitch/pyqwest/pull/194) | `Proxy` object (auth, headers, `no_proxy`, scheme) | rejecting `httpx.Proxy` headers; folding credentials into the URL userinfo |
| [#196](https://github.com/curioswitch/pyqwest/pull/196) | multipart form support, and the httpx adapter matching dual sync/async streams sync-first | `_SyncOnlyStream`, the rewrap of httpx's `MultipartStream` |
| [#197](https://github.com/curioswitch/pyqwest/pull/197) | `pyqwest` / `pyqwest.access` loggers, emitted from Rust at `DEBUG` | the "until pyqwest ships a logging middleware" note on the RPC interceptor (pyqwest#192 was superseded) |

> [!IMPORTANT]
> **Blocked on a pyqwest 0.8.0 release.** All three are on `curioswitch/pyqwest@main` but unreleased — the latest PyPI release is 0.7.0 (2026-07-19). So `uv.lock` can't be regenerated and CI's `uv sync --locked` fails until 0.8.0 is published. Everything below was verified against a wheel built locally from `pyqwest@6d56923`.

## How

**Proxy** — `proxy_to_url() -> Optional[str]` becomes `proxy_to_config() -> Optional[ProxyConfig]`, a `NamedTuple` of `(url, auth, headers)` that does double duty: it keys the transport caches (a `pyqwest.Proxy` compares by identity, so caching on one would hand every call its own connection pool) and builds the `Proxy` via `to_pyqwest()`. Consequences for the `proxy` option:

- An `httpx.Proxy`'s custom headers are now honored instead of raising `InvalidArgumentException`.
- Credentials pass through as a `(username, password)` pair rather than being percent-encoded back into the URL userinfo — reqwest sends `Proxy-Authorization` itself.
- A per-proxy `ssl_context` is still rejected; pyqwest has no counterpart.
- Transports are keyed on the whole config, so the same proxy URL with different credentials or headers gets its own pool.

It's a `NamedTuple` rather than a frozen dataclass on purpose: `tests/test_env_var_parsing.py` reloads `e2b.api`, which rebinds the class, and a dataclass's `__eq__` compares class identity — so cache keys created before and after a reload would silently stop matching. Tuples compare by value.

**Multipart** — `ApiPyqwestTransport.handle_request` no longer rewraps request streams. httpx's `MultipartStream` (`files=` uploads, i.e. `files.write`) implements both `SyncByteStream` and `AsyncByteStream`; the 0.7 adapter matched the async case first and raised `TypeError("unreachable")` from inside the body iterator, surfacing as a mid-request `WriteError`. Upstream now matches sync first. The regression test stays and covers the upstream fix.

**Loggers** — no wiring needed (the loggers are process-wide and emitted from Rust), so this is documentation plus a test that pins the behavior. Two comments record how the layers relate: `retrying_http_transport` notes that pyqwest logs requests below the SDK's `logger` option, and `LoggingInterceptor`'s docstring replaces the "stays until pyqwest#192 ships" note — that PR was superseded by #197, whose loggers can't carry a per-sandbox logger nor see streamed messages or the Connect error code of a stream that fails inside a `200 OK`, so the interceptor is permanent.

## Testing

- Unit suite: 240 passed, 1 skipped. `ruff check`, `ruff format --check`, `ty check` clean.
- New: `test_sync_transport_sends_proxy_credentials_and_headers` (the echo server stands in for a proxy, asserting absolute-form request target + `Proxy-Authorization` + the extra header actually reach it) and `test_transport_emits_pyqwest_access_log`. Rewrote the nine `proxy_to_url` narrowing tests for `proxy_to_config`, plus one asserting equal configs share a pool while the `Proxy` objects they build do not compare equal.
- E2E against production on the local 0.8 wheel, sync and async: create → `files.write` (single part) → `write_files` (multi-part body) → file-like write (octet-stream path) → `read(format="stream")` → `commands.run` → `get_info` → `kill`. All green, with the access log showing every REST, envd, and RPC request:

  ```
  DEBUG pyqwest.access - HTTP Request: POST https://api.e2b.app/sandboxes "HTTP/2 201 Created"
  DEBUG pyqwest.access - HTTP Request: POST https://sandbox.e2b.app/files "HTTP/2 200 OK"
  DEBUG pyqwest.access - HTTP Request: POST https://sandbox.e2b.app/process.Process/Start "HTTP/2 200 OK"
  DEBUG pyqwest.access - HTTP Request: DELETE https://api.e2b.app/sandboxes/ickrgvawk1s9mn0963eta "HTTP/2 204 No Content"
  ```

## Usage examples

A proxy that needs a custom header no longer has to be rewritten as a URL — it just works:

```python
import httpx
from e2b import Sandbox

# Previously raised InvalidArgumentException
sbx = Sandbox.create(
    proxy=httpx.Proxy(
        "http://corp-proxy.internal:8030",
        auth=("user", "p@ss"),                 # sent as Proxy-Authorization
        headers={"X-Proxy-Token": "secret"},   # sent to the proxy
    )
)
```

Transport-level HTTP logs, replacing the httpcore records that the move off httpx transports removed:

```python
import logging

logging.basicConfig()
logging.getLogger("pyqwest.access").setLevel(logging.DEBUG)   # one line per request
logging.getLogger("pyqwest").setLevel(logging.DEBUG)          # + request lifecycle

Sandbox.create()
# DEBUG pyqwest.access - HTTP Request: POST https://api.e2b.app/sandboxes "HTTP/2 201 Created"
```

Both are opt-in and independent of the SDK's own `logger=` option, which is unchanged.

## Follow-ups

- **Rebase note**: #1602 (volume) and #1603 (template uploads) still call `proxy_to_url`; they need the rename when they rebase onto this.
- Once pyqwest 0.8.0 is out, this can either land as its own commit or be squashed into the stack it corrects — nothing here is a behavior change relative to shipped code, so the changeset edits in #1601's entry are corrections rather than a new user-visible delta.
- pyqwest 0.8 also brings `no_proxy` and per-scheme proxy routing, which `httpx.Proxy` can't express. Not exposed; would need a new SDK option and JS parity.

### DIFF
--- a/.changeset/python-pyqwest-0-8.md
+++ b/.changeset/python-pyqwest-0-8.md
@@ -1,0 +1,27 @@
+---
+"@e2b/python-sdk": minor
+---
+
+Require [`pyqwest`](https://pypi.org/project/pyqwest/) 0.8, which lets the SDK
+drop the workarounds it carried for what the HTTP stack could not express
+before:
+
+- `proxy` honors an `httpx.Proxy`'s credentials and custom headers as given,
+  instead of rejecting headers and folding the credentials into the proxy URL
+  userinfo. A per-proxy `ssl_context` remains unsupported.
+- Multipart uploads (`files.write`) no longer need the SDK to rewrap httpx's
+  request stream on the way to the transport.
+- Low-level HTTP logs are available again — the equivalent of the httpcore
+  records that moving off the httpx transports removed. pyqwest logs one line
+  per request on the `pyqwest.access` logger and request lifecycle records on
+  `pyqwest`, both at `DEBUG` and off unless enabled:
+
+  ```python
+  import logging
+
+  logging.basicConfig()
+  logging.getLogger("pyqwest.access").setLevel(logging.DEBUG)
+  # DEBUG pyqwest.access - HTTP Request: POST https://api.e2b.app/sandboxes "HTTP/2 201 Created"
+  ```
+
+  The SDK's own `logger` option is unchanged and independent of these.
--- a/.changeset/python-pyqwest-envd-api.md
+++ b/.changeset/python-pyqwest-envd-api.md
@@ -1,0 +1,27 @@
+---
+"@e2b/python-sdk": minor
+---
+
+Move the envd HTTP API client (sandbox file transfers, health checks) onto
+[`pyqwest`](https://pypi.org/project/pyqwest/) via its httpx-compatible
+transport adapter. envd RPC already runs on pyqwest through `connectrpc`, so
+all sandbox traffic now shares one HTTP stack built from the same transport
+pieces (with separate connection pools per use).
+
+The per-thread (sync) and per-loop (async) envd httpx clients are gone: the
+pyqwest transports are thread-safe and loop-independent, so a single client
+per module serves all threads and event loops.
+
+Timeout semantics through the adapter:
+
+- Streamed downloads (`files.read(format="stream")`): a `request_timeout`
+  set explicitly for the call is the deadline for the whole transfer — by
+  default the transfer is unbounded in total, as before. A stalled stream is
+  reclaimed by a 60-second idle read timeout that resets on every chunk.
+  `stream_idle_timeout` keeps working on the async client (applied per
+  read); the sync client cannot interrupt a blocking read, so it relies on
+  the transport-wide idle bound and now ignores the parameter.
+- Uploads: a buffered upload is bounded by `request_timeout` as a
+  whole-request deadline, and a streamed (file-like) upload carries no
+  client-side timeout (a stalled one is bounded server-side by envd's idle
+  read timeout) — both matching the JS SDK.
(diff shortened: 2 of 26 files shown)


mishushakov:
Superseded: the changes are folded into the stack they correct — the `httpx.Proxy` → `pyqwest.Proxy` mapping and the pyqwest 0.8 bump into #1601, the multipart-workaround removal and the envd renames into #1623, and the `proxy_to_config` rename into #1602/#1603. Tracking stays on [SDK-302](https://linear.app/e2b/issue/SDK-302).

PR #1635, opened 2026-08-03
fix(js-sdk): export the Git argument and status types

karpovantonme:
`Git.reset()`, `Git.restore()` and `Git.status()` are public, but the types naming their arguments and results are not reachable from the package entry point.

`src/index.ts` re-exports fifteen `Git*` types and omits `GitResetMode`, `GitResetOpts` and `GitRestoreOpts`. `GitStatusLabel` is worse off — `src/sandbox/git/index.ts` re-exports `GitBranches`, `GitConfigScope`, `GitFileStatus` and `GitStatus` from `./utils` but not `GitStatusLabel`, so it is unreachable from anywhere in the package, even though it is the type of `GitFileStatus.status`.

The practical effect: you can call the methods, but you cannot name what you pass them, so you cannot write a typed wrapper.

```ts
// before — all four fail
import type {
  GitResetMode,
  GitResetOpts,
  GitRestoreOpts,
  GitStatusLabel,
} from 'e2b'

// the workaround people end up with
type ResetMode = Parameters<Git['reset']>[0] extends { mode?: infer M } ? M : never
```

```ts
// after
import { Sandbox } from 'e2b'
import type { GitResetMode, GitResetOpts, GitStatusLabel } from 'e2b'

async function hardResetTo(sbx: Sandbox, repo: string, target: string) {
  const mode: GitResetMode = 'hard'
  const opts: GitResetOpts = { mode, target, cwd: repo }
  return sbx.git.reset(opts)
}

function isBlocking(status: GitStatusLabel) {
  return status === 'conflict' || status === 'deleted'
}
```

## On SDK parity

Python already exports `GitResetMode` (`e2b.GitResetMode`), so this brings JS up to it. The other three have no Python counterpart by design: the sync and async implementations take keyword arguments rather than option objects, so there is nothing shaped like `GitResetOpts`/`GitRestoreOpts`, and `GitFileStatus.status` is typed as a plain `str` there, so there is no `GitStatusLabel` either. Nothing to mirror on the Python side.

## Test

The missing re-export is invisible to `tsc --noEmit`, because `packages/js-sdk/tsconfig.json` includes only `src`, so a test that merely writes `import type … from '../src'` compiles and passes whether or not the export exists — I tried that first and it guarded nothing.

Instead the test reads the generated declarations, the way `tests/bundle/edgeCompat.test.ts` reads `dist/index.mjs`, and skips locally when the bundle has not been built. It fails on `main` and passes with the change.

`pnpm run format`, `pnpm run lint` and `pnpm run typecheck` are clean. Changeset included.

## One thing I did not touch

The same gap exists for a few non-Git types — `FilesystemListOpts`, `WatchOpts`, `PtyCreateOpts` and `PtyConnectOpts` are exported from their own modules but not from `src/index.ts`. I kept this PR to the Git surface. **If you want those too, say so in a comment and I will send a follow-up** — or tell me they are deliberately internal and I will leave them alone.


mishushakov:
it's fine to leave tests out here

karpovantonme:
Thanks! Happy to drop it if you would rather keep the diff to the two export lines. The only reason I kept it is that it does fail on main, so it would catch a re-export going missing again later, and `tsc --noEmit` cannot see that since tsconfig only includes `src`. Your call, say the word and I will remove it.

karpovantonme:
@cla-bot check

mishushakov:
@karpovantonme drop the tests

karpovantonme:
Dropped, thanks. The PR is down to the two export lines and the changeset

mishushakov:
Retargeted this at `fix/export-git-argument-types` and squash-merged it there, so your commits land as one commit with you as the author. Promoted to `main` in #1642 — thanks @karpovantonme!

valeriupredoi:
> Retargeted this at `fix/export-git-argument-types` and squash-merged it there, so your commits land as one commit with you as the author. Promoted to `main` in #1642 — thanks @karpovantonme!

you are aware that the author is an AI agent, right?

Issue #1641, opened 2026-08-04
[Bug]: JS SDK: one HTTP/2 GOAWAY (ENHANCE_YOUR_CALM) from the sandbox.e2b.app edge permanently breaks files.write process-wide; undici 8.8.0 never recovers the session


nischayv:
### Sandbox ID or Build ID

ilsr5jnqvfbe9ez17h0b8 (one of many affected)

### Environment

e2b (JS SDK) 2.38.0, Node 22.23.2 on Linux (Fly.io) and Node 23.6.0 on macOS 14. `loadUndici` selects the `undici8` optional dependency, which resolves to undici 8.8.0.

### Timestamp of the issue

First seen 2026-08-04 00:42 UTC; cause chains captured 2026-08-04 14:29 UTC.

### Frequency

Happens every time

### Expected behavior

When the server ends an HTTP/2 session (GOAWAY), the cost should be at most the in-flight requests. The SDK (or undici underneath it) should open a fresh connection for subsequent requests, and ideally retry streams the server refused, per RFC 9113 §8.7.

### Actual behavior

Once one `files.write` fails with:

```
TypeError: fetch failed
  cause: Error: Stream closed with error code NGHTTP2_ENHANCE_YOUR_CALM (ERR_HTTP2_STREAM_ERROR)
```

every subsequent `files.write` in the process fails with `NGHTTP2_INTERNAL_ERROR`, including immediate retries, across all sandboxes, until the process restarts. Filesystem calls that go through the connect-RPC transport (`files.makeDir`) keep working on the same sandbox at the same moment, because that transport uses a separate dispatcher.

This took our chat product down: since 2.25.0 all sandboxes' envd traffic shares the single origin `sandbox.e2b.app`, and undici multiplexes up to 100 streams per H2 session, so a long-lived server effectively rides one session. One GOAWAY on it breaks file writes process-wide.

### Issue reproduction

Hard to trigger from a fresh script (a new process has a fresh session with no accumulated history). Reliable in a long-lived Node server that:

1. Creates a pool of sandboxes and uploads a few MB to each with `files.write`
2. Lets them idle 1-3 minutes
3. On acquire, bursts ~15 parallel small `files.write` calls

After some process uptime, the edge in front of `sandbox.e2b.app` sends `ENHANCE_YOUR_CALM` (its rapid-reset / stream-abuse protection) and the failure mode above begins.

### Additional context

Root cause is a combination of a server behavior and a client bug:

[...]

PR #1645, opened 2026-08-06
chore(js-sdk): bump `undici` to ^7.29.0 and optional `undici8` to 8.10.0

mishushakov:
Bumps both undici dependencies in the js-sdk past the 2026-07-24 security advisories: the required `undici` from `^7.28.0` to `^7.29.0`, and the optional `undici8` (`npm:undici@…`) from 8.8.0 to 8.10.0. Both releases patch one High ([GHSA-4cwx-7wf7-3272](https://github.com/nodejs/undici/security/advisories/GHSA-4cwx-7wf7-3272), cache-control parsing / cross-user disclosure) and four Medium advisories, clearing the open Dependabot alerts for undici; 8.10.0 additionally fixes HTTP/2 request settling, refused-stream retries and GOAWAY handling, which we exercise because every dispatcher the SDK builds sets `allowH2: true`.

A root `pnpm.overrides` entry (`undici@>=7.0.0 <7.29.0`) is included because miniflare pins undici at exactly 7.28.0, which would otherwise keep a vulnerable copy in the lockfile; with it, the lockfile carries only 7.29.0 and 8.10.0. No code change was needed and there is no user-facing API change — 7.29.0 still requires Node `>=20.18.1` and 8.10.0 still requires `>=22.19.0`, matching the `UNDICI_8_MIN_NODE` gate in `packages/js-sdk/src/undici.ts`, so `getUndiciPackageCandidates()` picks the same package on the same Node versions.

`format`, `lint` and `typecheck` pass, `tests/undici.test.ts` is 9/9, and `test:cf` was run to confirm miniflare still boots on the overridden undici. The remaining vitest projects need `E2B_API_KEY`, which isn't available locally, so they're left to CI. A patch changeset for `e2b` is included.

Linear: [SDK-317](https://linear.app/e2b/issue/SDK-317/js-sdk-bump-optional-undici8-dependency-to-8100)


mishushakov:
closes https://github.com/e2b-dev/E2B/issues/1641 as well

tvi (approved):

PR #1647, opened 2026-08-06
Add periodic volume tests

DennisSFO:
## What
New scheduled workflow (`.github/workflows/periodic_tests.yml`) that runs the JS + Python SDK suites every 6h with the **volume integration tests enabled**, and pings Slack on failure.

## Why
The volume integration tests (standalone volume-content access, no sandbox) are gated behind `ENABLE_VOLUME_TESTS` and never run in CI. When that path broke in prod (`inc-2026-08-04`, invalid token audience), nothing caught it for ~24h. This adds a periodic backstop.

## How
- Adds an `enable-volume-tests` input to the reusable `js_sdk_tests.yml` / `python_sdk_tests.yml`; when set, an extra step exports `ENABLE_VOLUME_TESTS=1`.
- `periodic_tests.yml` calls both with the flag on. **Per-PR CI is unchanged** (input defaults to false).

## Notes (easy to tune / drop)
- Cadence: `19 */6 * * *` (every 6h) — adjust as needed.
- Slack failure alert posts to `monitoring-releases` via the existing `SLACK_WEBHOOK` — change the channel if there's a better one.

Ref: SDK-313

### DIFF
--- a/.github/workflows/js_sdk_tests.yml
+++ b/.github/workflows/js_sdk_tests.yml
@@ -12,6 +12,11 @@ on:
         required: false
         type: boolean
         default: false
+      enable-volume-tests:
+        description: 'Run the volume integration tests (sets ENABLE_VOLUME_TESTS)'
+        required: false
+        type: boolean
+        default: false
     secrets:
       E2B_API_KEY:
         required: true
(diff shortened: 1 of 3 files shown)


DennisSFO:
Agreed — prod-availability probing doesn't belong in the public repo's CI. Superseding with a Grafana Synthetic Monitoring check on the e2bfoxtrot stack (tracked in SDK-313). The enable-volume-tests input plumbing can be resurrected from this branch if a manual pre-release volume run is ever wanted.

PR #1657, opened 2026-08-10
feat(sdks): honor custom TLS trust with a `ca_bundle` / `caBundle` option

mishushakov:
Closes [SDK-320](https://linear.app/e2b/issue/SDK-320/python-sdk-honor-custom-tls-trust-verify-ssl-ca-bundle-on-the-pyqwest).

Every transport the SDKs build trusted only the default CA store, so a self-hosted deployment behind a private CA had no way to make the SDK validate against it. In Python the gap was silent: `AuthenticatedClient._verify_ssl` only ever configures httpx's *default* transport, and since the pyqwest migration every client is handed an explicit transport — so `verify=self._verify_ssl` has been dead for every REST call. JS never exposed a knob at all.

## What this adds

A `ca_bundle` / `caBundle` connection option — the path of a PEM file — resolving through the standard precedence (explicit option → `E2B_CA_BUNDLE` → unset), wired into **every** transport in both SDKs:

- REST control-plane API (sync + async)
- envd RPC and envd HTTP API, including the streaming transport
- volume content, including the streaming transports
- the template build-context upload, which builds its own client

The certificates are trusted **in addition to** the default store (`tls_include_system_certs` stays on in Python, the bundle is appended to `tls.rootCertificates` in JS), so a deployment that mixes a private CA with public ones keeps working.

## Usage

```python
from e2b import Sandbox

sandbox = Sandbox(ca_bundle="/etc/ssl/certs/internal-ca.pem")
sandbox.files.write("/tmp/hello.txt", "world")
```

```python
import os
os.environ["E2B_CA_BUNDLE"] = "/etc/ssl/certs/internal-ca.pem"
```

```ts
import { Sandbox } from 'e2b'

  caBundle: '/etc/ssl/certs/internal-ca.pem',
})
```

```bash
E2B_CA_BUNDLE=/etc/ssl/certs/internal-ca.pem node app.mjs
```

Volumes and template builds take the same option:

```python
volume = Volume.connect(volume_id, ca_bundle="/etc/ssl/certs/internal-ca.pem")
Template.build(template, "my-template", ca_bundle="/etc/ssl/certs/internal-ca.pem")
```

## How this interacts with the ambient trust store

The option gap is older than the pyqwest migration — `ConnectionConfig` never exposed a TLS-trust knob — but the migration did change which *ambient* trust sources apply, in the SDK's favor. Verified against a locally-signed HTTPS server, on Linux (container) and macOS:

| trust source | httpx (pre-migration) | pyqwest (now) |
| --- | --- | --- |
| `SSL_CERT_FILE` / `SSL_CERT_DIR`, Linux | honored | honored |
| `SSL_CERT_FILE`, macOS / Windows | honored | ignored — the native store has no equivalent |
| CA installed in the system trust store | rejected (httpx carries certifi) | honored |
| `ca_bundle` / `E2B_CA_BUNDLE` | — | honored, every platform (this PR) |

So self-hosted deployments on Linux keep whatever already worked, and gain the system store; `ca_bundle` is what makes it uniform across platforms and explicit in code rather than ambient.

## Nothing fails silently

- Python: `verify_ssl` on the SDK's API/volume client factories now raises `InvalidArgumentException` pointing at `ca_bundle` instead of being ignored. `ssl.SSLContext` and disabling verification have no pyqwest equivalent and are named as unsupported in the message.
- An unreadable file, or one holding no PEM certificate, raises where the connection is configured rather than as an unspecific certificate error per connection attempt (reqwest accepts an unparsable bundle at construction time and only fails later).
- JS: the option is Node-only — no other runtime lets the SDK configure TLS trust per connection — so setting it under workerd/Deno/Bun, or when `undici` can't be loaded, raises instead of quietly connecting with the trust it was meant to extend.

## Refactors that came with it

- Python: a `TransportConfig` NamedTuple now carries the proxy and the TLS trust, keys every transport cache, and produces the pyqwest keyword arguments in one place — the six transport construction sites no longer each spell out `tls_include_system_certs=True`. The volume clients reuse the shared `retrying_http_transport` instead of repeating its body.
- JS: fetcher factories take a `FetchTransportOpts` object (`{ proxy, caBundle }`) instead of a bare proxy string, and cache per combination.
- JS: the build-context upload builds a dispatcher of its own when — and only when — a CA bundle needs one, and refuses one outside Node. Every other build stays on undici's default agent, the path that carries the Content-Length framing presigned uploads require (#1243).

## Tests

- Python (`tests/test_ca_bundle.py`, new `trustme` dev dependency): an HTTPS server whose certificate only a test CA vouches for, so the assertions are real handshakes — the API, envd, volume, and template-upload paths (sync and async) connect with `ca_bundle` and fail without it. Plus precedence, propagation through `get_api_params`, cache keying, `verify_ssl` rejection, and the bundle-loading errors.
- JS (`tests/caBundle.test.ts`): the certificates reach the `Agent`'s `connect` and the `ProxyAgent`'s `requestTls`/`proxyTls` on top of the default roots, the error paths, and the non-Node rejection; plus `E2B_CA_BUNDLE` precedence and per-bundle fetcher caching.


mishushakov:
Addressed the automated review:

- **Codex, `tls.rootCertificates` is not the effective default** — right, and it would have silently narrowed trust for anyone already using `NODE_EXTRA_CA_CERTS`, `--use-system-ca` or `tls.setDefaultCACertificates`. The bundle is now appended to `tls.getCACertificates('default')` where it exists (Node >= 22.15), falling back to `rootCertificates` below that.
- **Bugbot, uncached upload dispatcher** — the upload fetcher is cached per CA bundle now, so a multi-file build reads the bundle once and keeps one Agent instead of one per file. A failed build is evicted so an unreadable bundle isn't replayed forever.
- **CodeQL, insecure TLS versions in the test server** — the test HTTPS server is pinned to TLS 1.2+.

Also narrowed the upload change: it only builds its own dispatcher when a CA bundle needs one, so every other build stays on undici's default agent — the path that carries the Content-Length framing presigned uploads require (#1243). Dropping it onto a dispatcher is what broke the Deno leg.

mishushakov:
no one asked for this and the branch drifted off

PR #1663, opened 2026-08-11
fix: preserve Retry-After on 429 rate-limit errors in Python and JS SDKs

### LINKED ISSUE #1325: Python and JS SDKs — 429 WAIT signal lost at exception boundary, no Retry-After
## Finding

## The consequence

## Expected behavior

## Related pattern

SnowingFox:
## Problem

A 429 response carrying a `Retry-After` header loses the wait signal at the exception boundary. The Python and JS SDKs map a 429 to `RateLimitException` / `RateLimitError` without consulting `Retry-After`, so callers cannot back off for the server-specified duration and fall back to local retry heuristics inside the provider's cooldown window.

## Root cause

The 429 error-mapping sites (`packages/python-sdk/e2b/api/__init__.py`, `packages/python-sdk/e2b/envd/api.py`, `packages/js-sdk/src/envd/api.ts`, plus the JS `packages/js-sdk/src/api/index.ts` counterpart) construct the rate-limit exception from the status code and message only. The `Retry-After` response header — the authoritative wait hint — is never read, so it is dropped before the caller ever sees the error.

## Fix

- Add a shared `parse_retry_after` / `parseRetryAfter` helper that turns a `Retry-After` header (delta-seconds or HTTP-date) into a wait time in seconds.
- Thread the header through each 429 mapping and store it on the exception: `RateLimitException(retry_after=..., retry_after_header=...)` in Python and `retryAfter` / `retryAfterHeader` on `RateLimitError` in JS.
- Append a `Retry after N seconds.` hint to the message so the wait is visible even where the structured field is not inspected.

## Test

- Python: `packages/python-sdk/tests/test_retry_after_rate_limit.py` asserts `retry_after`, `retry_after_header`, and the message hint for both the API and envd error paths.
- JS: new cases in `handleApiError.test.ts` and `handleEnvdApiError.test.ts` assert the same fields through mocked 429 responses.

Both fail on the current code (the fields are absent) and pass with this change.

Fixes #1325

### DIFF
--- a/.changeset/retry-after-preserved.md
+++ b/.changeset/retry-after-preserved.md
@@ -1,0 +1,10 @@
+---
+"e2b": patch
+"@e2b/python-sdk": patch
+---
+
+Preserve the `Retry-After` header on 429 rate-limit errors across the exception
+boundary. Both SDKs now parse the header and surface it as `retryAfter` /
+`retry_after` plus `retryAfterHeader` / `retry_after_header` on
+`RateLimitError` / `RateLimitException`, and append a wait hint to the error
+message, so callers can back off instead of retrying immediately.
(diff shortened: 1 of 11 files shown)


mishushakov:
Hey there, thanks for the contribution!
Our team is independently working on a built-in retry mechanism, so external contributions are not required.

PR #1664, opened 2026-08-12
Add trace ID extraction and appending to error messages

djeebus:
## Summary

This change adds trace ID extraction from HTTP response headers and appends them to error messages across the SDK and CLI. When API or envd requests fail, the error message now includes the trace ID (e.g., `(trace ID: abc123)`) so users can report it to E2B support for correlation with server-side traces.

## Usage examples

The trace ID shows up in existing error messages when the failed response carries a trace header, and is readable off the error itself.

**JS SDK**

```ts
import { Sandbox } from 'e2b'

try {
  await Sandbox.connect('already-dead-sandbox-id')
} catch (err) {
  console.error(err.message)
  // 404: sandbox not found (trace ID: 105445aa7843bc8bf206b12000100000)

  // or read it off the error, without parsing the message
  reportToSentry({ traceId: (err as SandboxError).traceId })
}
```

**Python SDK (sync and async)**

```python
from e2b import Sandbox, SandboxException

try:
    Sandbox.connect("already-dead-sandbox-id")
except SandboxException as e:
    print(e)
    # 404: sandbox not found (trace ID: 105445aa7843bc8bf206b12000100000)

    # or read it off the exception, without parsing the message
    report_to_sentry(trace_id=e.trace_id)
```

**CLI**

```
$ e2b sandbox logs already-dead-sandbox-id
Error while getting sandbox logs: [404] not found: sandbox not found (trace ID: 105445aa7843bc8bf206b12000100000)
```

## Key Changes

- **Trace ID parsing**: `packages/js-sdk/src/traceId.ts` (single implementation, exported as `extractTraceId` from the `e2b` package and reused by the CLI) and `e2b/trace_id.py` (exported as `extract_trace_id` for JS/Python parity):
  - Extract trace IDs from response headers in priority order: `X-Trace-ID` (direct), `X-Cloud-Trace-Context` (GCP edge), `X-Amzn-Trace-Id` (AWS edge)
  - Normalize AWS trace IDs from `Root=1-<8 hex>-<24 hex>` to the 32-hex form the server logs as `edge_trace_id`

- **Error classes take the trace ID as constructor context** and append it to the message: a trailing options object in JS (`new SandboxError(message, { traceId })`, uniform across every class) and a keyword-only argument in Python (`SandboxException(message, trace_id=...)`). Two option types are exported: `ErrorOpts` (`traceId`) and `ErrorOptsWithStackTrace` (adds `stackTrace`). Only the four classes actually handed a stack trace take the latter — `InvalidArgumentError`, `TemplateError`, `BuildError`, `FileUploadError` — so #1732's narrowing is preserved and passing `stackTrace` to any other class is a type error rather than silently ignored. `apiErrorFromCode` swaps in `AuthenticationError`/`RateLimitError` for 401/429 and deliberately does not forward the caller's frame to them: a bad key or a rate limit is a property of the request, not of the builder step that made it. `apiErrorFromCode` / `api_exception_from_code` and the envd error maps thread both through.

- **Breaking (JS, hence the `minor` changeset)**: the options object replaces the positional `stackTrace` parameter on the four classes that had one, so `new TemplateError(message, stackTrace)` becomes `new TemplateError(message, { stackTrace })` (likewise `InvalidArgumentError`, `BuildError`, `FileUploadError`). TypeScript callers get a compile error on the old form; plain-JS callers silently lose the stack trace. Python is unaffected — its exceptions already took a single positional argument.

- **Direct 404 raises**: `VolumeNotFoundError` / `VolumePathNotFoundError` / `SecretNotFoundError` (and their Python counterparts) are thrown at the call site before `handleApiError` sees the response — per T-60, 404 meaning is decided per call site — so each of those raise sites reads the header itself. 11 sites in JS (`volume/index.ts`, `secret.ts`) and 11 in each Python flavor (`volume_sync`/`volume_async`, `secret_sync`/`secret_async`).

- **Wiring**: `handleApiError` (JS), `handle_api_exception` (Python), `handleEnvdApiError` / `handle_envd_api_exception` / `ahandle_envd_api_exception` (envd HTTP), and the CLI's `handleE2BRequestError` read the response headers and pass the extracted ID into the error constructors.

- **Not covered — envd RPC (connect) paths**: `sandbox.commands` / `sandbox.files` RPC errors do not carry the trace ID, because Python's `connectrpc.ConnectError` exposes no response metadata, and the JS side is kept symmetric. This is where most in-sandbox failures surface, so the feature currently reaches HTTP API and envd HTTP (file transfer) errors only.

- **Test coverage**: header extraction and normalization edge cases, priority ordering, case-insensitivity, missing-header fallbacks, custom `errorClass`/`error_map` pass-through, and CLI parsing — in JS, Python (sync + async), and the CLI.

## Note on the server side

**Verified against production (2026-08-13): no failed `api.e2b.dev` or sandbox-edge response currently returns any trace header**, so this change is inert until the server emits one. The API handlers already compute the trace ID (`c.Set("traceID", traceID)` in infra) — a one-line `c.Header("X-Trace-ID", traceID)` there would activate the primary path. The GCP/AWS edge-header branches are speculative (those are request-side conventions); whether to keep them or trim to `X-Trace-ID`-only once the infra change lands is an open reviewer question for @djeebus.

https://claude.ai/code/session_0147b6bTi2gq4Yvm7VrLrxLD

### DIFF
--- a/packages/cli/src/utils/errors.ts
+++ b/packages/cli/src/utils/errors.ts
@@ -1,3 +1,4 @@
+import { extractTraceId } from 'e2b'
 import status from 'statuses'
 
 /**
@@ -16,13 +17,19 @@ type E2BResponse<TData> =
   | {
       data: TData
       error?: undefined
+      response?: { headers?: Headers }
     }
   | {
       data?: undefined
       error: E2BResponseError
+      response?: { headers?: Headers }
     }
 
-function throwE2BRequestError(error: E2BResponseError, errMsg?: string): never {
+function throwE2BRequestError(
+  error: E2BResponseError,
+  errMsg?: string,
+  traceId?: string
+): never {
   let message: string
   const code = error.code ?? 0
   switch (code) {
@@ -49,12 +56,12 @@ function throwE2BRequestError(error: E2BResponseError, errMsg?: string): never {
   throw new E2BRequestError(
     `${errMsg && `${errMsg}: `}[${code}] ${message && `${message}: `}${
       error.message ?? 'no message'
-    }`
+    }${traceId ? ` (trace ID: ${traceId})` : ''}`
   )
 }
 
 export function handleE2BRequestError(
-  res: { error: E2BResponseError },
+  res: { error: E2BResponseError; response?: { headers?: Headers } },
   errMsg?: string
 ): never
 export function handleE2BRequestError<TData>(
@@ -68,5 +75,5 @@ export function handleE2BRequestError(
   if (!res.error) {
     return
   }
-  throwE2BRequestError(res.error, errMsg)
+  throwE2BRequestError(res.error, errMsg, extractTraceId(res.response?.headers))
 }
--- a/packages/js-sdk/src/api/index.ts
+++ b/packages/js-sdk/src/api/index.ts
@@ -39,20 +39,22 @@ export function apiErrorFromCode(
   ) => Error = SandboxError,
   opts?: ErrorOptsWithStackTrace
 ): Error {
+  // An expired key or a rate limit is a property of the request, not of the
+  // builder step that happened to make it, so these two keep the frame where
+  // they were constructed even when the caller supplied one.
   if (code === 401) {
     const message = 'Unauthorized, please check your credentials.'
     return new AuthenticationError(
       content ? `${message} - ${content}` : message,
-      opts
+      { traceId: opts?.traceId }
     )
   }
 
   if (code === 429) {
     const message = 'Rate limit exceeded, please try again later'
-    return new RateLimitError(
-      content ? `${message} - ${content}` : message,
-      opts
-    )
+    return new RateLimitError(content ? `${message} - ${content}` : message, {
+      traceId: opts?.traceId,
+    })
   }
 
   return new errorClass(`${code}: ${content}`, opts)
--- a/packages/js-sdk/src/api/index.ts
+++ b/packages/js-sdk/src/api/index.ts
@@ -5,7 +5,9 @@ import { defaultHeaders } from './metadata'
 import { createApiFetch } from './http2'
 import { ConnectionConfig } from '../connectionConfig'
 import { AuthenticationError, RateLimitError, SandboxError } from '../errors'
+import type { ErrorOptsWithStackTrace } from '../errors'
 import { createApiLogger } from '../logs'
+import { extractTraceId } from '../traceId'
 
 const API_KEY_PATTERN = /^e2b_[0-9a-f]+$/
 const API_KEY_EXAMPLE = `e2b_${'0'.repeat(40)}`
@@ -33,32 +35,37 @@ export function apiErrorFromCode(
   content: unknown,
   errorClass: new (
     message: string,
-    stackTrace?: string
+    opts?: ErrorOptsWithStackTrace
   ) => Error = SandboxError,
-  stackTrace?: string
+  opts?: ErrorOptsWithStackTrace
 ): Error {
   if (code === 401) {
     const message = 'Unauthorized, please check your credentials.'
     return new AuthenticationError(
-      content ? `${message} - ${content}` : message
+      content ? `${message} - ${content}` : message,
+      opts
     )
   }
 
   if (code === 429) {
     const message = 'Rate limit exceeded, please try again later'
-    return new RateLimitError(content ? `${message} - ${content}` : message)
+    return new RateLimitError(
+      content ? `${message} - ${content}` : message,
+      opts
+    )
   }
 
-  return new errorClass(`${code}: ${content}`, stackTrace)
+  return new errorClass(`${code}: ${content}`, opts)
 }
 
 export function handleApiError(
   response: FetchResponse<any, any, any>,
   errorClass: new (
     message: string,
-    stackTrace?: string
+    opts?: ErrorOptsWithStackTrace
   ) => Error = SandboxError,
-  stackTrace?: string
+  // `traceId` is read off the response, so only `stackTrace` is caller-supplied
+  opts?: Pick<ErrorOptsWithStackTrace, 'stackTrace'>
 ): Error | undefined {
   // openapi-fetch leaves `error` undefined for non-2xx responses with
   // Content-Length: 0, so check the status instead
@@ -66,13 +73,15 @@ export function handleApiError(
     return
   }
 
+  const traceId = extractTraceId(response.response.headers)
+
   const status = response.response.status
   if (status === 401 || status === 429) {
     return apiErrorFromCode(
       status,
       response.error?.message ?? response.error,
       errorClass,
-      stackTrace
+      { ...opts, traceId }
     )
   }
 
@@ -80,7 +89,7 @@ export function handleApiError(
     status,
     response.error?.message || response.error || response.response.statusText,
     errorClass,
-    stackTrace
+    { ...opts, traceId }
   )
 }
 
--- a/packages/js-sdk/src/errors.ts
+++ b/packages/js-sdk/src/errors.ts
@@ -133,13 +133,10 @@ export class AuthenticationError extends Error {
    */
   readonly traceId?: string
 
-  constructor(message: string, opts?: ErrorOptsWithStackTrace) {
+  constructor(message: string, opts?: ErrorOpts) {
     super(formatMessage(message, opts?.traceId))
     this.name = 'AuthenticationError'
     this.traceId = opts?.traceId
-    if (opts?.stackTrace) {
-      this.stack = opts.stackTrace
-    }
   }
 }
 
@@ -180,12 +177,9 @@ export class TemplateError extends SandboxError {
  * Thrown when the API rate limit is exceeded.
  */
 export class RateLimitError extends SandboxError {
-  constructor(message: string, opts?: ErrorOptsWithStackTrace) {
+  constructor(message: string, opts?: ErrorOpts) {
     super(message, opts)
     this.name = 'RateLimitError'
-    if (opts?.stackTrace) {
-      this.stack = opts.stackTrace
-    }
   }
 }
 
--- a/packages/js-sdk/src/errors.ts
+++ b/packages/js-sdk/src/errors.ts
@@ -1,7 +1,35 @@
+/**
+ * Optional context attached to an SDK error.
+ */
+export interface ErrorOpts {
+  /**
+   * Trace ID of the failed request, appended to the message.
+   */
+  traceId?: string
+}
+
+/**
+ * Context for errors that can point somewhere other than where they were
+ * constructed.
+ */
+export interface ErrorOptsWithStackTrace extends ErrorOpts {
+  /**
+   * Stack trace to use instead of the one captured where the error is
+   * constructed, so the error points at the user's call site or at where the
+   * failure happened server-side.
+   */
+  stackTrace?: string
+}
+
+function formatMessage(message?: string, traceId?: string) {
+  return message && traceId ? `${message} (trace ID: ${traceId})` : message
+}
+
 // This is the message for the sandbox timeout error when the response code is 502/Unavailable
-export function formatSandboxTimeoutError(message: string) {
+export function formatSandboxTimeoutError(message: string, opts?: ErrorOpts) {
   return new TimeoutError(
-    `${message}: This error is likely due to sandbox timeout. You can modify the sandbox timeout by passing 'timeoutMs' when starting the sandbox or calling '.setTimeout' on the sandbox with the desired timeout.`
+    `${message}: This error is likely due to sandbox timeout. You can modify the sandbox timeout by passing 'timeoutMs' when starting the sandbox or calling '.setTimeout' on the sandbox with the desired timeout.`,
+    opts
   )
 }
 
@@ -11,9 +39,15 @@ export function formatSandboxTimeoutError(message: string) {
  * Thrown when general sandbox errors occur.
  */
 export class SandboxError extends Error {
-  constructor(message?: string) {
-    super(message)
+  /**
+   * Trace ID of the failed request, when the response carried one.
+   */
+  readonly traceId?: string
+
+  constructor(message?: string, opts?: ErrorOpts) {
+    super(formatMessage(message, opts?.traceId))
     this.name = 'SandboxError'
+    this.traceId = opts?.traceId
   }
 }
 
@@ -29,8 +63,8 @@ export class SandboxError extends Error {
  * The [unknown] error type is sometimes caused by the sandbox timeout when the request is not processed correctly.
  */
 export class TimeoutError extends SandboxError {
-  constructor(message: string) {
-    super(message)
+  constructor(message: string, opts?: ErrorOpts) {
+    super(message, opts)
     this.name = 'TimeoutError'
   }
 }
@@ -39,11 +73,11 @@ export class TimeoutError extends SandboxError {
  * Thrown when an invalid argument is provided.
  */
 export class InvalidArgumentError extends SandboxError {
-  constructor(message: string, stackTrace?: string) {
-    super(message)
+  constructor(message: string, opts?: ErrorOptsWithStackTrace) {
+    super(message, opts)
     this.name = 'InvalidArgumentError'
-    if (stackTrace) {
-      this.stack = stackTrace
+    if (opts?.stackTrace) {
+      this.stack = opts.stackTrace
     }
   }
 }
@@ -52,8 +86,8 @@ export class InvalidArgumentError extends SandboxError {
  * Thrown when there is not enough disk space.
  */
 export class NotEnoughSpaceError extends SandboxError {
-  constructor(message: string) {
-    super(message)
+  constructor(message: string, opts?: ErrorOpts) {
+    super(message, opts)
     this.name = 'NotEnoughSpaceError'
   }
 }
@@ -64,8 +98,8 @@ export class NotEnoughSpaceError extends SandboxError {
  * @deprecated Use {@link FileNotFoundError} or {@link SandboxNotFoundError} instead. This class will be removed in the next major version.
  */
 export class NotFoundError extends SandboxError {
-  constructor(message: string) {
-    super(message)
+  constructor(message: string, opts?: ErrorOpts) {
+    super(message, opts)
     this.name = 'NotFoundError'
   }
 }
@@ -74,8 +108,8 @@ export class NotFoundError extends SandboxError {
  * Thrown when a file or directory is not found inside a sandbox.
  */
 export class FileNotFoundError extends NotFoundError {
-  constructor(message: string) {
-    super(message)
+  constructor(message: string, opts?: ErrorOpts) {
+    super(message, opts)
     this.name = 'FileNotFoundError'
   }
 }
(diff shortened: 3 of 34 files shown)


packages/cli/src/utils/errors.ts
```diff
@@ -16,13 +16,64 @@ type E2BResponse<TData> =
...
     }
   | {
       data?: undefined
       error: E2BResponseError
+      response?: ResponseHeadersLike
     }
 
-function throwE2BRequestError(error: E2BResponseError, errMsg?: string): never {
+type ResponseHeadersLike = {
+  headers?: { get(name: string): string | null }
```
mishushakov: you can use built-in Headers type from undici?

packages/js-sdk/src/api/index.ts
```diff
@@ -66,21 +67,31 @@ export function handleApiError(
...
+
   const status = response.response.status
   if (status === 401 || status === 429) {
-    return apiErrorFromCode(
-      status,
-      response.error?.message ?? response.error,
-      errorClass,
-      stackTrace
+    return appendTraceId(
+      apiErrorFromCode(
```
mishushakov: maybe just change apiErrorFromCode instead

packages/js-sdk/src/traceId.ts
```diff
@@ -0,0 +1,67 @@
...
+ * server-side traces.
+ *
+ * Checked in order:
+ * 1. `X-Trace-ID` — used verbatim when present.
+ * 2. `X-Cloud-Trace-Context` (GCP edge) — `TRACE_ID/SPAN_ID;o=OPTIONS`,
+ *    the part before `/` is the trace ID.
+ * 3. `X-Amzn-Trace-Id` (AWS edge) — `Root=1-<8 hex>-<24 hex>;...`, the two
+ *    hex parts joined are the 32-hex trace ID the server logs.
+ */
+export function extractTraceId(
```
mishushakov: this is same code as in packages/cli/src/utils/errors.ts

packages/js-sdk/src/traceId.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  return undefined
+}
+
+/**
+ * Append the trace ID of the failed request to the error message, so users
+ * can include it when reporting the failure to E2B.
+ */
+export function appendTraceId<T extends Error | undefined>(
```
mishushakov: you could also modify E2B error classes to have trace id in message instead of this

packages/python-sdk/e2b/api/__init__.py
```diff
@@ -163,14 +164,24 @@ def handle_api_exception(
...
 
+    # A trace ID header on the failed response ties the request to
+    # server-side traces, so users can report it to E2B support
+    trace_id = extract_trace_id(getattr(e, "headers", None))
+
     message = body["message"] if "message" in body else None
     if message is None and e.status_code not in (401, 429):
-        return default_exception_class(f"{e.status_code}: {e.content}").with_traceback(
-            stack_trace
+        return append_trace_id(
```
mishushakov: feel free to update default_exception_class instead

mishushakov (changes requested):
reviewed, please remove duplicate method in JS SDK and you can modify the default Error classes to accept a trace field in the constructor that will append it to the message, which is cleaner than relying on helpers side-effects

djeebus:
Addressed the review in 0adc234:

- **Duplicate method removed** — `extractTraceId` now lives only in the JS SDK (`src/traceId.ts`), is exported from the package index, and the CLI imports it from `e2b`.
- **Trace ID via constructors** — the error/exception classes in both SDKs now take an optional trace ID in the constructor and append it to the message (`SandboxError(message, stackTrace, traceId)` / `SandboxException(message, trace_id)`); the `appendTraceId`/`append_trace_id` side-effect helpers are gone. `apiErrorFromCode` / `api_exception_from_code` and the envd error maps thread the trace ID through, per the inline suggestions.
- **Built-in `Headers` type** — the CLI (and the SDK helper) now type headers with the standard `Headers` instead of a custom shape.

Also added CLI test cases for the AWS raw-value fallback and header priority.

---
_Generated by [Claude Code](https://claude.ai/code)_

mishushakov:
@djeebus Second pass — I re-reviewed after 0adc234 and verified the build locally. All five of my earlier comments are addressed; details plus a few independent findings below.

## My earlier comments

| Comment | Status |
|---|---|
| Remove duplicate `extractTraceId` in the CLI | ✅ One copy in `js-sdk/src/traceId.ts`, exported from `index.ts`, CLI imports from `'e2b'`. |
| Modify default Error classes to take a trace field in the constructor | ✅ Both SDKs; `appendTraceId`/`append_trace_id` deleted. ⚠️ see **B** |
| `maybe just change apiErrorFromCode instead` | ✅ Threads `traceId`. Extraction staying in `handleApiError` is right — `apiErrorFromCode` also serves errors embedded in bodies (per-fork results) and has no response to read. |
| Use built-in `Headers` type | ✅ Global `Headers` in both. ⚠️ `traceId.ts:12` still runtime-guards `typeof headers.get !== 'function'` — leftover from the old `HeadersLike` shape and now unreachable, since the `!headers` check already covers every mock. Please drop it. |
| `feel free to update default_exception_class instead` | ✅ Base `SandboxException`/`AuthenticationException`/`BuildException`/`VolumeException` take `trace_id`; `api_exception_from_code` passes it through. |

## A. The feature is inert against production today, and 2 of its 3 header paths likely never fire

You flagged this in the description and asked someone to verify which header a real failed response returns. I checked against live prod:

| Request | Result | Response headers |
|---|---|---|
| `GET api.e2b.dev/sandboxes/does-not-exist` (no auth) | 401 | `content-type, date, content-length, via, alt-svc` |
| same, with a real API key | 400 | identical set |
| `GET 49999-<bogus>.e2b.app/files?path=/x` (sandbox edge) | 502 | identical set |

No trace header on any path. And `via: 1.1 google` matters here: `X-Cloud-Trace-Context` and `X-Amzn-Trace-Id` are *request*-side conventions — the LB injects them toward the backend, neither is echoed back on the response. So only `X-Trace-ID` has a plausible future, and it needs the infra one-liner you described.

**Suggestion:** land `c.Header("X-Trace-ID", traceID)` in infra first (or in parallel), and drop the GCP/AWS branches plus the AWS normalization regex and their tests unless we can name a deployment that actually returns them. That's roughly half the new code, and the JS/Python duplication of the AWS parser is the part most likely to drift.

## B. JS error-class arity is now inconsistent, and it's a live trap

`SandboxError(message, stackTrace?, traceId?)` but `AuthenticationError(message, traceId?)` / `GitAuthError(message, traceId?)` — trace ID in slot 2 vs slot 3. Meanwhile `apiErrorFromCode`/`handleApiError` declare `errorClass: new (message, stackTrace?, traceId?) => Error`, which `AuthenticationError` satisfies structurally (fewer params is assignable). Confirmed by construction:

```ts
apiErrorFromCode(500, 'boom', AuthenticationError, 'STACK', 'realtrace')
// typechecks clean; message === "500: boom (trace ID: STACK)"
```

No current call site hits it — `handleApiError` is only ever passed `BuildError`/`FileUploadError`/`TemplateError`/`VolumeError` — so it's latent rather than broken, but it's the kind of thing that bites the next person. Python has no equivalent problem; it's uniformly `(message, trace_id)`. Please give `AuthenticationError`/`GitAuthError` the same 3-arg shape.

Related: six call sites now read `new NotFoundError(message, undefined, traceId)`. An options bag — `new SandboxError(message, { stackTrace, traceId })` — removes the `undefined` padding *and* the arity trap in one move, and matches our convention that optional parameters go in a trailing options object. Still a constructor field, so it satisfies the original request.

## C. Unrelated change bundled in

`VolumeError` gained `this.stack = stackTrace` (`errors.ts:182-184`), which has nothing to do with trace IDs. It also introduces a JS↔Python divergence: Python's `VolumeException.__init__` is `(message, trace_id)` with no stack-trace param. No caller passes a stack trace to `VolumeError` (every site is `handleApiError(res, VolumeError)`), so it's dead today. Drop it, or mirror it in Python.

## D. `extractTraceId` is now public JS API with no Python counterpart

`export { extractTraceId }` in `src/index.ts` puts an internal parser on the public `e2b` surface, while `extract_trace_id` is absent from `e2b/__init__.py`. We treat JS/Python parity of the public surface as non-negotiable. The CLI's tsconfig maps only bare `e2b` → `../js-sdk/src` (no `e2b/*`), so index was the only zero-config route — that's fine, but it should be a deliberate call, and neither the changeset nor the PR description mentions the new export.

## E. Smaller notes

- **`SupportsApiErrorResponse` not extended** — `handle_api_exception` reaches for `getattr(e, "headers", None)`, but the generated `Response` dataclass does declare `headers: MutableMapping[str, str]`, and every real caller (including the hand-built volume ones) passes `httpx.Headers`. Adding `headers` to the Protocol would let `ty` check it.
- **Trace ID silently dropped on empty messages** — both `formatMessage` and `format_message_with_trace_id` require a truthy message, so `new SandboxError(undefined, undefined, 'abc')` loses it. Marginal.
- **Connect-RPC paths uncovered** — documented as intentional, but that's where most `sandbox.commands`/`sandbox.files` errors surface, so the feature won't reach the majority of in-sandbox failures even once infra sets the header. Worth calling out explicitly in the description.
- **No test** that a custom `errorClass` (`BuildError`/`VolumeError`) receives the trace ID through `handleApiError`.
- Python's `_FILESYSTEM_HTTP_ERROR_MAP` needed no change (it holds classes, not lambdas) while the JS one did — correct, just asymmetric.

## Verification I ran locally

- **Lint/format:** js-sdk oxlint clean; `ruff check` + `ruff format --check` clean (406 files); prettier `--check` clean on js-sdk `src`+`tests` and cli `src`.
- **Typecheck:** js-sdk `tsc --noEmit` clean, cli clean, python `ty check` clean.
- **Tests:** js-sdk `traceId` + `handleApiError` + `handleEnvdApiError` → 53 passed. CLI `vitest run` → 114 passed (only `template/create.test.ts` fails, which needs `E2B_API_KEY`; unrelated). Python `test_trace_id.py` → 20 passed.
- **CI:** the single red check, *Build Packages*, is a transient GitHub 503 while `uv` downloaded CPython 3.10.19 — unrelated to this PR, just needs a re-run.

## Verdict

The feedback was implemented faithfully and the mechanics are sound. Before merge I'd like **B** fixed (real trap, two-line change) and **C** dropped. **A** is the judgment call worth your input: as it stands this ships ~690 lines that produce no trace ID against any live E2B endpoint, and two-thirds of the parsing targets headers that GCP-fronted responses don't carry. Pairing it with the infra header change and cutting the edge-header branches would leave a much smaller, actually-exercised feature.

djeebus:
Second-pass items addressed in b7eeae2:

- **B (arity trap)** — `AuthenticationError`/`GitAuthError` now take the same `(message, stackTrace?, traceId?)` shape as the other classes, so a custom `errorClass` can no longer receive the stack trace in the trace ID slot. The construction you demonstrated now produces `500: boom (trace ID: realtrace)`.
- **C** — `VolumeError` no longer applies the stack trace (pre-existing behavior restored); the parameter remains only to fit the `errorClass` constructor shape, with a comment saying so.
- **Leftover guard** — the unreachable `typeof headers.get` check in `extractTraceId` is gone.
- **D (parity)** — `extract_trace_id` is now exported from the Python package (`e2b/__init__.py` + `__all__`), and both exports are mentioned in the changeset and PR description.
- **E** — `SupportsApiErrorResponse` declares `headers` (so `ty` checks the access, `getattr` dropped); added a `handleApiError` test for trace ID pass-through with a custom `errorClass` (`BuildError`); the PR description now calls out the RPC-path limitation explicitly.

**A** is @djeebus's call, so I've left the GCP/AWS branches in place for now and updated the PR description with your production findings (no trace header on any live path, `via: 1.1 google`, request-side conventions). If the decision is to trim to `X-Trace-ID`-only and pair with the infra `c.Header("X-Trace-ID", traceID)` change, that's a small follow-up commit — happy to do it on this branch once decided.

*Build Packages* failure on 0adc234 was the transient 503 you diagnosed; the new push re-runs CI.

---
_Generated by [Claude Code](https://claude.ai/code)_

mishushakov:
@djeebus Consolidating **Finding A** into one self-contained write-up, since it's the last open call on this PR and the surrounding decisions have now shifted what's at stake. Everything you need to decide is below — no need to dig back through the thread.

---

## What this PR does, and where Finding A sits

The PR reads a trace ID off the headers of a *failed* HTTP response and appends `(trace ID: ...)` to the error message, so a user can quote the ID in a support ticket and we can join it against server-side traces. It checks three headers in priority order: `X-Trace-ID`, then GCP's `X-Cloud-Trace-Context`, then AWS's `X-Amzn-Trace-Id` (normalising `Root=1-<8 hex>-<24 hex>` into the 32-hex form).

The rest of the review is settled — error classes now take a trailing options object in JS / keyword-only `trace_id` in Python, the parser lives in one place rather than being duplicated into the CLI, and it's exported as documented public API (`extractTraceId` / `extract_trace_id`) for callers who handle an E2B response themselves.

Finding A is the one thing left: **do the GCP and AWS branches belong in the shipped surface at all?**

## The evidence: none of the three headers comes back from any live E2B endpoint

| Request | Result | Response headers |
|---|---|---|
| `GET api.e2b.dev/sandboxes/does-not-exist` (no auth) | 401 | `content-type, date, content-length, via, alt-svc` |
| same, with a real API key | 400 | identical set |
| `GET 49999-<bogus>.e2b.app/files?path=/x` (sandbox edge) | 502 | identical set |

No `X-Trace-ID`, no `X-Cloud-Trace-Context`, no `X-Amzn-Trace-Id` on any of them. As shipped, the feature is inert against production — which matches the note in your own PR description that nothing in infra/belt writes a trace ID onto the response.

## Why the two edge branches are structurally unlikely, not just currently unused

`X-Cloud-Trace-Context` and `X-Amzn-Trace-Id` are **request-side** conventions. The load balancer injects them on the way *to* the backend; neither is echoed back on the response. `via: 1.1 google` confirms we sit behind a GCP LB, which doesn't echo its own header.

So `X-Trace-ID` is the only branch with a plausible future, and it needs the one-liner you identified — `c.Header("X-Trace-ID", traceID)` in infra, where `c.Set("traceID", ...)` already happens.

## Why this now matters more than when I first raised it

When the parser was internal, shipping unused branches was just dead weight — delete it later, nobody notices. Now that `extractTraceId` / `extract_trace_id` are exported with docs and a worked `@example`, **the three-header priority order and the AWS normalisation become a public contract.** Removing them later stops being a cleanup and becomes a deprecation cycle on documented behaviour, for parsing that has never once produced a value in production.

That's the crux: publishing it freezes it. Worth getting right in this PR rather than the next one.

## What trimming to `X-Trace-ID`-only would remove

- **`traceId.ts`**: the GCP `split('/')` block and the AWS `;`-field loop with `/^1-([0-9a-f]{8})-([0-9a-f]{24})$/i`, plus their doc lines — roughly half the file.
- **`trace_id.py`**: `import re`, `_AWS_ROOT_PATTERN`, the equivalent two blocks and doc lines — again about half.
- **14 test cases** across five files: 4 in `js-sdk/tests/traceId.test.ts`, 4 in `test_trace_id.py`, 4 in `cli/tests/utils/errors.test.ts`, plus one GCP case each in `handleApiError.test.ts` and `handleEnvdApiError.test.ts`.

It would also retire a divergence the two implementations already carry: JS does `field.trim().split('=')` and destructures `[key, value]`, so a `Root` value containing `=` gets truncated, while Python's `partition("=")` keeps it. Harmless for well-formed AWS values — but it's two implementations that already disagree on a malformed input, inside the code that cannot currently fire.

What's left after the trim is essentially one line:

```ts
headers?.get('x-trace-id')?.trim() || undefined
```

## The alternative I'd genuinely push for

The SDKs send **no** trace or request-ID header today — no `traceparent`, `x-request-id`, `x-cloud-trace-context` or `x-amzn-trace-id` anywhere in either SDK's source. Combined with your finding that our edge *parses the GCP/AWS headers off the request* and logs them as `edge_trace_id`, the design inverts:

> If the edge logs whatever trace header arrives on the **request**, the client can pick the ID and know it up front. Nothing needs to be echoed back.

That's strictly better for the use case this PR exists to serve:

1. **It covers failures with no response at all** — connection resets, request timeouts, DNS failures, `handleEnvdApiFetchError`, the Connect-RPC paths. Response-header extraction can never reach those, and they're a large share of what users actually open tickets about. `formatSandboxTimeoutError` and `formatRequestTimeoutError` fire with no response in hand, so under the current approach they stay trace-less permanently.
2. **No infra change needed**, versus waiting on the `c.Header` one-liner.
3. **One ID spans retries**, so a flaky request correlates as a single story.

Caveat: I haven't verified the edge-logging behaviour myself — that's from your reading of infra/belt, and I don't have those repos to hand. Worth confirming with whoever owns the edge before betting on it. If infra is on OpenTelemetry, `traceparent` (W3C Trace Context) is likely the right header to send rather than the GCP-specific one.

## The fair counter-argument

Self-hosted / BYOC. An E2B behind API Gateway can return `X-Amzn-Trace-Id` in some configurations, so the AWS branch isn't dead code *forever* — it's speculative for E2B Cloud specifically. If we keep it, I'd want that anchored to a named deployment rather than "cloud providers generally," because as it stands it's three header formats maintained in two languages serving zero live endpoints — and now doing so as public API.

## Recommendation

1. **Ship `X-Trace-ID` only**, landing the infra one-liner in the same window. Half the parser and 14 tests lighter, nothing speculative, and the public surface we freeze is one header instead of three.
2. **Follow-up:** SDK generates and sends the ID, attaching it to *every* error including timeouts and connection failures. Gated on confirming the edge logs it.
3. **What I'd avoid:** merging the three-header parser as public API, since it commits us to a documented priority order and an AWS normalisation that produce nothing today and that either option above would want to remove.

Happy to do the trim on this branch if you want option 1 — it's a small commit. One wrinkle if we go all the way to inlining the one-liner: JS is safe, since `Headers.get` is spec-guaranteed case-insensitive, but Python isn't — `handle_api_exception` types headers as `Optional[Mapping[str, str]]` and the tests pass plain dicts keyed `"X-Trace-ID"`. Every real caller passes `httpx.Headers`, which is case-insensitive, so I'd keep the small Python helper rather than inline it, to keep that contract in one place.

mishushakov:
@devin-ai-integration please check

packages/js-sdk/src/errors.ts
```diff
@@ -94,18 +128,27 @@ export class SandboxNotFoundError extends NotFoundError {
  * Thrown when authentication fails.
  */
 export class AuthenticationError extends Error {
```
mishushakov, line 130: this error has no callers that pass a stack trace ?

packages/js-sdk/src/errors.ts
```diff
@@ -137,21 +180,30 @@ export class TemplateError extends SandboxError {
  * Thrown when the API rate limit is exceeded.
  */
 export class RateLimitError extends SandboxError {
-  constructor(message: string) {
-    super(message)
+  constructor(message: string, opts?: ErrorOptsWithStackTrace) {
```
mishushakov: this error has no callers that pass stack trace ?

PR #1668, opened 2026-08-13
fix(python-sdk): keep get_transport accepting the deprecated http2 argument

beran-t:
## Summary

Any fresh `pip install e2b-code-interpreter` currently raises on the first `run_code()`:

```
TypeError: get_transport() got an unexpected keyword argument 'http2'
```

`e2b_code_interpreter/code_interpreter_sync.py:84` builds its streaming client with `get_transport(self.connection_config, http2=False)`. Moving the REST API client onto pyqwest (#1601) dropped that parameter — correctly, since ALPN negotiates the HTTP version now and the flag no longer selects anything. But `e2b-code-interpreter` declares `e2b>=2.26.0,<3.0.0`, so pip and poetry resolve the broken pair by default. `e2b` 2.37.0 works, 2.38.0 does not, and code-interpreter 2.9.0, 2.8.1, 2.8.0 and 2.7.0 all call it the same way — so there is no currently-publishable combination a user gets by default that works.

This accepts `http2` again on both the sync and async `get_transport`, documented as deprecated and ignored.

`get_transport` is not re-exported at the package root, so it reads as internal — but a sibling package we ship depends on it, and the declared version range says the combination is supported. Since the flag is genuinely inert, accepting and ignoring it costs nothing and unbreaks every already-published code-interpreter without users having to upgrade two packages in lockstep.

Repro, before and after:

```bash
python3 -m venv .venv && . .venv/bin/activate
pip install -q e2b-code-interpreter==2.9.0   # resolves e2b 2.38.0

E2B_API_KEY=... python -c "
from e2b_code_interpreter import Sandbox
Sandbox.create().run_code(\"print('hello')\")
"
# TypeError: get_transport() got an unexpected keyword argument 'http2'
```

Verified end to end: with this branch installed alongside an unmodified `e2b-code-interpreter==2.9.0`, that same script returns `['hello\n']` against live E2B. The existing `tests/test_api_client_transport.py` suite passes (26 tests), plus one added case covering the keyword, positional and omitted forms.

The real follow-up belongs in `e2b-dev/code-interpreter` rather than here: that call site is envd streaming traffic, so `get_envd_transport(config, for_streaming=True)` from #1623 is what it actually wants, and its `e2b` floor should move to `>=2.38` once it does. This PR is the compatibility half, so existing installs stop failing in the meantime. Happy to close it in favour of a code-interpreter-only fix if you would rather not carry the shim — the tradeoff is that everyone stays broken until they upgrade both packages.

### DIFF
--- a/.changeset/python-get-transport-http2-compat.md
+++ b/.changeset/python-get-transport-http2-compat.md
@@ -1,0 +1,12 @@
+---
+"@e2b/python-sdk": patch
+---
+
+Accept and ignore a deprecated `http2` argument on `get_transport` in both the
+sync and async REST API clients. The pyqwest move dropped the parameter because
+ALPN negotiates the HTTP version now, but every published
+`e2b-code-interpreter` calls `get_transport(config, http2=False)`, and its
+`e2b>=2.26.0,<3.0.0` range resolves straight to a version that no longer accepts
+it. The result was a `TypeError` on the first `run_code()` of any fresh
+`pip install e2b-code-interpreter`. The flag is inert, so it is accepted and
+ignored rather than removed.
--- a/packages/python-sdk/e2b/api/client_async/__init__.py
+++ b/packages/python-sdk/e2b/api/client_async/__init__.py
@@ -78,10 +78,20 @@ _transport_lock = threading.Lock()
 _transports: Dict[Optional[ProxyConfig], AsyncPyqwestTransport] = {}
 
 
-def get_transport(config: ConnectionConfig) -> AsyncPyqwestTransport:
+def get_transport(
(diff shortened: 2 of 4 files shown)


beran-t:
Good catch on the cleanup convention - fixed in 50c8d94. Wrapped the assertions in `try/finally` with both resets, mirroring `test_sync_get_transport_keyed_by_proxy`, and switched to the `test_api_key` fixture rather than a hardcoded key so it matches its neighbours too. Suite still 26 passed.

mishushakov:
The fix is up on the other repo, but let me check

mishushakov:
superseded by https://github.com/e2b-dev/E2B/pull/1671

PR #1670, opened 2026-08-13
fix(sdk): omit autoPause when unset

### LINKED ISSUE #1669: SDKs should omit autoPause when timeout lifecycle is not configured
## Problem

When a caller creates a sandbox without configuring `lifecycle.onTimeout` / `lifecycle["on_timeout"]`, both SDKs currently serialize `autoPause: false` in `POST /sandboxes`.

This collapses two distinct states at the API boundary:

1. the caller did not express an auto-pause preference;
2. the caller explicitly selected the `kill` timeout action.

Because both arrive as `autoPause: false`, the service cannot distinguish an explicit client choice from an omitted preference. That prevents the API from safely owning or evolving its default behavior without SDKs unintentionally overriding it.

## Expected behavior

- No timeout lifecycle configured: omit `autoPause` from the request.
- Explicit `onTimeout: 'kill'` / `on_timeout: 'kill'`: send `autoPause: false`.
- Explicit `onTimeout: 'pause'` / `on_timeout: 'pause'`: send `autoPause: true`.

This preserves current behavior when users explicitly choose a lifecycle action while restoring the semantic distinction for callers that leave it unspecified.

## Current implementation

The JS SDK defaults the local action to `kill` and always evaluates `action === 'pause'` into the payload. The Python sync and async SDKs similarly default to `kill` and always pass the comparison result to the generated request model.

I have a small patch ready for both SDKs, with request-level regression coverage for all three cases.

0YHR0:
## Summary

Omit `autoPause` from `POST /sandboxes` when callers leave the timeout lifecycle unspecified, while preserving explicit choices:

- unspecified lifecycle: omit `autoPause`
- explicit `kill`: send `autoPause: false`
- explicit `pause`: send `autoPause: true`

The change is applied consistently to the JavaScript SDK and both synchronous and asynchronous Python SDK paths.

## Why

The current SDK behavior eagerly converts its local default (`kill`) into `autoPause: false`. On the wire, that makes an omitted user preference indistinguishable from an explicit opt-out.

Preserving the distinction lets the API own its default behavior and evolve that default without SDK clients unintentionally overriding it. At the same time, users who explicitly select `kill` retain exactly the same request semantics as before.

This is intentionally a narrow serialization change: the SDK's local validation and explicit lifecycle behavior remain unchanged.

Closes #1669.

## Validation

- JS request-level tests for omitted, explicit `false`, and explicit `true` payloads
- Python sync and async request-construction tests for the same three states
- JS TypeScript typecheck
- JS Prettier check
- Python Ruff lint and format checks
- targeted Python test suite: 22 passed

### DIFF
--- a/.changeset/omit-default-auto-pause.md
+++ b/.changeset/omit-default-auto-pause.md
@@ -1,0 +1,6 @@
+---
+'e2b': patch
+'@e2b/python-sdk': patch
+---
+
+Omit `autoPause` from sandbox creation requests when the timeout lifecycle is not configured, while continuing to send `false` for an explicit `kill` action. This preserves the distinction between an unset client preference and an explicit opt-out so the API can apply its own default.
--- a/packages/js-sdk/src/sandbox/sandboxApi.ts
+++ b/packages/js-sdk/src/sandbox/sandboxApi.ts
@@ -1455,7 +1455,10 @@ export class SandboxApi {
       allow_internet_access: opts?.allowInternetAccess ?? true,
       network: buildNetworkBody(opts?.network, iam),
       iam,
-      autoPause: action === 'pause',
+      autoPause:
+        opts?.lifecycle?.onTimeout === undefined
+          ? undefined
+          : action === 'pause',
       autoPauseMemory: action === 'pause' ? keepMemory : undefined,
       autoResume: { enabled: autoResume },
     }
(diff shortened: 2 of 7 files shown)


0YHR0:
@cla-bot check

mishushakov:

Issue #328, opened 2024-03-07
Code Interpreter v2.0 


mlejva:
The current version of the [code interpreter SDK](https://e2b.dev/docs/sandbox/templates/premade#2-code-interpreter-sandbox) allows to run Python code but each run has its own separate context. That means that subsequent runs can't reference to variables, definitions, etc from past code execution runs.

This is suboptimal for a lot of Python use cases with LLMs. Especially GPT-3.5 and 4 expects it runs in a Jupyter Notebook environment. Even when ones tries to convince it otherwise. In practice, LLMs will generate code blocks which have references to previous code blocks. This becomes an issue if a user wants to execute each code block separately which often is the use case.

Related #326

im-calvin:
What's the status on this?

mlejva:
We have new code interpreter in works and some users are already using it. You can go ahead and start using it - https://github.com/e2b-dev/E2B/pull/326

It'll be soon released on the main branch.

If you give it a try, please share your feedback, it's tremendously helpful for us

mlejva:
@im-calvin let me know if you're asking about anything specific such as a feature that's blocking you

im-calvin:
Thanks for the update. I'm trying it out right now, I'm getting this error

```
RPC Error (-32000): error reading file '/root/.jupyter/kernel_id': open /root/.jupyter/kernel_id: no such file or directory
```

mlejva:
@im-calvin Can you share the code? It looks like you might be using custom sandbox template

im-calvin:
```
# e2b.Dockerfile

# You can use most of the Debian-based base images
FROM ubuntu:22.04

# Install the ffmpeg tool/
RUN apt update \
  && apt install -y ffmpeg
```

```
const sandbox = await CodeInterpreterV2.create({
  template: "custom-sandbox",
  apiKey: process.env["E2B_API_KEY"],
});

const result = await sandbox.execPython(code);
```

jakubno:
I added section to the PR about custom templates. If there's anything unclear, let me know :)
#326

mlejva:
Implemented in a separate SDK - https://github.com/e2b-dev/code-interpreter

PR #1671, opened 2026-08-13
fix(python-sdk): restore the http2 parameter on the transport factories

mishushakov:
The pyqwest migration in 2.38.0 dropped the `http2` parameter from `get_transport` and `get_envd_transport` (added deliberately in #1347, 2.32.0) and collapsed the transport cache key to the proxy alone, so `e2b-code-interpreter`'s Jupyter requests — `get_transport(self.connection_config, http2=False)` — now raise `TypeError: get_transport() got an unexpected keyword argument 'http2'`; that is already live, since `e2b = "^2.26.0"` resolves to 2.38.x, and it blocks the Python half of code-interpreter [#328](https://github.com/e2b-dev/code-interpreter/pull/328). pyqwest supports the capability, it just was not threaded through: this restores the pre-2.38.0 signature (so no consumer code changes, only an `e2b` floor bump) by passing `http_version=None if http2 else HTTPVersion.HTTP1` into the pyqwest transports, and puts the HTTP version back into both cache keys — without that, whichever caller asks second is handed a transport of the wrong version. The default is unchanged: `None` leaves TLS connections to ALPN (HTTP/2 against the E2B API) and uses HTTP/1 for plaintext, exactly as today. HTTP/1.1 is not cosmetic for the consumer — with HTTP/2 multiplexing, abandoning a request only resets its stream, so the code-interpreter server never sees the `http.disconnect` it needs to interrupt the kernel, while HTTP/1.1's one connection per request closes the connection and the server observes it.

## Usage

Both factories are internal (nothing is exported from `e2b/__init__.py`), so there is no public API change; consumers reaching into them get the 2.32.0 call back:

```python
from e2b.api.client_sync import get_transport, get_envd_transport

# Unchanged: ALPN negotiates the version (HTTP/2 against the E2B API).
transport = get_transport(config)

# Its own pool, pinned to HTTP/1.1, so a cancelled request closes the
# connection and the server observes the disconnect.
http1 = get_transport(config, http2=False)
envd_http1 = get_envd_transport(config, http2=False)
```

The async mirror (`e2b.api.client_async`) is identical.

## Tests

Six new cases in `packages/python-sdk/tests/test_api_client_transport.py`, sync and async: cache separation and identity across `http2` / proxy / `for_streaming`, the `http_version` value actually reaching the pyqwest transport (`[None, HTTP1, HTTP1]`), and a round trip proving the pinned transport works. The negotiated version can't be observed locally — the test echo server is plaintext, where both settings speak HTTP/1 — so it is asserted at the constructor, with the reason in a comment; it was verified by hand against `https://api.e2b.app/health` via the `pyqwest.access` logger, which shows `"HTTP/2 200 OK"` on the default and `"HTTP/1.1 200 OK"` with `http2=False` on both factories (and confirms `httpx.Response.http_version` is unreliable through the adapter — it reports HTTP/1.1 either way). 256 unit tests pass, plus `make lint`, `make typecheck` and `make format`. No JS change: its transport is an undici-dispatcher `fetch` with no HTTP-version knob, and the JS half of code-interpreter #328 is a clean bump.

Closes [SDK-335](https://linear.app/e2b/issue/SDK-335/python-sdk-get-transport-lost-its-http2-parameter-in-2380-breaking-e2b)

### DIFF
--- a/.changeset/python-sdk-restore-http2-parameter.md
+++ b/.changeset/python-sdk-restore-http2-parameter.md
@@ -1,0 +1,5 @@
+---
+'@e2b/python-sdk': patch
+---
+
+Restore the `http2` parameter on `get_transport` and `get_envd_transport`, which the pyqwest migration dropped in 2.38.0. `http2=False` again returns a transport pinned to HTTP/1.1, on its own connection pool.
--- a/packages/python-sdk/e2b/api/client_async/__init__.py
+++ b/packages/python-sdk/e2b/api/client_async/__init__.py
@@ -3,7 +3,7 @@ from typing import Dict, Optional, Tuple, Union
 
 import httpx
 
-from pyqwest import HTTPTransport, Request, Response
+from pyqwest import HTTPTransport, HTTPVersion, Request, Response
 from pyqwest.httpx import AsyncPyqwestTransport
 from pyqwest.middleware.retry import RetryTransport
 
@@ -40,7 +40,9 @@ class ConnectionRetryTransport(RetryTransport):
 
 
 def retrying_http_transport(
-    proxy: Optional[ProxyConfig], read_timeout: Optional[float] = None
+    proxy: Optional[ProxyConfig],
+    read_timeout: Optional[float] = None,
+    http2: bool = True,
 ) -> ConnectionRetryTransport:
     """A fresh pyqwest transport (= its own connection pool) with the SDK's
     shared tuning — system CA certs (without which TLS through an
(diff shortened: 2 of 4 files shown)


beran-t:
One note from the architect harness, if you wanna add it.

Nothing guards against re-dropping the parameter. The break happened because a de-facto-public signature changed silently, and the new tests all pass http2= as a keyword — so moving it behind * or renaming it keeps the suite green and breaks e2b-code-interpreter again. Cheapest fix: a one-line comment on both factories naming the external consumer (e2b-code-interpreter reaches into this — #1347 / SDK-335), optionally an inspect.signature assertion. The comment alone would likely have prevented the original drop.

beran-t (approved):

PR #1334, opened 2026-05-19
fix: preserve Retry-After on rate limit errors

JGoP-L:
## Summary
- Preserve `Retry-After` from 429 responses on JS `RateLimitError.retryAfter`.
- Preserve `Retry-After` from 429 responses on Python `RateLimitException.retry_after`.
- Map envd 429 responses to rate-limit error types consistently.
- Add regression tests for API and envd 429 handling.

## Test plan
- [x] `pnpm --filter e2b exec vitest run tests/api/handleApiError.test.ts tests/envd/handleEnvdApiError.test.ts`
- [x] `pnpm --filter e2b lint`
- [x] `pnpm --filter e2b typecheck`
- [x] `python -m pytest tests/test_rate_limit_exceptions.py`
- [x] `ruff check`
- [x] `ty check`

Fixes #1325

PR #1674, opened 2026-08-14
fix(sdk): preserve Retry-After on rate-limit errors

### LINKED ISSUE #1325: Python and JS SDKs — 429 WAIT signal lost at exception boundary, no Retry-After

keshavkrishna:
## Problem

Fixes #1325.

`RateLimitException` (Python) / `RateLimitError` (JS) is raised on HTTP
429 from four call sites across both SDKs — the main API client and the
envd API client, in each language — but none of them capture the
response's `Retry-After` header. Callers have no signal for how long to
back off, so they either hardcode a guess or immediately retry, making
rate limiting worse.

Three prior attempts at this exact fix (#1333, #1334, #1353) were closed
without merging, but not for technical reasons — two never signed the
CLA, and the third was closed as "stale for now" with an invitation to
reopen. This PR follows the same reviewed design.

## Changes

- Add a shared `parse_retry_after` (Python, `e2b/exceptions.py`) /
  `parseRetryAfter` (JS, `src/utils.ts`) helper that parses a
  `Retry-After` header per RFC 9110 §10.2.3 — either delta-seconds or an
  HTTP-date. Returns `None`/`undefined` for a missing, negative, or
  otherwise unparsable value.
- Thread it through the main API client (`e2b/api/__init__.py`,
  `src/api/index.ts`) and the envd HTTP API client (`e2b/envd/api.py`,
  `src/envd/api.ts`), attaching the parsed value to the rate-limit
  exception as `retry_after` (Python) / `retryAfter` (JS).
- The `error_map`/`errorMap` override parameter types are unchanged —
  429 is special-cased before consulting the map, matching the existing
  pattern already used for 401.
- The Connect-RPC envd path (`e2b/envd/rpc.py`) is intentionally left
  untouched, since it's unclear whether Connect-RPC errors carry an
  equivalent header at that layer.
- Unit tests covering the parser (delta-seconds, HTTP-date, negative/
  invalid values, missing header) and the full path through both API
  clients in both SDKs.

## Usage

```python
from e2b import RateLimitException

try:
    sandbox = Sandbox.create()
except RateLimitException as e:
    if e.retry_after is not None:
        time.sleep(e.retry_after)
        sandbox = Sandbox.create()
```

```ts
import { RateLimitError, Sandbox } from 'e2b'

try {
} catch (err) {
  if (err instanceof RateLimitError && err.retryAfter !== undefined) {
    await new Promise((r) => setTimeout(r, err.retryAfter! * 1000))
  }
}
```

## Validation

- `pytest` (Python) and `vitest` (JS) unit suites pass, including 33
  new/extended test cases covering the parser and both SDKs' 429
  mappings.
- Confirmed no regressions: JS unit project shows the same 219
  pre-existing (environment-dependent, `E2B_API_KEY`-gated) failures
  before and after this change, plus 10 new passing tests.
- `ruff`, `ty`, `oxlint`, `tsc --noEmit`, and `prettier --check` all
  clean on touched files.


keshavkrishna:
@cla-bot check

PR #1675, opened 2026-08-15
fix(js-sdk): release the inflight concurrency slot on body end, not h…

### LINKED ISSUE #1666: Inflight Concurrency Cap Bypass on Streaming Bodies
Currently in the JS SDK, the `limitConcurrency` wrapper around `fetch` releases the semaphore early (on headers) instead of when the response body is fully consumed. This defeats the connection accounting for streaming requests (e.g., logs, command execution) and can exhaust HTTP/2 streams (`ERR_HTTP2_TOO_MANY_CONCURRENT_STREAMS`). There is a clear TODO for this in `js-sdk/src/api/inflight.ts`.

keshavkrishna:
## Problem

Fixes #1666.

`limitConcurrency` (`js-sdk/src/api/inflight.ts`) releases its semaphore slot as soon as `fetcher` resolves with the response headers, not once the body is fully consumed:

```ts
const release = await sem.acquire(signal)
try {
  return await fetcher(input, init)
} finally {
  release()
}
```

For streaming responses (logs, command output, file/volume transfers) this lets real concurrency exceed `inflightLimit` while bodies are still in flight, risking `ERR_HTTP2_TOO_MANY_CONCURRENT_STREAMS` under load. There was already a `TODO` for this.

## Changes

- Slot now releases on body end, error, or cancel, not on headers, via a passthrough `ReadableStream` (`releaseOnBodyEnd`).
- No-body responses (204/304, HEAD) release immediately.
- Reconstructing the response loses `.url`/`.redirected`/`.type` (no way to set them via the constructor). Confirmed nothing in this SDK reads them, and this wrapper is internal only.
- 5 new tests covering hold-during-stream, release on end/error/cancel, and the no-body path.

## Verified

- New "holds slot while streaming" test fails on old code, passes on the fix, confirmed by reverting and re-running.
- Real consumption paths, not just the test helper: 500KB chunked binary round-trips byte-exact via `.arrayBuffer()`, `.blob()`/`.json()` both work, 3 concurrent streams against a cap of 2 never exceed 2 mid-stream.
- Also used by `envd/http2.ts` (file read/write) and `volume/client.ts`, not just the main API client.
- Full JS unit project: same pre-existing failure count before/after.
- `oxlint` + `tsc --noEmit` clean.

### DIFF
--- a/packages/js-sdk/src/api/inflight.ts
+++ b/packages/js-sdk/src/api/inflight.ts
@@ -45,17 +45,63 @@ function abortReason(signal: AbortSignal | undefined): unknown {
 }
 
 /**
+ * Wrap a response body so `release` fires exactly once: when the body is
+ * fully drained, when reading it errors, or when the consumer cancels it —
+ * never on response headers alone. Reconstructing the response this way
+ * loses `.url`, `.redirected`, and `.type` (the `Response` constructor has
+ * no way to set them, and `Body` methods like `.text()`/`.json()` read an
+ * internal body slot rather than the public `.body` getter, so patching
+ * just that getter would not intercept them). Nothing in this SDK reads
+ * those three properties off a dispatched response, and this wrapper is
+ * internal to the fetch pipeline, never part of the public API.
+ */
+function releaseOnBodyEnd(
+  body: ReadableStream<Uint8Array>,
+  release: () => void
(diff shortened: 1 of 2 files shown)


mishushakov:

PR #1678, opened 2026-08-16
fix(sdk): omit autoResume when unset

### LINKED ISSUE #1677: SDKs should omit autoResume when it is not configured
## Problem

When a caller creates a sandbox without configuring `lifecycle.autoResume` / `lifecycle["auto_resume"]`, both SDKs currently serialize the following value in `POST /sandboxes`:

```json
{
  "autoResume": {
    "enabled": false
  }
}
```

This collapses two distinct states at the API boundary:

1. the caller did not express an auto-resume preference;
2. the caller explicitly disabled auto-resume.

Because both arrive as `enabled: false`, the service cannot distinguish an explicit client choice from an omitted preference. That prevents the API from safely owning or evolving its default behavior without SDKs unintentionally overriding it.

This is analogous to #1669, but applies to the optional nested `autoResume` configuration.

## Expected behavior

- No `autoResume` configuration: omit the entire `autoResume` field.
- Explicit `autoResume: false` / `auto_resume: False`: send `{"autoResume":{"enabled":false}}`.
- Explicit `autoResume: true` / `auto_resume: True`: send `{"autoResume":{"enabled":true}}`.

This preserves existing behavior for explicit choices while restoring the semantic distinction for callers that leave the option unspecified.

## Scope

The same request construction behavior exists in:

- JavaScript SDK
- Python synchronous SDK
- Python asynchronous SDK

I have a focused patch ready with request-level regression tests covering the omitted, explicit false, and explicit true cases in all implementations.

0YHR0:
## Summary

Omit `autoResume` from `POST /sandboxes` when callers leave it unspecified, while preserving both explicit choices:

- unspecified: omit `autoResume`
- explicit `false`: send `{"autoResume":{"enabled":false}}`
- explicit `true`: send `{"autoResume":{"enabled":true}}`

The change is applied consistently to the JavaScript SDK and the synchronous and asynchronous Python SDK paths.

## Why

The SDKs currently default the local value to `false` and always serialize it. On the wire, that makes an omitted user preference indistinguishable from an explicit opt-out.

Preserving the distinction allows the API to own and safely evolve its default without SDK clients unintentionally overriding it. Explicit `false` and `true` configurations retain their existing request semantics and validation.

This is analogous to #1669, but applies to the optional nested `autoResume` configuration.

Closes #1677.

## Usage

No application changes are required. Existing explicit configurations continue to work:

```ts
await Sandbox.create({
  lifecycle: { onTimeout: 'pause', autoResume: true },
})
```

```python
Sandbox.create(
)
```

Callers that omit `autoResume` now leave the preference unset at the API boundary.

## Validation

- JavaScript request-level tests for omitted, explicit `false`, and explicit `true`
- Python sync and async request-construction tests for the same three states
- JavaScript TypeScript typecheck, Oxlint, and targeted Prettier check
- Python Ruff lint/format and `ty` typecheck
- 3 JavaScript tests passed
- 8 targeted Python tests passed

The repository-wide JavaScript Prettier check reports existing Windows line-ending differences across 157 untouched files; all files changed by this PR pass the targeted Prettier check.

### DIFF
--- a/.changeset/omit-default-auto-resume.md
+++ b/.changeset/omit-default-auto-resume.md
@@ -1,0 +1,6 @@
+---
+'e2b': patch
+'@e2b/python-sdk': patch
+---
+
+Omit `autoResume` from sandbox creation requests when callers do not configure it, while preserving explicit `false` and `true` values. This lets the API distinguish an unset preference from an explicit opt-out and apply its own default.
--- a/packages/js-sdk/src/sandbox/sandboxApi.ts
+++ b/packages/js-sdk/src/sandbox/sandboxApi.ts
@@ -1457,7 +1457,10 @@ export class SandboxApi {
       iam,
       autoPause: action === 'pause',
       autoPauseMemory: action === 'pause' ? keepMemory : undefined,
-      autoResume: { enabled: autoResume },
+      autoResume:
+        opts?.lifecycle?.autoResume === undefined
+          ? undefined
+          : { enabled: autoResume },
     }
 
     if (opts?.volumeMounts) {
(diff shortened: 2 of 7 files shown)


mishushakov:

PR #1683, opened 2026-08-18
chore(ci): remove the Dependabot changeset workflow

mishushakov:
Dependabot is being turned off for this repo, so `.github/workflows/dependabot_changeset.yml` — which committed a `patch` changeset to every Dependabot PR that touched a released package's direct production dependencies — has nothing left to run on, and it goes away along with the paragraph describing it in `.changeset/README.md`. No other file referenced the workflow, and nothing about hand-written changesets changes: `npx changeset` is still the way to add one.

Two follow-ups live outside this diff. The repo has no `.github/dependabot.yml` (it never did), so the bumps we've been getting came from GitHub's **Dependabot security updates** toggle — that has to be switched off in *Settings → Advanced Security* for the PRs to actually stop. And if "Dependabot Changeset" is listed as a required check in branch protection, it needs removing there or PRs will wait on a check that no longer runs; the `VERSION_BUMPER_APPID` / `VERSION_BUMPER_SECRET` credentials this workflow used are worth double-checking against the other workflows before revoking.


tomassrnka (approved):
ok

PR #1697, opened 2026-08-19
docs: fix Dynatrace MCP server description

Excelius-Wang:
## Summary

- Fix `brining` to `bringing` in the Dynatrace MCP server schema.
- Sync the generated TypeScript and Python SDK types.
- Add patch changesets for both SDK packages.

## Testing

- `python -m json.tool spec/mcp-server.json`
- Regenerated the Python MCP types with `datamodel-code-generator==0.64.0`
- Verified that the source description matches the generated TypeScript and Python files
- `git diff --check`


mishushakov:
We auto-generate these, no need to interfere here.

PR #974, opened 2025-10-20
Add custom MCP to SDK

Code42Cate:
This adds (experimantal) support to add arbitrary MCP server from Github.
```python
sbx = await AsyncSandbox.beta_create(timeout=600, mcp={
    "github/modelcontextprotocol/servers": {
        "install_cmd": "npm install",
        "run_cmd": "sudo npx -y @modelcontextprotocol/server-filesystem /root",
    },
})
```

or in TS with better typing:

```typescript
const sandbox = await Sandbox.betaCreate({
    mcp: {
        duckduckgo: {},        
        'github/modelcontextprotocol/servers': {
            installCmd: 'npm install',
            runCmd: `sudo npx -y @modelcontextprotocol/server-filesystem /root`,
        },
    },
    timeoutMs: 600_000,
});
```

---

> [!NOTE]
> Adds beta support to launch custom MCP servers (including GitHub-defined ones) and updates MCP schemas/types; also switches default MCP template to `mcp-gateway`.
> 
> - **SDK (JS & Python)**:
>   - Add beta support for `mcp` config, including dynamic GitHub-based servers with custom `runCmd/installCmd/envs` and automatic `mcp-gateway` startup with token.
>   - Send `UNSET`/omit when `mcp` not provided to avoid API noise.
>   - Change default MCP template from `mcp-gateway-v0-2` to `mcp-gateway`.
> - **Types/Schema**:
>   - Extend `McpServer` typings (JS `mcp.d.ts`, Python type hints) to include many servers and descriptive metadata.
>   - Introduce GitHub MCP types (TS `GitHubMcpServer`, Python `GitHubMcpServerConfig`) and export in SDK surfaces.
>   - Regenerate and enrich `spec/mcp-server.json` with titles/descriptions and new server entries.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f3139b8fc2f1246c72734d8048ba277a13e3949d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #1700, opened 2026-08-19
feat(sdk): refresh the MCP server schema from Docker's MCP catalog

mishushakov:
Fixes [SDK-340](https://linear.app/e2b/issue/SDK-340/refresh-specmcp-serverjson-from-dockers-mcp-catalog-stale-since-oct).

`spec/mcp-server.json` lists the MCP servers a sandbox can run and the options each one takes, and the SDKs generate their `McpServer` types from it. It hadn't been refreshed since it was written in [#974](https://github.com/e2b-dev/E2B/pull/974) (October 2025), so it was describing a catalog that has moved a long way — which is also what [#1697](https://github.com/e2b-dev/E2B/pull/1697) ran into when it tried to fix a description by hand.

This regenerates it, and adds the projection from the catalog to the schema as a script so the next refresh is a command instead of a hand edit:

```sh
pnpm generate:mcp-spec  # spec/mcp-server.json, then the SDK types
```

## What changed in the schema

| | |
| --- | --- |
| Servers added | 52 |
| Servers removed | 6 |
| Titles rewritten | 61 |
| Descriptions rewritten | 10 |
| Options changed | 4 |
| Servers untouched | 145 |

New servers include `curl`, `docker`, `ffmpeg`, `n8n`, `neo4j`, `temporal`, `proxmox`, `testkube`, `zen`, and 26 AWS servers:

```ts
import { Sandbox } from 'e2b'

  mcp: {
    docker: {},
    ffmpeg: {},
    n8n: { apiUrl: 'https://n8n.example.com/api/v1', apiKey: process.env.N8N_API_KEY! },
  },
})
```

```python
import os

from e2b import Sandbox

sandbox = Sandbox.create(
    mcp={
        "docker": {},
        "ffmpeg": {},
        "n8n": {"apiUrl": "https://n8n.example.com/api/v1", "apiKey": os.environ["N8N_API_KEY"]},
    },
)
```

Titles are what the generated interface names come from, so renames upstream show up in the generated types — `AirtableMCPServer` became `Airtable`, for example. Only `McpServer` itself is exported from either SDK, so no public name changes.

## What stops type-checking

Six servers the catalog no longer publishes are gone from the type: `cdataConnectcloud`, `flexprice`, `postgres`, `root`, `tembo`, and `triplewhale`. `awsDiagram` and `context7` took no options before and now require one (`outputDir` and `apiKey`), so `awsDiagram: {}` and `context7: {}` stop compiling — verified with `@ts-expect-error` probes. `onlyofficeDocspace` is down to `baseUrl` and `docspaceApiKey`, and `neo4jCypher` gained `schemaSampleSize`. Configuration is still handed to the gateway as written, so a dropped server can be kept by casting past the type — whether it starts is up to the gateway.

That set of breaks is why this is `minor` on both packages rather than `patch`; it's a judgement call worth making deliberately, since the public `McpServer` type narrows.

## How the generator was checked

The naming in the committed schema isn't a plain copy of the catalog's own keys, so the script reproduces the rules the schema already followed: secrets are named after their environment variable minus the server's own name where the catalog repeats it (`ATLAN_API_KEY` on `atlan` is `apiKey`), compared against the catalog name as spelled — so a hyphenated name never matches the underscored variable and keeps the prefix (`AIRTABLE_API_KEY` on `airtable-mcp-server` is `airtableApiKey`, which is 33 properties across 23 servers). Parameters are named after their config key with nested objects flattened, and servers after their catalog name minus the "mcp server" wording.

To confirm those rules are the ones the schema was built with, I rebuilt the catalog as it stood the day the schema was last written (from `docker/mcp-registry` at that commit) and ran the generator against it. All 195 servers that exist in both come out with identical property names, identical property schemas, and identical `required` sets. Against the schema this replaces, the 216 carried-over servers keep their property names and `required` sets except for the four servers whose catalog config genuinely changed.

So the diff here is the catalog moving, not a renaming pass. `packages/python-sdk/tests/test_mcp_spec_generator.py` pins the rules — 28 cases, no network — because that suite is the only test runner above the repository root.

## Two inherited behaviours worth a decision

Both predate this PR and are reproduced rather than introduced, so nothing regresses either way, and relaxing either one is compile-compatible (making a required property optional never breaks an existing object literal).

- **A parameters block that declares no `required` list is read as requiring everything.** JSON Schema reads an absent list as requiring nothing. 78 of 138 config blocks omit it, which forces 187 properties across 78 servers, and 162 of the 207 servers with properties end up requiring all of them. Combined with "every secret is required", that gives `zen` a type demanding all four competing vendor keys and `awsApi` one demanding a named profile *and* static keys *and* a session token. Across the 216 carried-over servers 213 `required` sets are byte-identical to the schema this replaces, so this is established behaviour, but it is a one-line change here if you want the JSON Schema reading instead. (Bugbot and Claude both flagged this; their examples are partly misattributed — `zen` has no parameters at all, so its four keys come from the secrets rule, not this one.)
- **`playwright` resolves to ExecuteAutomation's server, and Docker's own is unreachable.** The catalog publishes two Playwright servers and they collide on the option name; the longer catalog name wins, which is what the previous schema shipped. Changing the winner would silently repoint anyone already configuring `playwright`, so I left it and made the drop visible instead — the script now names every entry that loses a collision. Same shape for `apify` and `needle`, where the dropped entry is only a remote stub.

## Notes

- Unlike `openapi.yml` and the envd specs, this schema has no upstream commit to pin — it tracks whatever the catalog publishes today. `make codegen` therefore leaves it alone and refreshing is a deliberate step, which keeps the generated-files CI check deterministic (it passes here, so the committed schema and types match what the pinned codegen toolchain produces).
- If there's an internal generator for this file already, the script here can be dropped in favour of it; the regenerated schema is the part that matters.
- `Okta1` in the generated TypeScript is a `json2ts` artifact: two servers now have titles that both normalize to `Okta`, so the second is suffixed, and it shows up in compiler diagnostics. Interface names come from `title`, the one field copied verbatim from the catalog, so de-duplicating would mean editing catalog data — left alone deliberately.
- Two commits are naming-only: CodeQL read the property names the generator derives from the catalog's `secrets` field as credential values being written to disk, so the locals now say what they hold (`_env_vars`, `option_name`, `prop`). Output is byte-identical across both.

<div><a href="https://cursor.com/agents/bc-147135c0-6ac7-47bd-9557-eb48accf0f04?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-147135c0-6ac7-47bd-9557-eb48accf0f04&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changeset/mcp-catalog-refresh.md
+++ b/.changeset/mcp-catalog-refresh.md
@@ -3,28 +3,32 @@
 '@e2b/python-sdk': minor
 ---
 
-Refresh the MCP server schema from Docker's MCP catalog, which the `mcp` sandbox option is typed from. 52 servers are new — among them `curl`, `docker`, `ffmpeg`, `n8n`, `neo4j`, `temporal`, `playwright`, and 26 AWS servers — and every server's title and description now matches what the catalog publishes today.
+Refresh the MCP server schema from Docker's MCP catalog, which the `mcp` sandbox option is typed from. 52 servers are new — among them `curl`, `docker`, `ffmpeg`, `n8n`, `neo4j`, `temporal`, `proxmox`, `testkube`, `zen`, and 26 AWS servers — and every server's title and description now matches what the catalog publishes today.
 
 ```ts
 import { Sandbox } from 'e2b'
 
-const sandbox = await Sandbox.betaCreate({
+const sandbox = await Sandbox.create({
   mcp: {
     docker: {},
+    ffmpeg: {},
     n8n: { apiUrl: 'https://n8n.example.com/api/v1', apiKey: process.env.N8N_API_KEY! },
   },
 })
 ```
 
 ```python
+import os
+
 from e2b import Sandbox
 
-sandbox = Sandbox.beta_create(
+sandbox = Sandbox.create(
     mcp={
         "docker": {},
+        "ffmpeg": {},
         "n8n": {"apiUrl": "https://n8n.example.com/api/v1", "apiKey": os.environ["N8N_API_KEY"]},
     },
 )
 ```
 
-Six servers the catalog no longer publishes are gone from the type: `cdataConnectcloud`, `flexprice`, `postgres`, `root`, `tembo`, and `triplewhale`. Four more changed what they accept: `awsDiagram` takes `outputDir`, `context7` takes `apiKey`, `neo4jCypher` takes `schemaSampleSize`, and `onlyofficeDocspace` is down to `docspaceApiKey`. The configuration is still handed to the gateway as you write it, so a server the catalog dropped can be kept by casting past the type — whether it starts is up to the gateway.
+Some servers changed in ways that stop existing configuration from type-checking. Six the catalog no longer publishes are gone: `cdataConnectcloud`, `flexprice`, `postgres`, `root`, `tembo`, and `triplewhale`. `awsDiagram` and `context7` took no options before and now require one (`outputDir` and `apiKey`), so `awsDiagram: {}` and `context7: {}` need a value. `onlyofficeDocspace` is down to `baseUrl` and `docspaceApiKey`, having lost `docspaceAuthToken`, `docspacePassword`, `docspaceUsername`, `dynamic`, `origin`, `toolsets`, and `userAgent`. `neo4jCypher` gained `schemaSampleSize`. Configuration is still handed to the gateway as you write it, so a server the catalog dropped can be kept by casting past the type — whether it starts is up to the gateway.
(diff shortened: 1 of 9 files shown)


mishushakov:
@cursor check

mishushakov:
@cursor check

PR #219, opened 2023-11-10
Update Custom Sandboxes Section

tizkovatereza:

PR #1701, opened 2026-08-19
fix(python-sdk): stop buffering streamed request bodies on the retrying transports

mishushakov:
pyqwest's retry middleware keeps a request replayable, and for a body that isn't already `bytes` it does that by growing a copy of the body as it is sent. A streamed upload therefore went to the wire incrementally *and* was mirrored whole in RAM, so peak memory scaled with file size for `files.write` of a file-like object and for `volume.write_file`, even though nothing in the SDK buffers ([SDK-332](https://linear.app/e2b/issue/SDK-332/python-sdk-streamed-uploads-are-mirrored-in-ram-by-the-pyqwest-retry)). Template build contexts were never affected — they build their own transport with no retry middleware — and an earlier revision of this description, the changeset and the module comment all wrongly claimed otherwise; that is fixed.

The copy is what makes retries work, so [pyqwest#219](https://github.com/curioswitch/pyqwest/pull/219) added a way to decline it: `RetryMode.UNBUFFERED` replays a streamed body only while nothing has been read from it. That is exactly what the SDK's connect-only policy needs — `ConnectionRetryTransport` retries the builtin `ConnectionError`, which pyqwest raises only before the request was written — so the shared retrying transports (sync and async) now declare it as their per-request policy. `bytes` bodies, which is every unary RPC payload and every in-memory write, are replayable as they are and keep their retries either way.

**Read this before reviewing the diff: the fix is dormant on the pinned pyqwest.** `RetryMode` landed upstream after 0.9.0 and PyPI's latest pyqwest *is* 0.9.0, so there is no release to pin to yet and neither `pyproject.toml` nor `uv.lock` can move in this PR. The policy is therefore resolved from `pyqwest.middleware.retry` at import: with `RetryMode` the transports go unbuffered, without it they keep the buffered behavior that release ships. Merging this alone does not change what a `pip install e2b` user gets; bumping the pin to the release carrying [#218](https://github.com/curioswitch/pyqwest/pull/218) and [#219](https://github.com/curioswitch/pyqwest/pull/219) is what activates it, and needs no further SDK change. The alternative — widening the pin to `<0.11` now — would adopt an unreleased, untested pyqwest minor sight unseen, which is not how the SDK has taken pyqwest 0.8 and 0.9. [SDK-332](https://linear.app/e2b/issue/SDK-332/python-sdk-streamed-uploads-are-mirrored-in-ram-by-the-pyqwest-retry) stays open until the pin moves, and the changeset says the fix is inert until then.

### Why the unbuffered policy is safe now and was not before

The earlier reading of this ticket was that a connect-only policy provably never needs the copy, and [SDK-332 corrected it](https://linear.app/e2b/issue/SDK-332/python-sdk-streamed-uploads-are-mirrored-in-ram-by-the-pyqwest-retry): on pyqwest 0.9.0 reqwest read ahead into its body channel while connecting, so a connect error surfaced with the body already started (1–2 chunks in sync, up to the whole body in async) and a replay would have sent a truncated request. That is why the fallback keeps buffering rather than going unbuffered everywhere: on 0.9.0 the choice is the copy *or* the connect retries, and connectrpc hands pyqwest a generator even for unary RPCs, so dropping the retries would hit every envd RPC.

[pyqwest#218](https://github.com/curioswitch/pyqwest/pull/218) closed that gap — the body stream is started when hyper first polls it, not while connecting — so with it the SDK gets both. Measured against a refused port, chunks pulled from the body before `ConnectionError`, 3 runs each:

| body | pyqwest 0.9.0 sync | 0.9.0 async | with #218, sync | with #218, async |
| -- | -- | -- | -- | -- |
| 8 × 1 KiB | 2, 2, 2 | 8, 8, 8 (all of it) | 0, 0, 0 | 0, 0, 0 |
| 8 × 1 MiB | 0, 0, 1 | 1, 2, 1 | 0, 0, 0 | 0, 0, 0 |

**No user-facing API change**, so there are no usage examples to add: same public surface, same options, same timeouts, and connect retries behave as before. JS has no counterpart — undici streams request bodies without buffering and the retry middleware is pyqwest-only.

### Verification

Against pyqwest built from its `main` (i.e. with #218 + #219), traced peak allocations for a genuinely streamed upload of a 64 MiB file object, with the full body confirmed at the far end each time:

| path | buffered (today) | unbuffered |
| -- | -- | -- |
| `files.write`, sync | 65.3 MiB | 0.5 MiB |
| `files.write`, async | 65.9 MiB | 1.0 MiB |
| `files.write` multipart (`use_octet_stream=False`) | 65.5 MiB | 0.8 MiB |
| `volume.write_file` (32 MiB) | 36.2 MiB | 0.6 MiB |

`tests/test_retry_streamed_bodies.py` (new, 21 tests) is split so that CI exercises whichever policy is installed rather than skipping the file wholesale — 13 pass / 8 skip on the pinned pyqwest, 19 pass / 2 skip on `main`:

- always: the policy resolves to pyqwest's mode when the release has it and to buffered retries when it doesn't; both transports declare the override rather than inheriting a default that happens to match; one test names which regime is live so a pin bump flips it visibly; a failed connect replays an untouched streamed body from the start, reading the source exactly once; and, through the real `PyqwestTransport`/`AsyncPyqwestTransport` httpx adapter, a `content=` generator and a multipart file object both reach the middleware as a streamed (non-`bytes`) body — the join the transport-level tests would miss if the adapter ever flattened a body.
- unbuffered only: a 16 MiB streamed body reaches the transport whole with traced peak under 4 MiB, both at the transport and through the httpx adapter; a failure after the first chunk went out surfaces instead of replaying a truncated request; and a refused connect through a real pyqwest transport leaves the body unread.
- buffered only: the same mid-body failure *is* replayed from the copy (the first chunk goes out twice), which is why the fallback keeps it.

Mutation-checked on the pinned pyqwest, where the review found two mutants surviving: deleting both `should_retry_request` overrides now fails 2 tests, and deleting the `RetryMode` lookup fails 1 (both previously passed, since 0.9.0's inherited hook also returns `True`); a fallback of `False` fails 11.

- python-sdk unit suite: 432 passed / 9 skipped on the pinned pyqwest 0.9.0, 438 passed / 3 skipped on pyqwest `main` — so adopting that release does not disturb the rest of the SDK either.
- Against prod on pyqwest `main`: the full `files/` suites sync and async (123 tests, in-memory, streamed octet-stream and multipart writes), and `tests/sync` + `tests/async` minus the template build suites — 378 passed, with the same two `test_firewall_transform_injects_headers` failures the pinned pyqwest produces on that account (its `httpbin` test template is missing, unrelated).
- `ruff` lint/format and `ty` typecheck are green. No JS or TS files touched.
- CI: the required checks pass, including both production Python SDK jobs. The two failing *staging* Python jobs are `template_{sync,async}/test_build.py::test_build_template` timing out at 180s, which reproduces on unrelated PRs against staging right now (e.g. #1700) and touches no code in this diff.

### Notes for review

- The policy lives in `e2b.api._retry_request_policy`, resolved by `_resolve_retry_request_policy` so both of its arms are testable on any pyqwest. It is `_`-prefixed like the other internals in this family (the unprefixed neighbours each mirror a documented `E2B_*` knob; this one doesn't) and named for the policy it holds rather than as a flag, since its buffered value is `True`.
- `Any` on the constant and the overrides is forced, and the comment now says so: the pinned release declares `should_retry_request` as `-> bool`, so a union that admits `RetryMode` fails `ty`'s Liskov check until the pin moves.
- Streamed *downloads* were never affected: a response body is not buffered by the middleware, and a download request carries an empty `bytes` body, so it stays on the replayable path.
- Not in scope, now unblocked: template context uploads still build their own inline transport (a reqwest pool per build) specifically to stay off the retrying stack. With unbuffered retries they could join the shared pool without mirroring a build context in RAM.

<div><a href="https://cursor.com/agents/bc-f7148dd6-ba7f-4746-8ab1-79f8eb97f12f?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f7148dd6-ba7f-4746-8ab1-79f8eb97f12f&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changeset/python-unbuffered-retry-streamed-bodies.md
+++ b/.changeset/python-unbuffered-retry-streamed-bodies.md
@@ -1,0 +1,14 @@
+---
+"@e2b/python-sdk": patch
+---
+
+Stop mirroring streamed request bodies in memory while they are being sent. The
+shared retrying transports now declare pyqwest's unbuffered retry policy, so a
+streamed upload — `files.write` of a file-like object, `volume.write_file`,
+template build contexts — is no longer copied into a replay buffer as it goes to
+the wire, and peak memory stays flat instead of scaling with file size. Connect
+retries are unchanged: a connect error means the body was never read, so the
+same body is handed to the next attempt, and `bytes` bodies (unary RPC payloads,
+in-memory writes) were already replayable for free. Requires a pyqwest whose
+retry middleware exposes `RetryMode`; on older releases the SDK keeps that
+release's buffered behavior.
--- a/.changeset/python-unbuffered-retry-streamed-bodies.md
+++ b/.changeset/python-unbuffered-retry-streamed-bodies.md
@@ -2,13 +2,13 @@
 "@e2b/python-sdk": patch
 ---
 
-Stop mirroring streamed request bodies in memory while they are being sent. The
-shared retrying transports now declare pyqwest's unbuffered retry policy, so a
-streamed upload — `files.write` of a file-like object, `volume.write_file`,
-template build contexts — is no longer copied into a replay buffer as it goes to
-the wire, and peak memory stays flat instead of scaling with file size. Connect
-retries are unchanged: a connect error means the body was never read, so the
-same body is handed to the next attempt, and `bytes` bodies (unary RPC payloads,
-in-memory writes) were already replayable for free. Requires a pyqwest whose
-retry middleware exposes `RetryMode`; on older releases the SDK keeps that
-release's buffered behavior.
+Stop mirroring streamed request bodies in memory while they are being sent, on a
+pyqwest whose retry middleware exposes `RetryMode` (older releases keep the
+buffered replay they ship, so this is inert until the `pyqwest` dependency
+allows one). The shared retrying transports declare pyqwest's unbuffered retry
+policy, so a streamed `files.write` of a file-like object or `volume.write_file`
+is no longer copied into a replay buffer as it goes to the wire, and peak memory
+stays flat instead of scaling with file size. Connect retries are unchanged: a
+connect error means the body was never read, so the same body is handed to the
+next attempt, and `bytes` bodies (unary RPC payloads, in-memory writes) were
+already replayable for free.
--- a/packages/python-sdk/e2b/api/__init__.py
+++ b/packages/python-sdk/e2b/api/__init__.py
@@ -86,15 +86,17 @@ connection_retries = int(os.getenv("E2B_CONNECTION_RETRIES") or "3")
 pool_idle_timeout = float(os.getenv("E2B_KEEPALIVE_EXPIRY") or "300")
 pool_max_idle_per_host = int(os.getenv("E2B_MAX_KEEPALIVE_CONNECTIONS") or "20")
 
+
 # The retry policy the shared transports declare per request (see
 # `ConnectionRetryTransport` in client_sync/client_async).
 #
 # pyqwest's retry middleware keeps a request replayable, and for a body that
 # isn't already `bytes` it does that by growing a copy of it in memory as it is
 # sent. A streamed upload therefore went to the wire incrementally *and* was
-# mirrored whole in RAM, so peak memory scaled with file size for
-# `files.write`, `volume.write_file` and template context uploads even though
-# nothing in the SDK buffers (SDK-332).
+# mirrored whole in RAM, so peak memory scaled with file size for `files.write`
+# of a file-like object and for `volume.write_file`, even though nothing in the
+# SDK buffers (SDK-332). Template context uploads were never affected: they
+# build their own transport with no retry middleware in it.
 #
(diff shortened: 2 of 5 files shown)


mishushakov:
@cursor check

PR #1740, opened 2026-08-21
docs(js-sdk): 3 docstrings name parameters the functions do not take

karpovantonme:
`Volume.makeDir` and `Volume.writeFile` each document `options` and `opts`, and the signature takes 1 merged parameter, `opts?: VolumeWriteOpts & VolumeApiOpts`. I merged the 2 doc lines instead of dropping one, since both halves are still true of the single argument.

`TemplateBase.build` documents `tags`, which is not in the signature, and says nothing about `config`, which is. I added a line for `config`; drop that hunk if you would rather leave the internal method as it was.

Comments only, no behaviour change. Found by a checker I wrote that compares every documented `@param` name against the signature under it. It read 183 of the 185 documented comments in this repository and these 3 were all it had to say.

### DIFF
--- a/packages/js-sdk/src/template/index.ts
+++ b/packages/js-sdk/src/template/index.ts
@@ -1043,8 +1043,8 @@ export class TemplateBase
    * Internal implementation of the template build process.
    *
    * @param client API client for communicating with E2B backend
+   * @param config connection configuration for the API client
    * @param name Template name in 'name' or 'name:tag' format
-   * @param tags Additional tags to assign to the build
    * @param options Build configuration options
    * @throws BuildError if the build fails
    */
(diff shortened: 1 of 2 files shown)


mishushakov:
Hey @karpovantonme would you be able to follow Devin's guidance? Otherwise we will have to close the PR as stale.

karpovantonme:
Done -- both docstrings now name the metadata fields, what `force` means in each (parents for `makeDir`, overwrite for `writeFile`), and the write timeout defaulting to 1 hour instead of the usual 60s. `@returns` lists what comes back in `VolumeEntryStat` rather than restating the method name

mishushakov:
add a changeset and we're good to go

PR #1767, opened 2026-08-25
fix(sdk): git status parsing false-positives on detached HEAD and mis-splits repeated ellipsis

### LINKED ISSUE #1373: BUG: parse_git_status returns detached=True when upstream branch name contains the word 'detached'
## What breaks

`parse_git_status` in `packages/python-sdk/e2b/sandbox/_git/parse.py` incorrectly sets `detached=True` and leaves `current_branch`/`upstream` unpopulated for a repository that is **not** in detached HEAD mode when the upstream tracking branch name contains the word `detached` (e.g., `origin/detached-work`).

When branch `main` tracks `origin/detached-work`, `git status --porcelain=1 -b` outputs `## main...origin/detached-work`. The variable `raw_branch` is set to `"main...origin/detached-work"`, so the condition `"detached" in raw_branch` on line 129 evaluates to `True`, triggering the detached-HEAD code path and returning completely wrong state.

```
AssertionError: BUG: 'detached' in upstream name causes false-positive: is_detached=True
```

## Root cause

Line 128-130 in `parse.py`:
```python
is_detached = raw_branch.startswith("HEAD (detached at ") or (
    "detached" in raw_branch
)
```

The `"detached" in raw_branch` sub-check matches any upstream ref whose name contains the word `detached`, not just actual detached-HEAD porcelain output.

## Fix

Remove the `"detached" in raw_branch` sub-check. The `raw_branch.startswith("HEAD (detached at ")` check is sufficient for real detached HEAD, and the subsequent `normalized_branch.startswith("HEAD")` check on line 130 catches `HEAD (no branch)` and other HEAD variants.

oleg-ai-dev:
## Summary

`parse_git_status` (Python) / `parseGitStatus` (JS) share two bugs in the branch-header parsing at the top of `## <branch>...<upstream> [ahead N, behind N]`:

- **#1373** — `detached` was being detected with a bare substring check (`"detached" in raw_branch`), so any upstream branch name that merely *contains* the word "detached" (e.g. `origin/detached-work`) got misreported as a detached-HEAD state, with `current_branch`/`upstream` left empty. Fixed by relying only on the actual porcelain markers (`HEAD (detached at `, `HEAD (no branch)`) already checked a few lines down — same fix proposed in the issue.
- **#1371** — splitting on `"..."` with no limit breaks when the branch header itself contains more than one `...` separator: Python raises `ValueError: too many values to unpack`, and the JS SDK (same bug, not separately filed) silently truncates the upstream name via array destructuring instead of crashing. Fixed by splitting on only the first occurrence in both.

I independently verified the maintainer's comment on #1371 — `git check-ref-format` does reject branch names containing `..`/`...`, so the exact reproduction in the issue may not be reachable through normal git usage. I'm treating this as a defensive fix (it's a one-line change, and it converts undefined/incorrect behavior into correct behavior with no downside) rather than claiming it's a live crash for typical users.

Since the Python and JS SDKs implement this parser twice (checked for consistency by `test_parity.py`), I checked whether the same substring-based detection bug existed on the other side of the fence — it did, in `packages/js-sdk/src/sandbox/git/utils.ts`, so both are fixed together in one PR for parity rather than shipping two divergent behaviors.

## Test plan

- [x] Added `packages/python-sdk/tests/shared/git/test_parse.py` (no unit tests existed for this pure parser before) — 5 tests covering both fixes plus regression coverage for real detached HEAD, `HEAD (no branch)`, and ahead/behind parsing.
- [x] Added `packages/js-sdk/tests/sandbox/git/parse.test.ts` — same 5 cases, JS side.
- [x] `uv run pytest tests/shared/git/test_parse.py -v` — 5 passed.
- [x] `pnpm vitest run tests/sandbox/git/parse.test.ts` — 5 passed.
- [x] `uv run ruff check` / `uv run ruff format --check` on the changed Python files — clean.
- [x] `pnpm run typecheck` / `pnpm run lint` (oxlint) / `prettier --check` on the changed JS files — clean.
- [x] Added a changeset (`patch` bump on `e2b` and `@e2b/python-sdk`).

Fixes #1373, #1371

### DIFF
--- a/.changeset/fix-git-status-detached-ellipsis-parsing.md
+++ b/.changeset/fix-git-status-detached-ellipsis-parsing.md
@@ -1,0 +1,6 @@
+---
+'e2b': patch
+'@e2b/python-sdk': patch
+---
+
+Fix `git.status()` reporting `detached: true` when an upstream branch name merely contains the substring "detached" (e.g. `origin/detached-work`), and fix incorrect branch/upstream parsing when the branch header contains more than one `...` separator
--- a/packages/js-sdk/src/sandbox/git/utils.ts
+++ b/packages/js-sdk/src/sandbox/git/utils.ts
@@ -381,14 +381,14 @@ export function parseGitStatus(output: string): GitStatus {
       aheadStart === -1 ? undefined : branchInfo.slice(aheadStart + 2, -1)
     const normalizedBranch = normalizeBranchName(branchPart)
     const rawBranch = branchPart
-    const isDetached =
-      rawBranch.startsWith('HEAD (detached at ') ||
-      rawBranch.includes('detached')
+    const isDetached = rawBranch.startsWith('HEAD (detached at ')
 
     if (isDetached || normalizedBranch.startsWith('HEAD')) {
       detached = true
     } else if (normalizedBranch.includes('...')) {
-      const [branch, upstreamBranch] = normalizedBranch.split('...')
+      const separatorIndex = normalizedBranch.indexOf('...')
+      const branch = normalizedBranch.slice(0, separatorIndex)
+      const upstreamBranch = normalizedBranch.slice(separatorIndex + 3)
       currentBranch = branch || undefined
       upstream = upstreamBranch || undefined
     } else {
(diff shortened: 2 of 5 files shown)


mishushakov:
Hey there and thanks for starting a PR. We're not accepting any contributions to the Git integration at the moment.

oleg-ai-dev:
Understood — closing this since the Git integration isn't currently open to contributions. Thanks for the review pass; the parity findings (Python + JS both had the same detached-HEAD/ellipsis-split bugs) are documented in the PR description if useful for internal tracking.

PR #1788, opened 2026-08-28
fix(python-sdk): strip uppercase/mixed-case AS stage alias in from_dockerfile

eeshsaxena:
## What

`parse_dockerfile` (used by `Template.from_dockerfile`) strips a stage alias from the base image, but only when it is written as lowercase ` as `:

```python
if " as " in base_image.lower():
    base_image = base_image.split(" as ")[0].strip()
```

The guard lowercases the string, but the `split` runs on the original-case value with a **lowercase** delimiter. So the canonical, uppercase Docker form is not stripped:

```
FROM node:24 AS builder   -> base_image = "node:24 AS builder"   (broken image ref)
FROM node:24 As Build     -> base_image = "node:24 As Build"
FROM node:24 as builder   -> base_image = "node:24"              (only this worked)
```

`AS` is case-insensitive in Dockerfiles (and the docs use uppercase), so this is the common case.

## Fix

Match `AS` case-insensitively:

```python
base_image = re.split(r"\s+as\s+", base_image, maxsplit=1, flags=re.IGNORECASE)[0].strip()
```

Verified against the real `parse_dockerfile`: `FROM node:24 AS builder`, `as builder`, `As Build`, and `FROM node:24` all resolve to `node:24`. Added sync + async regression tests.

## How the code was written

Authored with the help of a coding agent; I verified the fix by running the actual `parse_dockerfile` and can explain the change.

### DIFF
--- a/packages/python-sdk/e2b/template/dockerfile_parser.py
+++ b/packages/python-sdk/e2b/template/dockerfile_parser.py
@@ -98,9 +98,14 @@ def parse_dockerfile(
 
         # Set the base image from the first FROM instruction
         base_image = from_instructions[0]["value"]
-        # Remove AS alias if present (e.g., "node:18 AS builder" -> "node:18")
-        if " as " in base_image.lower():
-            base_image = base_image.split(" as ")[0].strip()
+        # Remove a stage alias if present (e.g., "node:18 AS builder" -> "node:18").
+        # AS is case-insensitive in Dockerfiles, so it must be matched that way:
+        # splitting on a lowercase " as " left the canonical uppercase form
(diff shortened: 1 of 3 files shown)


eeshsaxena:
Concrete impact: a single-stage Dockerfile like FROM node:24 AS builder (the canonical uppercase form the Docker docs use) currently yields the base image "node:24 AS builder", which is not a valid image reference, so from_dockerfile builds a broken template. Lowercase happened to work, which is why it slipped through.

eeshsaxena:
@cla-bot check

PR #1789, opened 2026-08-29
fix(js-sdk): parse Dockerfile ENV/ARG values with spaces

eeshsaxena:
## Problem

The JS SDK's Dockerfile parser (`Template.fromDockerfile`) parsed each `ENV`/`ARG` argument token in isolation. `dockerfile-ast` splits arguments on whitespace — **including inside quotes** — so a value with a space landed in the "traditional `key value`" branch and produced a malformed key.

```ts
// Dockerfile:  ENV NAME="John Doe"
// dockerfile-ast tokens: ["NAME=\"John", "Doe\""]
//
// before: env var  {'NAME="John': 'Doe"'}   ❌
// after:  env var  { NAME: 'John Doe' }      ✔
```

This breaks the standard, documented way to put a space in an env value. `ENV KEY=hello world` was affected too. The **Python** SDK already parses these correctly (regex on the raw string + quote strip), so the same Dockerfile produced different env vars depending on which SDK you used.

## Fix

Rejoin the tokens and parse the raw value the same way as the Python SDK:
- `key=value` form: a value runs across whitespace until the next `key=` token; surrounding quotes are stripped.
- `key value` form unchanged; multiple `key=value` pairs on one line stay separated.

```ts
const template = Template().fromDockerfile(`FROM node:24
ENV NAME="John Doe"
ENV GREETING=hello world
ENV A=1 B=2`)
// NAME     -> "John Doe"
// GREETING -> "hello world"
// A -> "1", B -> "2"
```

## Notes

- Only the JS SDK is changed; this brings it in line with the already-correct Python implementation, so no Python change is needed.
- Added a `fromDockerfile` test asserting the quoted, spaced, and multi-pair cases, plus a changeset. I verified the new parser against `dockerfile-ast`'s real tokenization for all the cases above; the monorepo install was too slow to run the full `pnpm test` locally, so CI runs the suite.

### DIFF
--- a/.changeset/fix-dockerfile-env-quoted-values.md
+++ b/.changeset/fix-dockerfile-env-quoted-values.md
@@ -1,0 +1,5 @@
+---
+'e2b': patch
+---
+
+Fix `Template.fromDockerfile` parsing of `ENV`/`ARG` values that contain whitespace. A quoted value like `ENV NAME="John Doe"` (and an unquoted `ENV KEY=hello world`) was split into a malformed key; it is now parsed as a single value with surrounding quotes stripped, matching the Python SDK. Multiple `key=value` pairs on one line stay separated.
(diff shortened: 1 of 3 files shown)


eeshsaxena:
@cla-bot check

PR #1797, opened 2026-09-02
Preserve Retry-After on 429 RateLimit errors in Python and JS SDKs

### LINKED ISSUE #1325: Python and JS SDKs — 429 WAIT signal lost at exception boundary, no Retry-After

BetterAndBetterII:
RateLimit errors were dropping the `Retry-After` header, so clients could not back off. Preserve it on 429 in the Python and JS SDKs.

Fixes #1325

### DIFF
--- a/.changeset/retry-after-preserved.md
+++ b/.changeset/retry-after-preserved.md
@@ -1,0 +1,6 @@
+---
+'e2b': patch
+'@e2b/python-sdk': patch
+---
+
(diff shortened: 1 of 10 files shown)


BetterAndBetterII:
@cla-bot check

mishushakov:
Hey, thanks for the contribution! We will be dealing with this issue internally.

PR #1802, opened 2026-09-04
fix(python-sdk): don't crash the build error path on a non-numeric reason.step

nblintao:
## Problem

On a failed template build, `get_build_step_index` (`packages/python-sdk/e2b/template/utils.py`) assumes `reason.step` is `"base"`, `"finalize"`, or a numeric string — `int(step)` has no fallback, and the call sites in `template_sync/build_api.py` / `template_async/build_api.py` don't catch. Any other value raises `ValueError` from inside the error path, masking the real `BuildException` the user should have seen.

Non-numeric values occur in practice:

- e2b Cloud itself fills `step` with the literal strings `"optimize"` and `"resize-disk"` when those phases fail (`e2b-dev/infra`, `packages/orchestrator/pkg/template/build/phases/{optimize,ensurefreedisk}/builder.go` → `UnwrapUserError` passes `StepType` through verbatim) — so the crash is reachable against the official backend, not only third-party ones.
- The OpenAPI spec leaves the field an unconstrained string (`BuildStatusReason.step`: "Step that failed"), so nothing constrains it to be numeric.

The JS SDK already degrades gracefully here (`Number(step)` → `NaN` → no traceback, error still thrown); this brings the Python SDK to parity.

## Change

- `get_build_step_index` returns `Optional[int]`: non-numeric, non-sentinel steps map to `None` ("no matching local stack trace") instead of raising.
- Both call sites guard `None` and negative indices, so the `BuildException` is raised without a local traceback rather than being masked.
- Unit tests for all four cases (base / finalize / numeric / non-numeric).

### DIFF
--- a/.changeset/tolerate-non-numeric-step.md
+++ b/.changeset/tolerate-non-numeric-step.md
@@ -1,0 +1,5 @@
+---
+"@e2b/python-sdk": patch
+---
+
+fix(python-sdk): don't crash the build error path on a non-numeric reason.step
--- a/packages/python-sdk/e2b/template/utils.py
+++ b/packages/python-sdk/e2b/template/utils.py
@@ -406,7 +406,7 @@ def pad_octal(mode: int) -> str:
     return f"{mode:04o}"
 
 
-def get_build_step_index(step: str, stack_traces_length: int) -> int:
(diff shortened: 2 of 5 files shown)


nblintao:
@cla-bot check

PR #1806, opened 2026-09-05
docs(desktop-python): fix docstring placement and stale parameter docs

simpleqt:
Three docstring fixes in `packages/desktop-python/e2b_desktop/main.py`, each verified against the signature:

1. `right_click`: the docstring sat **after** the validation `raise`, making it a dead no-op string (`right_click.__doc__ is None`) — moved to the top of the function, matching the sibling click methods.
2. `screenshot`: documented a `'blob'` format that neither it nor the underlying `files.read` accepts (copy-paste from the JS SDK, whose `screenshot` does take `'blob'`). Now documents `'bytes'` or `'stream'`.
3. `drag(fr, to)`: documented `:param from:` — the parameter is named `fr` (`from` is a Python keyword).

### DIFF
--- a/packages/desktop-python/e2b_desktop/main.py
+++ b/packages/desktop-python/e2b_desktop/main.py
@@ -366,7 +366,7 @@ class Sandbox(SandboxBase):
         """
         Take a screenshot and return it in the specified format.
 
-        :param format: The format of the screenshot. Can be 'bytes', 'blob', or 'stream'.
+        :param format: The format of the screenshot. Can be 'bytes' or 'stream'.
         :returns: The screenshot in the specified format.
         """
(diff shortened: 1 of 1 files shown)


simpleqt:
CLA signed via cla.e2b.dev (simpleqt). Triggering recheck.

mishushakov (approved):

mishushakov:
@simpleqt can you please check Devin's comment?

simpleqt:
Checked Devin's comment and applied the suggestion: the right_click docstring now documents both params (with the together-constraint) and the ValueError raise, in the same reST form the rest of the file uses. Good catch — it was indeed the one docstring silent on the only raise in the file.

PR #1814, opened 2026-09-07
feat(desktop): stream option to composite the remote cursor into the framebuffer

Deep-Axe:
## Summary

Adds an opt-in composite cursor mode for desktop streams so view-only spectators can see the sandbox's remote pointer.

By default, x11vnc sends the cursor through the cursor-shape pseudo-encoding, which clients such as noVNC render at the viewer's pointer position rather than at positions set by `sandbox.moveMouse()` or clicks. With `cursor: 'composite'`, x11vnc receives `-nocursorshape` and paints the remote cursor into framebuffer updates instead. The default remains `'shape'` for backward compatibility.

Solves: [e2b-dev/desktop#249](https://github.com/e2b-dev/desktop/issues/249)

## API

```js
await desktop.stream.start({ requireAuth: true, cursor: 'composite' })
```

```python
desktop.stream.start(require_auth=True, cursor="composite")
```

## Changes

- Adds `VNCServerOptions.cursor?: 'shape' | 'composite'` to the JS SDK.
- Adds `cursor: Literal["shape", "composite"] = "shape"` to the Python SDK.
- Adds unit coverage and README examples for both SDKs.
- Adds a patch changeset for `@e2b/desktop` and `@e2b/desktop-python`.

Composite mode remains explicit instead of being inferred from `getUrl({ viewOnly: true })`, since client URL is generated after x11vnc starts.

## Verification

- JS SDK and Desktop SDK builds, typecheck, and lint pass; all Desktop SDK tests pass.
- Python lint, typecheck, build, and all Desktop SDK tests pass.
- Live RFB verification against a desktop sandbox:
  - Default `'shape'`: cursor pseudo-encoding is used, and remote mouse movement changes no framebuffer pixels at the pointer.
  - `'composite'`: `-nocursorshape` disables the pseudo-encoding, and remote movement updates the framebuffer at both the old and new pointer positions without affecting the control region.


Deep-Axe:
@mishushakov

PR #1818, opened 2026-09-07
feat(cli): add `template rebuild --refresh-envd`

AdaAibaby:
## Summary

Old templates bake an old envd into their final snapshot, so a sandbox started from one resumes the old envd out of snapshot RAM. That blocks features gated on a newer envd — e.g. volume mounts need **envd ≥ 0.5.14**. The only way to lift such a template is a rebuild, and users have been scripting that by hand against the raw API, which has two footguns:

- omitted specs silently default to **2 vCPU / 1024 MiB**, shrinking the template;
- registering under a fresh template id **orphans the alias**.

## Change

Adds **`e2b template rebuild <template> --refresh-envd`** (alias `rb`). It calls the companion server endpoint `POST /v2/templates/{templateID}/refresh-envd`, which derives a new build **FROM the template's own latest ready build** (base layer cached, only the envd binary swapped) and **inherits the source build's cpu/ram and alias in place**. The command then streams build logs via the existing status endpoint until the new build is ready.

Scope is deliberately **CLI-only**: it calls the raw endpoint through `client.api` (the `publish` command's pattern) and polls with the already-public `Template.getBuildStatus`. No js-sdk / python-sdk public surface changes, so no JS/Python parity work.

## Usage

```console
$ e2b template rebuild my-old-template --refresh-envd

Refreshing envd for my-old-template (from v0.4.1); rebuilding...
  ... build logs ...

✅ my-old-template rebuilt with the current envd.

   Confirm the binary changed: start a sandbox from my-old-template,
   then run /usr/bin/envd -version.
```

Without the flag it refuses (the flag is the only rebuild mode today, and gates a destructive-ish rebuild behind an explicit opt-in):

```console
$ e2b template rebuild my-old-template
Nothing to rebuild. Pass --refresh-envd to rebuild the template with the current envd.
```

## Companion infra PR (server half)

This is the **client half**. The server endpoint it calls is added in **e2b-dev/infra#3624** — https://github.com/e2b-dev/infra/pull/3624. That PR traces the full root cause (memory-snapshot resume keeps the old envd; live-upgrade is gated at envd ≥ 0.6.12; offline-upgrade only runs on the cold-boot path) and implements the endpoint that derives the cached-source-layer build which swaps envd.

**Merge order:** infra#3624 first, then here.

> ⚠️ **Preview files, blocked on the sync.** `spec/openapi.yml` and `packages/js-sdk/src/api/schema.gen.ts` in this PR carry the new endpoint so the CLI typechecks and builds. But per `AGENTS.md`, `spec/` is synced from `e2b-dev/infra` via Copybara at the pin in `spec/infra-ref` — it is not hand-authored. Once infra#3624 merges, the correct final step is: bump `spec/infra-ref` to the merged commit and run `make codegen`, which re-fetches the spec and regenerates `schema.gen.ts` **identically**, dropping the manual preview. This PR should not merge before that sync is possible.

## Tests run

- `pnpm --filter @e2b/cli run typecheck` — OK (new typed path resolves).
- `pnpm --filter @e2b/cli run build` — OK.
- `pnpm --filter @e2b/cli run lint` — 0 warnings / 0 errors.
- `pnpm --filter @e2b/cli run format` — clean.
- `pnpm --filter e2b run typecheck` (js-sdk) — OK (schema change doesn't break the SDK).
- `vitest run tests/commands/template/rebuild.test.ts` — PASS (2 tests: refuses without `--refresh-envd`; command registered + documented). The rest of the CLI suite passes; the only failure in a full run is the pre-existing `create.test.ts` backend-integration test, which is gated on a live `E2B_API_KEY` and unrelated to this change.

A changeset (`@e2b/cli: minor`) is included per `AGENTS.md`.

## AI assistance

AI assistance was used to trace the resume/build call chains, design the command, and draft the code and tests. A human submitter has reviewed every changed line.

### DIFF
--- a/.changeset/template-rebuild-refresh-envd.md
+++ b/.changeset/template-rebuild-refresh-envd.md
@@ -1,0 +1,5 @@
+---
+'@e2b/cli': minor
+---
+
+Add `e2b template rebuild <template> --refresh-envd`: rebuild an existing template with the host's current envd while keeping its specs and alias. Old templates bake an old envd into their snapshot, which blocks features gated on a newer envd (e.g. volume mounts need envd >= 0.5.14); this rebuilds in place FROM the template's own latest ready build so only the envd binary is swapped. Streams build logs until the new build is ready. Requires the companion `POST /v2/templates/{templateID}/refresh-envd` endpoint (e2b-dev/infra#3624).
--- a/packages/cli/src/commands/template/index.ts
+++ b/packages/cli/src/commands/template/index.ts
@@ -7,6 +7,7 @@ import { initCommand } from './init'
 import { listCommand } from './list'
 import { migrateCommand } from './migrate'
 import { publishCommand, unPublishCommand } from './publish'
+import { rebuildCommand } from './rebuild'
 
 export const templateCommand = new commander.Command('template')
   .description('manage sandbox templates')
@@ -19,3 +20,4 @@ export const templateCommand = new commander.Command('template')
   .addCommand(publishCommand)
   .addCommand(unPublishCommand)
   .addCommand(migrateCommand)
+  .addCommand(rebuildCommand)
--- a/packages/cli/src/commands/template/rebuild.ts
+++ b/packages/cli/src/commands/template/rebuild.ts
@@ -1,0 +1,101 @@
+import * as commander from 'commander'
+import { defaultBuildLogger, Template } from 'e2b'
+
+import { client, ensureAPIKey } from 'src/api'
+import { handleE2BRequestError } from '../../utils/errors'
+import {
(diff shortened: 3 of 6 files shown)


mishushakov:
Hey, thanks for the PR - we'll handle this one internally.

AdaAibaby:
> Hey, thanks for the PR - we'll handle this one internally.

@mishushakov Thanks for taking this over internally! Just a gentle follow-up — we currently have a lot of older templates running on low envd versions that require migration for volume mount compatibility. Would you have a rough timeline for when this `--refresh-envd` capability is expected to land in the main branch?

PR #1824, opened 2026-09-07
fix(js-sdk): validate sandbox create options before requiring an API key

### LINKED ISSUE #1820: JS Sandbox.create throws AuthenticationError for invalid lifecycle when the API key is missing
With no `E2B_API_KEY`, `pnpm --dir packages/js-sdk exec vitest run tests/sandbox/lifecyclePayload.test.ts --project unit` fails the two client-side tests with `AuthenticationError` instead of `InvalidArgumentError`. They die at `new ApiClient` in `createSandbox` (`sandboxApi.ts`), which runs before the lifecycle options are validated. Python `_create_sandbox` calls `build_lifecycle_config` before `get_api_client`, so moving the JS validation ahead of the client construction would match it.

claxman:
Fixes https://github.com/e2b-dev/E2B/issues/1820

`createSandbox` built `ApiClient` before validating `lifecycle`, so with no API key the client-side guards (`keepMemory: false` with `autoResume: true`, and `keepMemory` on `kill`) threw `AuthenticationError` instead of `InvalidArgumentError`. Python already calls `build_lifecycle_config` before `get_api_client`. `forkSandbox` already validates `count` before constructing the client.

The lifecycle guards and create body run first. `resolveOpts`, `ConnectionConfig`, and `ApiClient` sit together immediately before POST.

Two new cases in `lifecycleRequest.test.ts` unset `E2B_API_KEY` and expect `InvalidArgumentError`.

```
pnpm --dir packages/js-sdk exec vitest run --project unit tests/sandbox/lifecycleRequest.test.ts
```

12 passed. Restoring `createSandbox` to construct `ApiClient` first makes the two new tests fail with `AuthenticationError`.

```ts
await Sandbox.create('base', {
  lifecycle: { onTimeout: { action: 'pause', keepMemory: false }, autoResume: true },
})
// throws InvalidArgumentError with no E2B_API_KEY set
```

### DIFF
--- a/.changeset/lifecycle-validate-before-auth.md
+++ b/.changeset/lifecycle-validate-before-auth.md
@@ -1,0 +1,5 @@
+---
+'e2b': patch
+---
+
+Reject invalid `Sandbox.create` lifecycle options before requiring an API key, matching the Python SDK.
--- a/packages/js-sdk/src/sandbox/sandboxApi.ts
+++ b/packages/js-sdk/src/sandbox/sandboxApi.ts
@@ -1634,7 +1634,6 @@ export class SandboxApi extends ClientFactory {
   ) {
     const apiOpts = this.resolveOpts(opts)
     const config = new ConnectionConfig(apiOpts)
-    const client = new ApiClient(config)
     // onTimeout accepts a bare action (`'pause'` / `'kill'`) or the object form
     // `{ action, keepMemory }`. The discriminated union type forbids `keepMemory`
     // on `action: 'kill'`; re-check at runtime for untyped callers.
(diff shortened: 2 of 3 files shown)


mishushakov:
good find, can you perhaps move entire block down?

```ts
    const apiOpts = this.resolveOpts(opts)
    const config = new ConnectionConfig(apiOpts)
    const client = new ApiClient(config)
```

claxman:
Moved. `resolveOpts`, `ConnectionConfig`, and `ApiClient` now sit together after the lifecycle guards and the create body, immediately before POST. `pnpm --dir packages/js-sdk exec vitest run --project unit tests/sandbox/lifecycleRequest.test.ts`: 12 passed, including the two cases with `E2B_API_KEY` unset.

PR #1826, opened 2026-09-07
fix(code-interpreter): deserialize SuperChart from elements

### LINKED ISSUE #1821: JS SuperChart deserializer reads data instead of elements, and Result never deserializes charts
`deserializeChart` in `packages/code-interpreter-js/src/charts.ts` maps SuperChart children from `data.data`. The `SuperChart` type and Python `_deserialize_chart` use `elements`. `Result` in `messaging.ts` assigns `this.chart = data['chart']` and never calls `deserializeChart`. Python `Result` does call `_deserialize_chart`.

claxman:
Fixes https://github.com/e2b-dev/E2B/issues/1821

`deserializeChart` mapped SuperChart children from `data.data`. The kernel, the `SuperChart` type, and Python `_deserialize_chart` use `elements`. Calling the helper on a real SuperChart threw `Cannot read properties of undefined (reading 'map')`. `Result` assigned the raw `chart` payload and never called it, so nested unknown types stayed as the raw string.

Walk `elements`, keep `elements` on the result, and run `Result` through `deserializeChart` like Python. I did not export `deserializeChart`; it is still internal. There is no fallback to a `data` key. A null `chart` on the wire becomes `undefined`, not `null`.

Testing (no sandbox):

```
pnpm --dir packages/code-interpreter-js exec vitest run tests/charts/deserialize.test.ts tests/messaging.test.ts
```

7 passed. Restoring `charts.ts` and `messaging.ts` from main makes the SuperChart tests fail (`.map` TypeError, nested `type: 'not-a-chart'`). Restoring only the `chart: null` else-branch to `data['chart']` makes `Result` keep `null` instead of `undefined`.

```ts
const result = new Result(
  {
    chart: {
      type: 'superchart',
      title: 'Multiple Charts Example',
      elements: [
        { type: 'line', title: 'Sine Wave', elements: [] },
        { type: 'not-a-chart', title: 'nested', elements: [] },
      ],
    },
  },
  true
)
result.chart.elements[1].type // 'unknown'
```

### DIFF
--- a/.changeset/superchart-deserialize-elements.md
+++ b/.changeset/superchart-deserialize-elements.md
@@ -1,0 +1,5 @@
+---
+'@e2b/code-interpreter': patch
+---
+
+Deserialize SuperChart from `elements` and run `Result` through `deserializeChart`, matching the Python SDK.
--- a/packages/code-interpreter-js/src/charts.ts
+++ b/packages/code-interpreter-js/src/charts.ts
@@ -127,13 +127,9 @@ export function deserializeChart(data: any): Chart {
     case ChartType.BOX_AND_WHISKER:
       return { ...data } as BoxAndWhiskerChart
     case ChartType.SUPERCHART: {
-      const nested = data.elements ?? data.data ?? []
-      const charts: Chart[] = nested.map((g: any) => deserializeChart(g))
-      const rest = { ...data }
-      delete rest.data
       return {
-        ...rest,
-        elements: charts,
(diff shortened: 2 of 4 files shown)


claxman:
Done. Result now maps a falsy chart to undefined. With `{ chart: null }`, `result.chart` is undefined; restoring `: data['chart']` makes the new test fail with expected undefined, received null.

PR #1831, opened 2026-09-07
fix(sdk): reject unrecognized files.read format

### LINKED ISSUE #1823: files.read / volume.read return None for an unrecognized format instead of raising
`format` dispatch in `read()` has no fallback branch, so a value outside the union silently returns `None` (Python) rather than the file contents or an error.

Verified on staging with a real sandbox:

```
format='text'   -> str: 'hello'
format='bytes'  -> bytearray: bytearray(b'hello')
format='Text'   -> NoneType: None      # falls through every branch
format='txt'    -> NoneType: None
```

The dispatch is `if format == "text": ... elif format == "bytes": ...` with no `else`, so control falls off the end of the function after the request has already succeeded. The caller gets `None` where content should be, with no exception and nothing to indicate the argument was wrong.

Affected (both Python mirrors, and the JS equivalents):

- `packages/python-sdk/e2b/sandbox_sync/filesystem/filesystem.py` — `read()`
- `packages/python-sdk/e2b/sandbox_async/filesystem/filesystem.py` — `read()`
- `packages/python-sdk/e2b/volume/volume_sync.py` / `volume_async.py` — `read()`
- `packages/js-sdk/src/volume/index.ts` — the fall-through is marked by a `// format === 'blob'` comment, so an unrecognized value silently returns a Blob rather than the requested shape
- `packages/js-sdk/src/sandbox/filesystem/index.ts` — same shape

`Literal` / the TS union catch this for typed callers; it reaches anyone passing the value through from config, a CLI flag, or plain JS.

Suggested fix: an `else: raise InvalidArgumentException(...)` / `throw new InvalidArgumentError(...)` naming the valid formats, matching `GitResetMode` and `GitConfigScope`, which already validate their literal domains in both SDKs. Ideally before the request is issued rather than after, so a bad argument doesn't cost a round trip.

Related: the same class on `onTimeout` / `onResume`, where the value is coerced to a boolean before the request is built — being fixed in #1822. Those two are more severe (a mistyped `on_timeout` deletes the sandbox at timeout); this one is a wrong return value rather than data loss, which is why it's split out.

claxman:
Fixes https://github.com/e2b-dev/E2B/issues/1823

`files.read` / `Filesystem.read` and `Volume.readFile` / `read_file` dispatch on `format` with no fallback. A value outside the documented literals (`Text`, `txt`) returned `None` from the Python filesystem after a successful GET, and Python volume returned the file as text. JS passed the string through to `parseAs` and did not throw.

Validate immediately after resolving `format`, before the GET. `InvalidArgumentError` / `InvalidArgumentException`: `format must be one of text, bytes, stream.` (JS also allows `blob`). The accepted set is one `ReadFormat` alias per SDK (`get_args` / `as const`). JS defaults only when `format` is `undefined`, so `null` is rejected like Python `None`.

Python shares `validate_read_format` next to `validate_metadata`. JS shares `assertReadFormat` from `src/utils.ts` (named exports, not `export *`).

Testing (no sandbox):

```
cd packages/python-sdk && .venv/bin/python -m pytest tests/test_read_format.py -q
pnpm --dir packages/js-sdk exec vitest run --project unit tests/readFormat.test.ts
```

4 python + 4 JS passed. No-op `validate_read_format` / `assertReadFormat` and the unrecognized-format cases fail with `GET should not be called`.

```py
sandbox.files.read('/tmp/x', format='Text')
```

InvalidArgumentException: format must be one of text, bytes, stream.

### DIFF
--- a/.changeset/read-format-validate.md
+++ b/.changeset/read-format-validate.md
@@ -1,0 +1,6 @@
+---
+'e2b': patch
+'@e2b/python-sdk': patch
+---
+
+Reject an unrecognized `format` on `files.read` / `Volume.readFile` (`read_file`) with `InvalidArgumentError` / `InvalidArgumentException` instead of returning None or the file as text.
--- a/packages/js-sdk/src/sandbox/filesystem/index.ts
+++ b/packages/js-sdk/src/sandbox/filesystem/index.ts
@@ -53,7 +53,7 @@ import {
   TemplateError,
 } from '../../errors'
 import { isReadableStreamLike } from '../../is'
-import { runtime, toBlob, toUploadBody } from '../../utils'
+import { runtime, toBlob, toUploadBody, assertReadFormat } from '../../utils'
 
 const FILESYSTEM_HTTP_ERROR_MAP: Record<number, (message: string) => Error> = {
   404: (message: string) => new FileNotFoundError(message),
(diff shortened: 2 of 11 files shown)


claxman:
@cla-bot check

claxman:
`ReadFormat` is now the single source: Python `get_args(Literal[...])`, JS `as const`. Implementation signatures use the alias, overload literals stay so return types stay discriminated.

JS defaults only when `format` is `undefined`, so `{ format: null }` throws `InvalidArgumentError` before the GET, same as Python `format=None`. `tests/readFormat.test.ts` 4 passed including the null case, `tests/test_read_format.py` 4 passed.

Volume imports the helper from `sandbox.filesystem`, next to `validate_metadata`. No `@throws` / `:raises` added since `rg` finds none on the existing validators.

PR #1832, opened 2026-09-07
feat(sdk): add ServiceBusyError for 503 and expose the HTTP status on API errors

bchalios:
## Why

The API answers **503** when it refuses an operation because the service or the node running the sandbox is temporarily busy: today on create, connect, fork and resume when there is no capacity, and — with the pause-admission work now released (belt#2529, api v0.11.0, flag-gated) — on `pause` when the node refuses a snapshot while the previous one is still deduplicating. In that last case the sandbox is **unchanged and still running**, and the right reaction is to retry or simply keep using it.

Neither SDK let a caller tell that apart from a real failure: both mapped 503 to the generic `SandboxError` / `SandboxException` with the status only in the message text (`503: node is busy persisting sandbox, please retry`), so the only way to branch was a string prefix check.

## What changes

- **`ServiceBusyError`** (JavaScript) / **`ServiceBusyException`** (Python), raised for every 503 from the API, whatever operation raised it. A standalone error like `AuthenticationError` — it does **not** subclass `SandboxError` / `SandboxException`, so catch it explicitly.
- **`statusCode`** (JavaScript) / **`status_code`** (Python) on the base `SandboxError` / `SandboxException`: the HTTP status of the API response that produced the error, when there was one. `RateLimitError` / `RateLimitException` now carry 429, `ServiceBusyError` carries 503 on its own `statusCode`, and the generic mapping carries whatever status it saw. Callers can branch on the status without parsing the message.
- Mapping lives in the one place both SDKs already share for API errors (`apiErrorFromCode` / `api_exception_from_code`), so the sync and async Python clients and every JS call site get it at once. A 503 with an empty body is still a busy error.
- Exports added to `e2b` (JS) and `e2b` (Python) top level. Changeset: minor for both packages.

## Usage

```ts
import { Sandbox, ServiceBusyError } from 'e2b'

try {
  await sandbox.pause()
} catch (e) {
  if (e instanceof ServiceBusyError) {
    // the sandbox is still running; retry a little later, or keep using it
    await new Promise((r) => setTimeout(r, 5_000))
    await sandbox.pause()
  } else {
    throw e
  }
}
```

```python
from e2b import Sandbox, ServiceBusyException

try:
    sandbox.pause()
except ServiceBusyException:
    time.sleep(5)
    sandbox.pause()
```

```ts
} catch (e) {
  if (e instanceof SandboxError && e.statusCode === 503) { ... }  // equivalent, without the class
}
```

No automatic retry is added: that stays with the caller, per the design decision behind the pause change (revisit once refusal-rate data exists).

## Not in this PR

The `spec/openapi.yml` copy here predates belt#2529 and does not yet declare 503 on the pause path. The spec is Copybara-synced from infra (never hand-edited), so the generated types will pick it up on the next `spec/infra-ref` bump; behaviour does not depend on it, since both SDKs branch on the raw status.

## Tests

- JS: `tests/api/handleApiError.test.ts` gains cases for 503 with and without a body, the status on busy / rate-limit / generic errors. `pnpm run lint`, `pnpm run typecheck` clean; `vitest run tests/api/handleApiError.test.ts tests/envd/handleEnvdApiError.test.ts` 39 passed. The rest of the `unit` project needs an API key and was not run here.
- Python: new `tests/test_api_exception.py` (busy with and without body, generic keeps its class and gets its status, a 503 through another hierarchy keeps that hierarchy, a dataclass exception that bypasses the initializer still exposes `status_code`, rate limit carries 429); `ruff`, `ty` clean; `pytest tests/test_api_exception.py tests/test_logging_option.py tests/test_envd_api_exception.py` 38 passed.
- Review rounds folded: per the human review the 503 error is standalone (not a `SandboxError`) and raised for every 503 regardless of the caller's error class; Python's generic paths stamp `status_code` like JS; `status_code` has a class-level default for dataclass exceptions; the attribute is documented in the class docstring.
- Also in this PR: `packages/desktop-python/uv.lock` relocked for e2b 2.47.0 — the last release commit bumped python-sdk without relocking, so Lint's uncommitted-changes step and the desktop-python install were already failing on main.
- Staging integration jobs: on 717d82c the Windows JS job failed on staging template builds and the Ubuntu Python job on two network-restriction tests; both passed on the previous head with the same error-mapping paths exercised, and neither test touches this diff — staging environment flakes, not this change.

Related: Linear EN-2608 (the pause-error investigation this stems from).


packages/python-sdk/e2b/api/__init__.py
```diff
@@ -199,6 +200,12 @@ def api_exception_from_code(
...
 
+    if status_code == 503:
+        text = f"{status_code}: Service temporarily unavailable, please retry."
+        if message:
+            text += f" - {message}"
+        return SandboxBusyException(text)
+
     return default_exception_class(f"{status_code}: {message}").with_traceback(
         stack_trace
     )
```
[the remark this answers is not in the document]
bchalios: Fixed at 717d82c: both returns of the generic path now stamp `status_code` when the class is a `SandboxException`, mirroring the JS `instanceof` stamp, and `tests/test_api_exception.py` asserts `== 500` (plus a body-less 502 case).

packages/js-sdk/src/api/index.ts
```diff
@@ -33,7 +38,17 @@ export function apiErrorFromCode(
     return new RateLimitError(content ? `${message} - ${content}` : message)
   }
 
-  return new errorClass(`${code}: ${content}`, stackTrace)
+  if (code === 503) {
+    const message = 'Service temporarily unavailable, please retry'
+    return new SandboxBusyError(content ? `${message} - ${content}` : message)
+  }
```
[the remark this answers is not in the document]
bchalios: Fixed at 717d82c, both SDKs: the busy class is returned only when the caller is in the sandbox domain (`errorClass === SandboxError` / `default_exception_class is SandboxException`); any other hierarchy keeps its class and carries the 503 as the status. Tests cover a 503 through a `BuildLike` subclass in each SDK.

packages/python-sdk/e2b/exceptions.py
```diff
@@ -17,7 +20,10 @@ class SandboxException(Exception):
     Raised when a general sandbox exception occurs.
     """
 
-    pass
+    def __init__(self, *args, status_code: Optional[int] = None):
+        super().__init__(*args)
+        # HTTP status of the API response that produced this error, when there was one.
+        self.status_code = status_code
```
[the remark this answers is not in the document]
bchalios: Fixed at 717d82c: `status_code` is documented as a reST `:param:` in the `SandboxException` docstring and the inline comment is gone.

packages/python-sdk/tests/test_api_exception.py
```diff
@@ -0,0 +1,46 @@
...
+    assert isinstance(err, SandboxBusyException)
+    assert err.status_code == 503
+
+
+def test_generic_failure_stays_a_sandbox_exception_without_a_status():
+    err = handle_api_exception(response(500, b'{"message":"Internal error"}'))
+
+    assert isinstance(err, SandboxException)
+    assert not isinstance(err, SandboxBusyException)
+    assert err.status_code is None
```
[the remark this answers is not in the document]
bchalios: Fixed at 717d82c (same point as the TASTE review's T-1): both generic paths in `handle_api_exception` / `api_exception_from_code` now stamp `status_code` on `SandboxException` instances, and the test asserts `== 500`, plus a body-less 502 case.

packages/python-sdk/e2b/exceptions.py
```diff
@@ -15,9 +18,13 @@ class SandboxException(Exception):
...
 
     Raised when a general sandbox exception occurs.
+
+    :param status_code: HTTP status of the API response that produced this error, when there was one.
     """
 
-    pass
+    def __init__(self, *args, status_code: Optional[int] = None):
+        super().__init__(*args)
+        self.status_code = status_code
```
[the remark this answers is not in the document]
bchalios: Good catch — fixed at 8d46d60: `status_code` now has a class-level default of `None` on `SandboxException`, so dataclass subclasses that bypass the initializer (`CommandExitException`) expose it too; a test constructs one and reads the attribute.

packages/python-sdk/e2b/api/__init__.py
```diff
@@ -199,6 +200,12 @@ def api_exception_from_code(
             text += f" - {message}"
         return RateLimitException(text)
 
+    if status_code == 503:
+        text = f"{status_code}: Service temporarily unavailable, please retry."
+        if message:
+            text += f" - {message}"
+        return SandboxBusyException(text)
```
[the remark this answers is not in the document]
bchalios: Already addressed at 717d82c (the TASTE review raised the same point as T-57): the 503 branch now fires only when `default_exception_class is SandboxException`; any other hierarchy (`VolumeException`, `SecretException`, `BuildException`, …) keeps its class and carries the 503 as `status_code`. Test: `test_503_through_another_hierarchy_keeps_that_hierarchy`. Same guard on the JS side (`errorClass === SandboxError`).

packages/python-sdk/e2b/api/__init__.py
```diff
@@ -199,6 +200,12 @@ def api_exception_from_code(
...
+
     return default_exception_class(f"{status_code}: {message}").with_traceback(
         stack_trace
```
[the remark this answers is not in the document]
bchalios: Already addressed at 717d82c: both generic paths in Python now stamp `status_code` on `SandboxException` instances, matching JS; the test asserts `== 500` and a body-less 502 case.

packages/js-sdk/src/api/index.ts
```diff
@@ -33,7 +38,20 @@ export function apiErrorFromCode(
...
   }
 
-  return new errorClass(`${code}: ${content}`, stackTrace)
+  // Only the sandbox domain gets the dedicated class: a 503 raised through
+  // another hierarchy (BuildError, TemplateError, ...) keeps that hierarchy
+  // and carries the status instead.
+  if (code === 503 && errorClass === SandboxError) {
+    const message = 'Service temporarily unavailable, please retry'
+    return new SandboxBusyError(content ? `${message} - ${content}` : message)
+  }
```
mishushakov: I am thinking whether we should instead go for generic ServiceBusyError instead ?

bchalios: yup, that makes sense. I'll update.

bchalios: Renamed at 1a46e7b: `ServiceBusyError` (JS) / `ServiceBusyException` (Python), still subclassing `SandboxError` / `SandboxException` so existing handlers keep working; docstrings and changeset updated to describe it as the generic 503 (placement capacity, or a node declining a request it cannot serve yet).

bchalios: done

packages/js-sdk/src/errors.ts
```diff
@@ -144,6 +149,24 @@ export class RateLimitError extends SandboxError {
...
+/**
+ * Thrown when the API refused the operation because the service is
+ * temporarily busy (HTTP 503): no capacity to place a sandbox right now, or
+ * the node running the sandbox declined a request it cannot serve yet.
+ *
+ * Nothing was changed by the refused call: for example a refused pause
+ * leaves the sandbox running with its state intact, so the same call can be
+ * retried after a short wait.
+ */
+export class ServiceBusyError extends SandboxError {
```
mishushakov: if it's generic it should not extend SandboxError

packages/js-sdk/src/api/index.ts
```diff
@@ -33,7 +38,20 @@ export function apiErrorFromCode(
     return new RateLimitError(content ? `${message} - ${content}` : message)
   }
 
-  return new errorClass(`${code}: ${content}`, stackTrace)
+  // Only the sandbox domain gets the dedicated class: a 503 raised through
+  // another hierarchy (BuildError, TemplateError, ...) keeps that hierarchy
+  // and carries the status instead.
+  if (code === 503 && errorClass === SandboxError) {
```
mishushakov: if it's generic it also should not only be thrown when it's a SandboxError

mishushakov (changes requested):
if we are going for a generic version it should be decoupled from SandboxError altogether

packages/js-sdk/src/errors.ts
bchalios: Done at 1dc39f6: `ServiceBusyError` extends `Error` (and `ServiceBusyException` extends `Exception`), the same shape as `AuthenticationError`; it carries `statusCode = 503` itself. Docstrings and the changeset say to catch it explicitly.

packages/js-sdk/src/api/index.ts
bchalios: Done at 1dc39f6: the guard is gone — every 503 is a `ServiceBusyError` / `ServiceBusyException` whatever error class the caller asked for, in both SDKs. Tests flipped accordingly.

packages/js-sdk/tests/api/handleApiError.test.ts
```diff
@@ -73,7 +73,7 @@ describe('handleApiError', () => {
       const res = createMockResponse(503, undefined)
       const err = handleApiError(res as any)
       assert.instanceOf(err, ServiceBusyError)
-      assert.instanceOf(err, SandboxError)
+      assert.notInstanceOf(err, SandboxError)
```
mishushakov: remove notInstanceOf checks

packages/python-sdk/tests/test_api_exception.py
```diff
@@ -18,7 +18,7 @@ def test_refused_pause_is_a_busy_exception_with_the_status_and_message():
     )
 
     assert isinstance(err, ServiceBusyException)
-    assert isinstance(err, SandboxException)
+    assert not isinstance(err, SandboxException)
```
mishushakov: here too

mishushakov (approved):
approved but update the tests please

packages/js-sdk/tests/api/handleApiError.test.ts
bchalios: Removed at d95f104.

packages/python-sdk/tests/test_api_exception.py
bchalios: Removed at d95f104.

PR #1835, opened 2026-09-07
fix(sdk): validate secret names in Secret.create

### LINKED ISSUE #1833: Secret.create skips the name rules that Secret.fill already enforces
Secret.fill('') and Secret.fill('a}b') raise InvalidArgumentException, but Secret.create('', 'v') and Secret.create('a}b', 'v') skip that check: in the python-sdk venv with E2B_API_KEY unset they raise AuthenticationException from get_api_client instead. JS matches: in packages/js-sdk/src/secret.ts validateSecretName is only called from fill(), and create() constructs ApiClient first.

claxman:
Fixes https://github.com/e2b-dev/E2B/issues/1833

`Secret.fill` already rejects empty names and names that contain `{`, `}`, or control characters. `Secret.create` constructed `ApiClient` / `get_api_client` first, so the same names raised `AuthenticationError` / `AuthenticationException` when the API key was missing, and a mocked API would store the name. I did not apply the check to update/get/destroy; those selectors can be secret IDs.

Call `validateSecretName` / `_validate_secret_name` at the start of `create` (JS, Python sync, Python async). JSDoc `@throws` and Sphinx `:raises` document that failure mode on `create`.

Testing (no live API):

```
pnpm --dir packages/js-sdk exec vitest run --project unit tests/secret/createName.test.ts
cd packages/python-sdk && .venv/bin/python -m pytest tests/sync/secret_sync/test_secret.py tests/async/secret_async/test_secret.py -q
```

6 JS + 42 Python passed. Removing the three `create` checks makes the new cases fail: JS with `AuthenticationError` or `POST should not be called`, Python with `AuthenticationException` or a stored mock secret.

```ts
await Secret.create('', 'v')
// InvalidArgumentError, including when E2B_API_KEY is unset
```

### DIFF
--- a/.changeset/secret-create-validate-name.md
+++ b/.changeset/secret-create-validate-name.md
@@ -1,0 +1,6 @@
+---
+'e2b': patch
+'@e2b/python-sdk': patch
+---
+
+Reject unusable secret names in `Secret.create` with `InvalidArgumentError` / `InvalidArgumentException` before constructing the API client, matching `Secret.fill`.
--- a/packages/js-sdk/src/secret.ts
+++ b/packages/js-sdk/src/secret.ts
@@ -163,6 +163,8 @@ export class Secret extends ClientFactory {
    *
    * @returns the secret's ID, name, current version (`1` for a new secret),
    * metadata, and creation and update times.
+   *
+   * @throws {InvalidArgumentError} if `name` is empty or contains `{`, `}` or control characters.
    */
   static async create(
     name: string,
--- a/packages/js-sdk/src/secret.ts
+++ b/packages/js-sdk/src/secret.ts
@@ -169,6 +169,8 @@ export class Secret extends ClientFactory {
     value: string,
     opts?: SecretCreateOpts
   ): Promise<SecretInfo> {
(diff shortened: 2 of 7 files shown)


claxman:
@cla-bot check

claxman:
Documented. `Secret.create` now has JSDoc `@throws {InvalidArgumentError}` and Sphinx `:raises InvalidArgumentException` for an empty name or one that contains `{`, `}`, or control characters. I left `fill()` alone; that gap was already there.

mishushakov:
same question I have as your other PR, do we actually need this because the backend is validating the format already

claxman:
The create check matches fill(), not the OpenAPI pattern: empty, `{`, `}`, control chars, because fill() interpolates `${e2b.secrets.<name>}`. Happy to drop it if you want create to stay a pass-through.

mishushakov:
for fill it's critical because it doesn't run against any API so it has to be a proper secret name

claxman:
Agreed. fill() never hits the API so that check stays. create stays a pass-through; the backend 400 is the source of truth. Closing this.

PR #1842, opened 2026-09-08
fix(python): keep timeout=0 on sandbox create and connect

### LINKED ISSUE #1840: Python Sandbox.create(timeout=0) sends the 300s default instead of 0
Python `Sandbox.create(timeout=0)` with a mocked `post_sandboxes` posted `timeout: 300`. Same for `Sandbox.connect(..., timeout=0)` via mocked `post_sandboxes_sandbox_id_connect`. Sync `_create` uses `timeout or cls.default_sandbox_timeout` (300) and sync `_cls_connect` uses `timeout or SandboxBase.default_sandbox_timeout`, so 0 is treated as unset. JS create uses `timeoutMs ?? defaultSandboxTimeoutMs`, which leaves 0 alone.

claxman:
Fixes https://github.com/e2b-dev/E2B/issues/1840

`Sandbox.create(timeout=0)` and `Sandbox.connect(..., timeout=0)` used `timeout or default`, so 0 became 300 seconds. JS create uses `timeoutMs ?? default` and keeps 0. Fork already used `if timeout is not None`.

Same `if timeout is not None` default in sync and async `_create` and `_cls_connect`. I did not reject `timeout <= 0`. JS sends 0.

Testing (no live API):

```
cd packages/python-sdk && .venv/bin/python -m pytest tests/shared/sandbox/test_timeout.py -q --tb=short
```

6 passed. Restoring the four `or` defaults makes the four `timeout=0` cases fail with posted timeout 300 instead of 0.

```python
Sandbox.create(timeout=0, api_key=...)
# POST body timeout is 0, not 300
```

### DIFF
--- a/.changeset/python-timeout-zero.md
+++ b/.changeset/python-timeout-zero.md
@@ -1,0 +1,5 @@
+---
+'@e2b/python-sdk': patch
+---
+
+Keep an explicit `timeout=0` on sandbox create and connect instead of substituting the 300s default.
--- a/packages/python-sdk/e2b/sandbox_async/main.py
+++ b/packages/python-sdk/e2b/sandbox_async/main.py
@@ -1158,7 +1158,9 @@ class AsyncSandbox(SandboxApi):
         else:
             response = await SandboxApi._create_sandbox(
                 template=template or cls.default_template,
-                timeout=timeout or cls.default_sandbox_timeout,
+                timeout=(
+                    timeout if timeout is not None else cls.default_sandbox_timeout
+                ),
                 metadata=metadata,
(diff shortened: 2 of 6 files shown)


claxman:
@cla-bot check

mishushakov:
I believe this is on purpose, keep in mind also that we will be dropping all the SDK defaults from the SDK in favor of backend-defined values:
https://github.com/e2b-dev/E2B/pull/1749

claxman:
You're right. `timeout or default` treating 0 as unset is the current Python idiom, and #1749 is dropping those SDK defaults. Closing this.

PR #1843, opened 2026-09-08
fix(sdk): validate Volume.create names against the OpenAPI pattern

### LINKED ISSUE #1841: Volume.create skips the OpenAPI name pattern
OpenAPI `NewVolume.name` pattern is `^[a-zA-Z0-9_-]+$`. In the Python SDK with E2B_API_KEY unset, `Volume.create('')` and `Volume.create('vol name')` both raised AuthenticationException from get_api_client, so nothing checks the name before the API client is built. JS `Volume.create` in packages/js-sdk/src/volume/index.ts also builds ApiClient and POSTs /volumes with no name check.

claxman:
Fixes https://github.com/e2b-dev/E2B/issues/1841

OpenAPI `NewVolume.name` is `^[a-zA-Z0-9_-]+$`. `Volume.create('')` with no API key raised `AuthenticationException` / `AuthenticationError` from the API client. A name like `vol name` never got a client-side check.

`validateVolumeName` / `validate_volume_name` at the start of `create` (JS, Python sync, Python async). I did not validate `connect` selectors. Those are volume IDs.

Testing (no live API):

```
cd packages/python-sdk && .venv/bin/python -m pytest tests/sync/volume_sync/test_volume.py tests/async/volume_async/test_volume.py -q --tb=short
pnpm --dir packages/js-sdk exec vitest run --project unit tests/volume/createName.test.ts
```

19 sync + 19 async Python passed (6 new name cases each). 6 JS passed. Removing the three `create` checks makes the new cases fail with `AuthenticationException` / `AuthenticationError` or a stored mock volume.

```ts
await Volume.create('')
// InvalidArgumentError, including when E2B_API_KEY is unset
```

### DIFF
--- a/.changeset/volume-create-validate-name.md
+++ b/.changeset/volume-create-validate-name.md
@@ -1,0 +1,6 @@
+---
+'e2b': patch
+'@e2b/python-sdk': patch
+---
+
+Reject volume names outside `^[a-zA-Z0-9_-]+$` in `Volume.create` before constructing the API client.
--- a/packages/js-sdk/src/volume/index.ts
+++ b/packages/js-sdk/src/volume/index.ts
@@ -15,6 +15,7 @@ import {
 } from '../connectionConfig'
 import { isArrayBufferLike, isBlobLike } from '../is'
 import {
+  InvalidArgumentError,
   VolumeError,
   VolumeNotFoundError,
   VolumePathNotFoundError,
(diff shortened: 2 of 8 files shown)


claxman:
@cla-bot check

mishushakov:
I think it's fine though as it's reaches the backend it should be throwing the validation error from there.

claxman:
Agreed. The name pattern is the API's. A 400 from the backend is the source of truth. Closing this.

PR #1844, opened 2026-09-08
fix(code-interpreter): map unrecognized scale types to UNKNOWN

### LINKED ISSUE #1839: JS chart deserializer keeps unrecognized x_scale and y_scale
Python `LineChart(..., x_scale='not-a-scale')` is `ScaleType.UNKNOWN`. JS `ScaleType` in packages/code-interpreter-js/src/charts.ts has no UNKNOWN, and `deserializeChart` for line/scatter returns `{ ...data }`, so `x_scale` stays `'not-a-scale'`. ChartType already has UNKNOWN on the default branch.

claxman:
Fixes https://github.com/e2b-dev/E2B/issues/1839

Python `PointChart` maps unrecognized `x_scale` / `y_scale` to `ScaleType.UNKNOWN`. JS `ScaleType` had no UNKNOWN, `deserializeChart` spread the payload, and `Result` assigned `data['chart']` raw, so `runCode` never parsed scales.

Calling `deserializeChart` from `Result` also hit SuperChart. That branch mapped `data.data` and mutated the payload. The kernel and Python SuperChart use `elements`, so a subplot figure threw TypeError in the Result constructor.

Added `ScaleType.UNKNOWN`, parse line/scatter scales, `Result.chart` via `deserializeChart`, and SuperChart `elements` without mutating the input.

```
cd packages/code-interpreter-js && ./node_modules/.bin/vitest run tests/charts/deserializeScale.test.ts
```

5 passed. Restoring `this.chart = data['chart']` fails the line Result case with `not-a-scale`. Restoring SuperChart to `data.data.map` throws on `.map` for the SuperChart Result case.

### DIFF
--- a/.changeset/js-scale-unknown.md
+++ b/.changeset/js-scale-unknown.md
@@ -2,4 +2,4 @@
 '@e2b/code-interpreter': patch
 ---
 
-Map unrecognized chart `x_scale`/`y_scale` to `ScaleType.UNKNOWN`, matching Python.
+Map unrecognized chart `x_scale`/`y_scale` to `ScaleType.UNKNOWN`, matching Python. Deserialize SuperChart from `elements` so `Result` can run `deserializeChart`.
--- a/.changeset/js-scale-unknown.md
+++ b/.changeset/js-scale-unknown.md
@@ -1,0 +1,5 @@
+---
+'@e2b/code-interpreter': patch
+---
+
+Map unrecognized chart `x_scale`/`y_scale` to `ScaleType.UNKNOWN`, matching Python.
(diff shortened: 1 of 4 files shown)


claxman:
@cla-bot check

claxman:
Done. Result runs deserializeChart. SuperChart children come from `elements`, not `data.data`, so a subplot payload no longer throws in the Result constructor. A nested line with `x_scale: 'not-a-scale'` is ScaleType.UNKNOWN. Restoring `this.chart = data['chart']` fails the line Result test with not-a-scale. Restoring `data.data.map` throws on `.map` for the SuperChart Result test.

PR #1849, opened 2026-09-08
Add an experimental Windows-in-sandbox template (nested QEMU/KVM)

tomassrnka:
Adds `templates/windows-nested/`: a Debian + QEMU/KVM + noVNC template that boots a pre-installed Windows disk inside the sandbox, in two modes — saved-state relaunch (stock Firecracker; RDP answers 4.4–4.7 s after `Sandbox.create()` on E2B Local, verified from these files) and live (Windows kept running through the snapshot; needs a nested-state Firecracker and `no-kvmapf`, documented only) — plus the dockur/windows recipe for the disk. Not added to `templates.yml` and no SDK code or binaries: CI cannot build it.

Prerequisites: a guest kernel with `CONFIG_KVM_INTEL`, a registry the template builder can pull the ~10 GB image from, 8 vCPU / 8 GiB sandboxes with ~40 GB of node disk per build, and Windows evaluation media (180-day trial, unactivated). Sponsor: Tomas Srnka.

### DIFF
--- a/templates/windows-nested/README.md
+++ b/templates/windows-nested/README.md
@@ -1,0 +1,162 @@
+# windows-nested — a Windows desktop inside an E2B sandbox (experimental)
+
+**Status: experimental.** This template is a working proof of concept, not a supported
+product feature. It runs on a self-hosted [E2B Local](https://github.com/e2b-dev/infra-local)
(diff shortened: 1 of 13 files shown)


mishushakov:
hey, maybe move this to cookbook instead for now?

PR #1876, opened 2026-09-14
fix(sdk): send the upload-request headers the API returns for template file uploads

tomassrnka:
> **Stacked on #1875** (`chore/remove-template-build-v1`): the generated `TemplateBuildFileUpload.headers` field only exists there. Lands after it; GitHub retargets to `main` when the base merges.

### Summary

Template file uploads now send the request headers the file-upload-link response returns.

The API's `GET /templates/{templateID}/files/{hash}` response carries `headers` — "Request headers
that must be sent with the upload request" — populated when the storage backend requires a header
the signed URL cannot carry. Azure signs layer-file uploads with a SAS and returns
`x-ms-blob-type: BlockBlob`, because Put Blob requires that header and a SAS cannot express it.

Neither SDK sent them. The JS builder destructured only `{ present, url }`; both Python builders
passed only `file_info.url` to an `upload_file` that had no headers parameter. On an Azure-backed
cluster every uncached `COPY` in `Template.build()` / `AsyncTemplate.build()` failed with a
storage `400 MissingRequiredHeader`, surfacing as a file-upload error. GCS- and S3-backed
clusters return no headers, which is why this never showed up in the live-API suites.

All three upload paths now merge the returned headers into the PUT, with `Content-Length` applied
last so body framing still wins. A missing or empty map leaves the request byte-identical to
before.

No API surface change — the header set comes from the API, so no user code changes:

```ts
// unchanged; the upload now carries whatever headers the API asked for
await Template.build(
  Template().fromBaseImage().copy('folder/*', 'folder'),
  'my-template'
)
```

### Verification

- New `packages/js-sdk/tests/template/uploadHeaders.test.ts`: mocks the upload link with
  `headers: {"x-ms-blob-type": "BlockBlob"}`, drives `Template.buildInBackground` and asserts the
  PUT reaching a local server carries the header plus `Content-Length`; a second case with no
  `headers` asserts the request is unchanged.
- `uploadFile.test.ts` and both Python `test_upload_file.py` suites extended with the same two
  cases at the `buildApi` / `build_api` boundary, plus `Template._build` /
  `AsyncTemplate._build` threading tests parametrized over headers-set and `UNSET`.
- A precedence row per path returns `content-length: 1` from the API alongside `x-ms-blob-type` and
  asserts the PUT still carried the real archive size, one `content-length` header and no chunked
  encoding. Without the filter the sync Python path uploads a one-byte archive under a 200, the
  async path emits a duplicate `content-length`, and undici rejects the JS request outright.
- All new tests fail without the fix; the `index.ts` and `main.py` threading is separately forcing.
- `pnpm run lint`, `pnpm run format`, `pnpm run typecheck`, `packages/js-sdk` build;
  `uv run make lint`, `uv run make typecheck`,
  `uv run pytest tests --ignore=tests/sync --ignore=tests/async` (534 passed) and the two upload
  suites (18 passed).
- Not covered here: a real upload against an Azure-backed cluster. No Azure cluster was reachable
  from this environment.

### Risk

Low. Additive: the only behaviour change is extra request headers on the build-context PUT, and
only when the API returns them — which today means Azure clusters only. GCS and S3 paths are
byte-identical, guarded by the no-headers test cases. The `headers` map is an open
`additionalProperties: string`, so `content-length` and `transfer-encoding` are dropped from it
case-insensitively and each path applies its own `Content-Length` last — nothing the API returns can
truncate or double-frame the body.

Reviewed adversarially by one vendor (Claude `code-review`, high): five findings, all applied — the framing-header filter, the explicit sync `Content-Length`, the preserved error cause, the precedence test rows, and the msw unhandled-request guard. The Codex pass could not run (account out of credits), so this PR carries no cross-vendor review.

### Rollout

`patch` on `e2b` and `@e2b/python-sdk`. No migration, no config.

### Follow-ups

- Confirm an uncached `COPY` upload against an Azure-backed cluster end-to-end; the live-API
  template suites run only on GCS/S3 today, which is the gap that let this ship.
- Consider asserting the header round-trip in the live template build suite when an Azure cluster
  is in the matrix.


tomassrnka:
Closing as a duplicate of #1870, which fixes the same defect (opened 2026-09-11, validated live against an Azure BYOC environment — evidence this PR could not produce). Three small deltas from this branch are offered on #1870 as a comment: dropping `transfer-encoding` alongside `content-length` from the API map, keeping `error.cause` in the JS `FileUploadError` message, and an end-to-end JS test through `Template.build` that catches the `index.ts` threading half on its own.

My miss for the record: this branch was dispatched without first checking open PRs against the same code.

PR #1891, opened 2026-09-22
feat(python-sdk): mirror async result accessors on the sync CommandHandle

Atishyy27:
## Summary

`AsyncCommandHandle` and the JS `CommandHandle` expose `stdout`, `stderr`, `error` and `exit_code` as public accessors. The sync `CommandHandle` exposes only `pid`, even though it already stores all four values internally (`_stdout_chunks`, `_stderr_chunks`, `_result`). `AGENTS.md` asks for JS + sync + async parity, so this mirrors the async twin's four properties on the sync handle.

To be clear on scope: the values are already reachable from `wait()`'s return value and from `CommandExitException` on the non-zero path, so this is a consistency and ergonomics change for people moving between the sync and async handles, not a correctness fix.

## Change

Add `stdout`, `stderr`, `error` and `exit_code` to the sync `CommandHandle`, identical to `AsyncCommandHandle` including the "None while the command is still running" contract. Existing methods are unchanged.

## Testing

Added `test_sync_command_handle_exposes_result_accessors` in `tests/test_command_handle.py`, hermetic (in-memory events, no sandbox or API key). Run with `uv run pytest tests/test_command_handle.py`.

## AI-WATERMARK
yes
- Tools: Claude Code
- Model(s): Claude
- Scope: the four accessor properties (mirrored from the async twin) and the unit test
- Human verification: checked against sandbox_async/commands/command_handle.py and js-sdk/src/sandbox/commands/commandHandle.ts


Atishyy27:
@cla-bot check

